PR #4, opened 2013-01-30
Add LDAP authentication support

abustany:
A new class LdapStorage acts as a proxy for an actual storage, intercepting
authentication requests to serve them via LDAP.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,6 +55,7 @@ option(WITH_PHONON   "Enable Phonon support (for audio notifications)"	ON)
 option(WITH_LIBINDICATE "Enable Ayatana notification support"           ON)
 option(WITH_KDE      "Enable KDE4 integration"				OFF)
 option(WITH_CRYPT    "Enable encryption support if present on system"	ON)
+option(WITH_LDAP     "Enable LDAP authentication support if present on system"	ON)
 option(WITH_SYSLOG   "Use syslog for storing log data"			ON)
 
 # We use icon paths from KDE 4.3.x, which are partially invalid on older and possibly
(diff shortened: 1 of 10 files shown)


jaakristioja:
Quassel with LDAP is very useful in a corporate setting. I find it rather unfortunate that this has not been merged so far.

kradalby:
I have compiled and have it running on one of my test server. I think i will deploy it as a best effort service for my student society.

If you clone @abustany repo, the setup was pretty painless.

TC01:
I'd like to echo the request that this (or something like it) be merged or worked on-- I've been trying to find the time to write an LDAP patch but I see one has already been done.

Would it be useful if I volunteered to rebase this patch against current quasselcore?

AlD:
Hey Ben,
Help is always greatly appreciated. In it's current form it's rather unlikely for this feature to get merged though.
The right thing to do would be to make the authentication more abstract and modular in order to support different backends like e. g. PAM. This would then make the integration of various authentication methods much e easier.
Any contribution in that direction would be awesome.
Cheers, Daniel

Ben Rosser <notifications@github.com> schrieb am 10.07.2015:

kalaksi:
Btw, I needed this feature and a wrapper is better than nothing, so I rebased the commit for the current master. Can be found here: https://github.com/kalaksi/quassel/tree/ldap-wrapper
Seems to be working OK, but might be generating SQL-error-messages at least on login.

TC01:
@AlD 

So, I'd love to work towards that direction and help to get some better authentication in Quassel. Unfortunately I haven't had much free time until recently (as evidenced by the fact that I'm replying to this several months later :( ), but this past weekend I took a look at the quassel code and started trying to figure out what the "right" way to do this would be.

My thinking would be that something like the following might be a better way to modularize authentication? 

* Teach quassel-{client, core} that in addition to a storage backend, it can have an "authentication backend".
* Refactor the current database authentication behind a 'storage auth provider', that calls ```storage->validateUser()``` for whatever the currently configured storage is.
* Add a page to the core setup UI to allow an authentication provider to be selected and configured.
* Write at least an LDAP auth provider, and perhaps a PAM or SASL/GSSAPI one as well.

Does this sound reasonable? I did start hacking at the top of this, but I don't want to get too involved without checking with upstream first, since the first step turned out to be a little more involved than I was initially expecting. (Notably, it might mean making changes to e.g. [these data structures in the protocol](https://github.com/quassel/quassel/blob/master/src/common/protocol.h#L82) in order to pass around the auth provider and its configuration info, although there might be a better way to do it).

I'm idling in the quassel IRC channel if that would be a better place to discuss these things too (as TC01).

AlD:
Superseded by PR #170

PR #18, opened 2013-05-16
Second try to fix the issue in PR #16

winks:
I do agree that #16 was a quick hack - does this look better?

Not really sure how system proxies are determined in Qt, but maybe someone with a clue has a good starting point ;)

From what I found in the docs on Linux http_proxy is used (and I tested that it's found, but couldn't try to connect).

Can't build on Windows, unfortunately.


AlD:
I think this is being superseded by PR #181. Please reopen if I'm missing anything this PR did that #181 is missing.

PR #19, opened 2013-05-19
Remove jumpkeys feature and add custom buffer shortcuts

fish-face:
This feature replaces the current mechanism of "jumpkeys" which are a fixed size array of shortcuts which can be altered by using _another_ shortcut.

The new feature is quite simple. Each buffer may have a shortcut assigned to it (the idea would be for it to be mnemonic, for instance Ctrl+Shift+Q for #quassel) which can be assigned in one of two ways. The first is by right clicking on the buffer (either in sidebar or in the main chat view) and clicking "Set shortcut" (is this text obvious in the main chat? hmm...) which just opens a (quite ugly) popup with a widget to set the shortcut.
Alternatively a new page in the Shortcut settings dialog lists all buffers and (if set) the corresponding shortcut. This can then be altered in the usual fashion.

When KDE support is enabled we use the KDE shortcut settings dialog, which makes things a bit trickier. Instead of trying to add a page to that dialog, we just add a menu item to set buffer shortcuts. Furthermore, since we cannot alter the KDE settings page, it will not check for conflicts with buffer shortcuts. I don't think this is a particularly bothersome flaw, though, and the buffer shortcuts page checks for conflicts with all shortcuts.

I created issue #1221 for this merge request.

### DIFF
--- a/src/qtui/mainwin.cpp
+++ b/src/qtui/mainwin.cpp
@@ -489,7 +489,6 @@ void MainWin::setupActions()
             this, SLOT(on_jumpHotBuffer_triggered()), QKeySequence(Qt::META + Qt::Key_A)));
     coll->addAction("SetBufferShortcut", new Action(tr("Set Current Buffer's Shortcut"), coll,
                     this, SLOT(setBufferShortcut()), QKeySequence(Qt::CTRL + Qt::Key_J)));
-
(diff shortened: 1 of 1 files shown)


romibi:
I've added this to my Custom builds.
✔ Works good,
✔ less accidental changing of channel shortcuts,
✔ still quite fast remapping…
✘ no automatic upgrade of existing shortcuts (or did i miss that?). But no big deal!
I vote for merge …

PR #47, opened 2013-11-20
Add option to force ident response to the user's quassel username

Ashfire908:
Added a command-line option --oidentd-forceuser to force all ident responses to be the quassel username of the connecting user, ignoring the ident set in the user's active identity.

This does not effect the username sent on login, only the ident sent.

The purpose of this option is to comply with network requirements for an ident unique to the actual user connecting be sent, such as QuakeNet's 'Trust' system, and I know others have the same requirement for their core. I've been running a version of this patch for quite some time on my core (and now this patch) as well as on a test core without any issues.

(This is an updated version of PR #38)


AlD:
Closing due to unresolved issue. Please re-open if needed.

PR #55, opened 2014-02-03
Add button to toggle Joins/Parts/Quits with a single click.

Zren:
Added to the top of the context menu along with a separator underneath.

From a users perspective, separating the 3 events is quite silly. Parts/Quits should probably be merged together probably as well.

Couldn't find any issues for this on the tracker.


Sput42:
Merged for 0.12.0, thanks!

PR #76, opened 2014-05-26
Add a "local" display-only command

phhusson:
It's part of feature #1281:
It's a command named "local" which will only display something on current buffer with type Info, without sending anything to the server.


Sput42:
Merged for 0.12.0 - thanks!

PR #101, opened 2014-11-25
Implement password changing from client.

sandsmark:
Add menu item and dialog for getting a new password from the user,
signal from client to core to give the new password, and wire up the
signal on the core to change the password in the database.

Also update the stored password on the client side.

No idea if this could be simplified, it seems like a really round-about way to do it, passing the signal through a ton of classes. But it was the simplest way I could find.

### DIFF
--- a/src/client/client.cpp
+++ b/src/client/client.cpp
@@ -48,6 +48,7 @@
 #include "quassel.h"
 #include "signalproxy.h"
 #include "util.h"
+#include "clientauthhandler.h"
 
 #include <stdio.h>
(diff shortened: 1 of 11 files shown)


Sput42:
Great patch, but since it extends the core/client protocol, we need to protect it by the Quassel::Features mechanism: add a new enum there, and hide the UI (and don't attach signals/slots) in case the Core is too old. See, for example, how "Quassel::SaslExternal" is used...

Sput42:
Manually merged for 0.12.0. Thanks!

PR #106, opened 2015-01-07
QtMultimedia notification backend

Llamatron2112:
I used PhononNotificationBackend as base and replaced all what was done
by Phonon by QtMultimedia commands.

Phonon is not supported anymore in Qt5 and should be replaced by
QtMultimedia. Some modifications on CMakeLists.txt probably need to be
done to have the best behaviour: compiling Phonon for Qt4 and with
QtMultimedia for Qt5.


Llamatron2112:
Icons are not displaying in the settings dialog

PR #107, opened 2015-01-10
set GIT_HEAD and GIT_DESCRIBE from environment

agaida:
some additional cmake magic, make sense in a build chroot where the upstream repo is not available - i don't want to build git snapshots local - but i would like to have the information, what is build in the about box

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -555,11 +555,11 @@ git_describe(GIT_DESCRIBE --long)
 # If not in a Git repo try to read GIT_HEAD and GIT_DESCRIBE from
 # enviroment
 if (NOT GIT_HEAD OR NOT GIT_DESCRIBE)
-  if (ENV_GIT_HEAD)
-      set(GIT_HEAD ${ENV_GIT_HEAD}) 
+  if (DEFINED ENV{GIT_HEAD})
+      set(GIT_HEAD ${GIT_HEAD}) 
   endif ()
-  if (ENV_GIT_DESCRIBE)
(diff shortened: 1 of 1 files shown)


Sput42:
Not sure what you want to do with this. Would you give those variables directly to CMake (-DENV_GIT_HEAD=whatever)? Because I don't see you reading it from an environment variable...

agaida:
absolutely right - i don't have git repos in my build chroot (and i don't want to have to be true) - so i create the files at the time i do the checkout/clone from repo 

<pre>
GIT_HEAD=$(shell [ -f debian/git_head ] && cat debian/git_head)
GIT_DESCRIBE=$(shell [ -f debian/git_describe ] && cat debian/git_describe)
</pre>

and later in debian/rules

<pre>
override_dh_auto_configure:
        dh_auto_configure -- \
            -DUSE_QT5=ON \
            ...
            -DENV_GIT_HEAD=$(GIT_HEAD) \
            -DENV_GIT_DESCRIBE=$(GIT_DESCRIBE)
</pre>

it is the least invasive way to provide these informations - as we build often and not only releases in siduction i think it is useful to have the commit on which the package is based on

Sput42:
Hmm, I would change the names then, because the ENV_* is confusing (as we're not actually reading from the environment). But I'd have to think about something better...

agaida:
-DBUILD_GIT_$foo or
-DSCRCIPT_GIT_$foo or
-DDIRECT_GIT_$foo or
-DGIVEN_GIT_$foo 
only brainstorming

PR #108, opened 2015-01-25
Add QNetworkConfigurationManager-based connection detection

hagabaka:
A new network detection mode is added to use QNetworkConfigurationManager, available in Qt4.7+ including Qt5. This becomes the default mode when KDE4 integration is not available.

CMakeLists.txt just assumes the availability of QNetworkConfigurationManager based on Qt version. I haven't figured out how to actually check it, and I don't know if it will be necessary.

I've tested that it compiles with -DUSE_QT4 or -DUSE_QT5.

By the way, Networking is not available in KF5 Solid, and [porting notes](https://community.kde.org/Frameworks/Porting_Notes#Solid_Networking) point to QNetworkConfigurationManager.

### DIFF
--- a/src/client/clientsettings.cpp
+++ b/src/client/clientsettings.cpp
@@ -252,7 +252,7 @@ CoreConnectionSettings::NetworkDetectionMode CoreConnectionSettings::networkDete
 #ifdef HAVE_KDE4
     NetworkDetectionMode def = UseSolid;
 #else
-    NetworkDetectionMode def = UsePingTimeout;
+    NetworkDetectionMode def = UseQNetworkConfigurationManager;
 #endif
     return (NetworkDetectionMode)localValue("NetworkDetectionMode", def).toInt();
 }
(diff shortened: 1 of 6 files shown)


hagabaka:
It turns out I wasn't actually testing Qt4 when I created the pull request due to CMake caching. But after the above fixes, it does build for both Qt4 and Qt5 now, and automatic disconnection/reconnection work as expected.

Sput42:
Great stuff! Also good to know that it's the preferred way of doing it also for KF5.

However, I think we can make things simpler now: since we now rely on Qt 4.8+, there is no need to check for existence of the class. So the checks and the #define could be removed, making things much more readable.

I'd appreciate if you could do that and maybe afterwards squash the commits together :) Thanks!

hagabaka:
I've removed the #defines and squashed the commits. The build is failing but that seems to be because of quassel/quassel@ef13687df76b292f36f330bed44cf9f35ac53632

Sput42:
Thanks a lot!

PR #109, opened 2015-02-02
Allow buffer merge over networks.

kipe:
Allows merging buffers over networks, feature request at http://bugs.quassel-irc.org/issues/1334

> As stated in Issue 83 merge is only possible with chats on the same network.
> As the usage of Bitlbee etc has increased, it would be useful to be able to merge chats across all networks as well. The issue rises as people are now using alternative methods (Facebook, Lync etc) during the day and returning to IRC in the evening.

### DIFF
--- a/src/uisupport/bufferview.cpp
+++ b/src/uisupport/bufferview.cpp
@@ -253,16 +253,12 @@ void BufferView::dropEvent(QDropEvent *event)
     if (bufferList.count() != 1)
         return QTreeView::dropEvent(event);
 
-    NetworkId networkId = bufferList[0].first;
     BufferId bufferId2 = bufferList[0].second;
 
     if (index.data(NetworkModel::ItemTypeRole) != NetworkModel::BufferItemType)
         return QTreeView::dropEvent(event);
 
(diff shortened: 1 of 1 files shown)


Zren:
`message.senderid` is not shared across networks right? Merging buffer's just changed the `message.bufferid` to the new buffer and deletes the old buffer.  Would that cause any problems?

kipe:
`senderid' seems to be shared across networks, as there's only `senderid` and `sender` in the `sender` -table, so that shouldn't cause any problems.

Sput42:
Would be great if you could update the commit message to contain the information from the first comment in its body.

Also please add a "Closes #1334" in the last line, so we can have the issue tracker add a link to this commit and close this issue automatically.

Thanks!

kipe:
Pull request updated as requested.

kipe:
Also, I've been using my fork with this patch for few weeks now and haven't noticed any problems, even though I merge multiple chats over networks daily.

Sput42:
Merged for 0.12.0, thanks!

PR #110, opened 2015-02-08
Introduce new password hash versioning system and new salted SHA2-512 hash

mamarley:
See the individual commit messages for more info.


AlD:
Thanks for this, looks great!

A couple of questions:
- Can you provide some quick sample quasseluser table entries?
- What are the reasons for the choice of hash algos? E. g. Why Sha2 and not Sha3? Should we support both and let the core admin decide?
- Is the additional hashversion field better (i. e. in terms of maintainability) than collapsing the hash algo into the password field? (like prefixing the hash with {SSHA2} or $<version>$ etc.)
- Since the hashing isn't done in a CoreSession (or is it?) will it block the core main thread if there's a lack of entropy?

mamarley:
An example quasseluser record is "3 | testUser | edb05ba5a2c1618dc25677e59b9a13db69edfde1b0e5be79bd713d94589a4f056baac2a7282ddb599f161e7d8aee1eb4ce8237e1eae30259959c76b369891d20:8ca3cf91d053ced6ad88be1935545e2c49050f448251eb6b4f2b2f8fd64e21b5afeb18810dc4341819455db0c07b5c684c22a987bbb446ee6da6b4cc7a751d66 | 1".  This user has the password of "test".

I picked SHA2 solely because SHA3 is not supported by default in PHP, which would make adding support to Quasselsuche difficult (an extra PHP extension would be necessary in order to use it.)  I don't really think letting the user specify SHA3 is a good idea, since some users would likely pick SHA3 because it is higher, and then come complaining to us because Quasselsuche doesn't work.

I think an additional hashversion field is better because it separates the version from the hash itself.  This would be useful in the future if we support bcrypt or some other hash that has its own defined format.  It that case, it would be easier and slightly less error-prone to pull the version from the hashversion field and pass the contents of the password field directly to the hash verification function.

I am a bit fuzzy on what is actually in a CoreSession, but I did some benchmarking and determined that creating a new salt does not drain entropy noticeably and that there isn't a perceivable difference in performance even on a system that I artificially starved for entropy.  Additionally, since creating a new salt isn't a common operation (it only takes place on password changes and user adds), this further reduces the possibility of any issues.

mamarley:
Also. I would definitely be OK with adding SHA3 or bcrypt or something in the future, once it is widely supported enough on PHP not to cause problems for Quasselsuche.

mamarley:
Since I mentioned bcrypt in the previous comment, I should probably clarify that the reason I didn't choose bcrypt was because it would require a dependency on another third-party library, and Sput didn't like that idea.

allanjude:
Using something crypt() based is better, and it obviated the need for a version field, as the version is included in the hash that is stored.

https://www.freebsd.org/cgi/man.cgi?query=crypt&sektion=3

Most libc's will provide a crypt() that does SHA256 and SHA512, plus usually bcrypt() as well

PR #111, opened 2015-02-14
Don't require SSL for loopback connections with --require-ssl

mamarley:
Encrypting a loopback connection is a pointless waste of computing
power.  This patch will allow people to use loopback connections
from things like quassel-webserver without encryption while still
requiring encryption for remote connections.

### DIFF
--- a/src/common/main.cpp
+++ b/src/common/main.cpp
@@ -161,7 +161,7 @@ int main(int argc, char **argv)
     cliParser->addSwitch("oidentd", 0, "Enable oidentd integration");
     cliParser->addOption("oidentd-conffile", 0, "Set path to oidentd configuration file", "file");
 #ifdef HAVE_SSL
-    cliParser->addSwitch("require-ssl", 0, "Require SSL for client connections");
+    cliParser->addSwitch("require-ssl", 0, "Require SSL for remote (non-loopback) client connections");
 #endif
     cliParser->addSwitch("enable-experimental-dcc", 0, "Enable highly experimental and unfinished support for CTCP DCC (DANGEROUS)");
 #endif
--- a/src/core/coreauthhandler.cpp
+++ b/src/core/coreauthhandler.cpp
@@ -158,7 +158,7 @@ void CoreAuthHandler::handle(const RegisterClient &msg)
     else
         useSsl = _connectionFeatures & Protocol::Encryption;
 
-    if (Quassel::isOptionSet("require-ssl") && !useSsl) {
+    if (Quassel::isOptionSet("require-ssl") && !useSsl && !_peer->isLocal()) {
         _peer->dispatch(ClientDenied(tr("<b>SSL is required!</b><br>You need to use SSL in order to connect to this core.")));
         _peer->close();
         return;


mrngm:
Connecting to localhost is no guarantee that the connection stays on that host. It is possible to construct a tunnel, listening on localhost but connecting to some other host. It is also possible to fake the localhost address in, for example, ```/etc/hosts```. Then, ```127.0.0.1``` or ```::1``` can resolve to something completely different. See the [QT documentation](http://qt-project.org/doc/qt-4.8/qhostaddress.html#SpecialAddress-enum) on QHostAddress and special addresses and their interpretation.

I'd suggest to drop this patch, mainly because of the reason above and that ```--require-ssl``` really should **require** SSL on **all** connections.

Do you have a specific use case for this patch?

rsalinas:
Maybe a secure approach would be to add an option "*--no-ssl-for-local-port=port*".  This way you could manually override ports that are known by you to be secure because they are locally served.

mamarley:
I am completely aware that one can create tunnels that allow for a remote connection to appear to be from localhost.  However, doing so requires that one have priviliges to log in to the server question.  Additionally, of the two most common methods of doing this (found by googling for "linux port forward"), one (iptables) requires root privileges (preventing a regular user from using it) and the other (SSH) uses encryption anyway.

I am also aware that the system can be tricked into having other IPs resolve to localhost using /etc/hosts, but doing that would require root privileges on the server.  It is therefore the server administrator's responsibility to configure the system to make sure that doesn't happen.  (Most/all Linux distributions are configured this way out-of-the-box.)  If an attacker has root privileges to modify this file, then all bets are off anyway because the attacker could simply restart Quassel without --require-ssl at that point.  Even in the case that someone tried this attack, I still doubt it would work because the QHostAddress documentation states that using QHostAddress::LocalHost is the same thing as using "QHostAddress("127.0.0.1")", so it seems to me as if DNS trickery wouldn't even affect it.

I do have a specific use case for this patch: quassel-webserver.  This is an web application that is commonly hosted on the same system with a Quassel core that allows for using Quassel with an interface in the browser.  Since all the connections from this client are loopback connections, it makes so sense to encrypt them because that would be a waste of time and computing resources.  At the time I wrote the original patch to add "--require-ssl", quassel-webserver did not exist yet, so I had not thought of its use case.

I also know there is at least one other server application (the Dovecot IMAP/POP3 server) that has the same behavior.  Even when SSL is required in the configuration, it allows for unencrypted connections from localhost.  That behavior has existed for the entire time I have been using Dovecot.

I don't think the "--no-ssl-for-local-port=" option would work, since Quassel only listens on one port.

It is also worth mentioning that both the official Quassel client and Quasseldroid default to using SSL for all connections and have no way for the user to disable SSL besides recompiling the application after modifying the code.

mrngm:
> Even in the case that someone tried this attack, I still doubt it would work because the QHostAddress documentation states that using QHostAddress::LocalHost is the same thing as using "QHostAddress("127.0.0.1")", so it seems to me as if DNS trickery wouldn't even affect it.

Correct. My apologies.

> I also know there is at least one other server application (the Dovecot IMAP/POP3 server) that has the same behavior. Even when SSL is required in the configuration, it allows for unencrypted connections from localhost.

Of course, that is a decision made in the Dovecot project. I would suggest not punching *possible* security holes in an application that already tried to prevent that by requiring SSL on all connections. I see `quassel-webserver` suggests its users to connect through HTTPS. Why do you want to be able to strip the security when it hasn't reached the quasselcore yet? I'd rather be wasting my time on securing the entire connection than expect the connection to be secure, while the last hop isn't.

For what it's worth: an intruder may want to choose to keep its presence undetected. Intercepting plain text traffic is always easier than encrypted traffic.

Regarding:
> It is therefore the server administrator's responsibility to configure the system to make sure that doesn't happen.

I think it is also the responsibility of the developers of applications to make sure an application that **requires** SSL to not punch holes in it defeating that requirement/flag.

mamarley:
quassel-webserver does suggest that its users use HTTPS to connect, but that is something completely different than the "--require-ssl" flag on Quassel.  The "--require-ssl" flag only affects connections from the client application to the core.  In the case of quassel-webserver, this means between the quassel-webserver server application and the Quassel core, which would both be running on the same box.

HTTPS is the encryption between quassel-webserver (or a reverse proxy serving quassel-webserver) and the user's web browser, which would be completely unaffected by this patch.

Also, this patch does not actively disable SSL under any circumstances.  quassel-webserver defaults to using SSL, and that won't change with this patch.  The patch only allows the server administrator to not use SSL for the loopback connection from quassel-webserver to the Quassel core.

AlD:
Connections to localhost are pretty much guaranteed to terminate on the local machine without modifying the local routing table, employing NAT rules etc, which all require root privileges as already mentioned.
In a situation where the core has not bound the local port it is possible for an unprivileged user to grab the socket and intercept data. This is something dovecot doesn't have to worry about because it only deals with privileged ports. This could be worked around by using safe guards like inetd, systemd etc.
In conclusion I don't see a problem with this patch, as it is 1) reasonably unlikely to be exploited and 2) also the client's obligation to perform safety checks.

sgielen:
I wouldn't introduce such unexpected behaviour in an application for the sake of solving a little bit of extra overhead. Priority should always be behaving as expected. Maybe you could add a --no-ssl-on-localhost flag if you really have a strong use-case for this behavior?

mamarley:
As far as security is concerned, whether or not SSL is used for loopback connections simply doesn't matter.  You can't sniff traffic from the loopback adapter unless you have root privileges.  The expected behavior of --require-ssl is that all connections will be secure, and this patch doesn't change that.

sgielen:
The option is --require-ssl, not --require-secure, so that's not true. Also, loopback connection payloads can be forwarded over tunnels (not necessarily SSH or otherwise within a secure layer), as mrngm said, so even if it would be --require-secure that's not true. Yet, the only mentioned benefit, a little bit of saved performance overhead, still seems insignificant; do you have some numbers?

Sput42:
I personally think that users who are able to setup an unencrypted tunnel on their server can also ensure that whatever client goes through that tunnel uses SSL without the core enforcing it. In other words, I don't see the need to add yet another option for this particular case...

sgielen:
Was this an accidental merge after the last comment?

PR #112, opened 2015-02-17
Increase varchar to avoid Migration-fault due to really long identities

Deph0:
Fixes http://bugs.quassel-irc.org/issues/1339

Basically, if an irc-server has a long nickname limit, you can have a really long nickname, e.g. spam of letters or symbols, Instead of giving the users whom is migrating from SQLite to PostgresSQL an error, and forcing the user to truncate the values by finding (sqlite> 'select * from sender where length(sender) > 128;') and updating.


AlD:
What is the limit imposed for this by the IRC protocol? We should use that.
Please also squash fixed commits into the original commits.

Deph0:
I think I managed to squash it, my bad.

Deph0:
Closing because need to update my fork, as i did this on master branch (doh)

PR #113, opened 2015-02-19
Add option to disable ctcp and add logging of irc messages to syslog 

Captain-Matrix:
This should add two command line arguments to quasselcore , 

--no-ctcp   - this one as it implies disables the handling of ctcp requests by quassel , users obviously can still make ctcp queries by doing something like "/ctcp Captain-Matrix version" .  the reasoning for this is that while using irc ctcp requests can be annoying to some people (myself included) , they are distracting and should there be a  0 day (or public) vulnerability in quassel I wouldn't want the whole world to know the client I am using and what version, some people simply do not want to use ctcp . 

--syslog-irc  - this when enabled should log all messages between quasselcore and irc servers to the local syslog facility using a simple syslog() call . this is  a useful and very harmless feature that simply allows the user to manage irc chat as if it was normal chatter . there are many scenarios where this would be useful specially for a distributed client like quassel . if you have many quassel cores it is tedious to keep up with many local sqlite logs . it is also helpful because there are various tools out there (like splunk,logstash and elastic search as well as simple grep/sed/awk scripting ) that one can use to search logs, provide irc chat and chatroom statistics (splunk offers graphs and charts based on log queries for example) . there are many uses for this, I understand that 90% of the users won't need this but for the 10% that do please consider this for now. 

Thank you!!


Zren:
Tab found instead of space here

https://github.com/quassel/quassel/pull/113/files#diff-719f9457fc0dd1d7ed06270794fdc282R981

PR #114, opened 2015-02-20
Optional feature to log irc chat to syslog

Captain-Matrix:
--syslog-irc - this when enabled should log all messages between quasselcore and irc servers to the local syslog facility using a simple syslog() call . this is a useful and very harmless feature that simply allows the user to manage irc chat as if it was normal chatter . there are many scenarios where this would be useful specially for a distributed client like quassel . if you have many quassel cores it is tedious to keep up with many local sqlite logs . it is also helpful because there are various tools out there (like splunk,logstash and elastic search as well as simple grep/sed/awk scripting ) that one can use to search logs, provide irc chat and chatroom statistics (splunk offers graphs and charts based on log queries for example) . there are many uses for this, I understand that 90% of the users won't need this but for the 10% that do please consider this for now.

Thank you!!


AlD:
Thanks for your contribution!
Can you provide a bit more detail on what the intended usage of this interface is? As far as I understood you want to do statistical analysis of chat messages resp. all data sent to IRC servers?
I see several problems with this. The most obvious one would be the privacy and security implications. Syslog is by default unencrypted (as is IRC, but the user can choose to use encrypted connections).
I also believe that it's inefficient. I think syslog is the wrong interface for this (splunk and all the others offer database connectors, consider switching to Postgres) and it's rather inefficient to send every raw message unaggregated.
What we could think about is adding queries to the database that provide aggregated and anonymous data. This could then also be logged in regular intervals.

Captain-Matrix <notifications@github.com> schrieb am 20.02.2015:

Zren:
Er, I think messed up when trying to make two PRs as you pushed a [second commit](https://github.com/Captain-Matrix/quassel/commit/e9f68a71dd0b5f5b3be903e195aa85da4e25dbf4) that deletes the [original commit](https://github.com/Captain-Matrix/quassel/commit/5d8c13d3ef0a2d04711dd6e92b8372cfef2828b2), and implements the other feature from your original PR. You need to create a branch for each PR. If you push to a branch which currently has a pending PR, it'll update the code in the PR.

AlD:
Closing due to lack of response to raised questions. Feel free to reopen when you're ready to work on this again.

PR #119, opened 2015-03-16
use the Qca-qt5 import target if avalible

TheOneRing:


Sput42:
Merged for 0.12.0, thanks a lot!

PR #127, opened 2015-05-07
Don't embed the build date in the binary

debfx:
For compatibility send a dummy date in the client/core handshake.

This makes it possible to create reproducible builds.

See also:
https://wiki.debian.org/ReproducibleBuilds/About

### DIFF
--- a/src/core/coreauthhandler.cpp
+++ b/src/core/coreauthhandler.cpp
@@ -174,9 +174,7 @@ void CoreAuthHandler::handle(const RegisterClient &msg)
     int uphours = uptime / 3600; uptime %= 3600;
     int upmins = uptime / 60;
     QString coreInfo = tr("<b>Quassel Core Version %1</b><br>"
-                          "Built: %2<br>"
                           "Up %3d%4h%5m (since %6)").arg(Quassel::buildInfo().fancyVersionString)
-                          .arg(Quassel::buildInfo().buildDate)
                           .arg(updays).arg(uphours, 2, 10, QChar('0')).arg(upmins, 2, 10, QChar('0')).arg(Core::instance()->startTime().toString(Qt::TextDate));
 
     // useSsl and coreInfo are only used for the legacy protocol
--- a/src/client/clientauthhandler.cpp
+++ b/src/client/clientauthhandler.cpp
@@ -284,7 +284,8 @@ void ClientAuthHandler::startRegistration()
     useSsl = _account.useSsl();
 #endif
 
-    _peer->dispatch(RegisterClient(Quassel::buildInfo().fancyVersionString, Quassel::buildInfo().buildDate, useSsl));
+    // dummy build date for compatibility
+    _peer->dispatch(RegisterClient(Quassel::buildInfo().fancyVersionString, "Jan  1 1970 00:00:00", useSsl));
 }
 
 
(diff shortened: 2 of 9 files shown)


Nevcairiel:
I like knowing when my core was built, I would vote against removing harmless features for some silly reasons.

src/core/coreauthhandler.cpp
```diff
@@ -174,9 +174,7 @@ void CoreAuthHandler::handle(const RegisterClient &msg)
     int uphours = uptime / 3600; uptime %= 3600;
     int upmins = uptime / 60;
     QString coreInfo = tr("<b>Quassel Core Version %1</b><br>"
-                          "Built: %2<br>"
                           "Up %3d%4h%5m (since %6)").arg(Quassel::buildInfo().fancyVersionString)
-                          .arg(Quassel::buildInfo().buildDate)
```
Zren: Removing a format argument will shift the index of all other arguments.

netzvieh:
Going with Nevcairiel here, if you want reproducible builds I'd suggest keeping this as patch in Debian or change the PR so you have an option for that in cmake.

src/core/coreauthhandler.cpp
debfx: It still works as arg() replaces the marker with the lowest number.
I'll clean up the markers anyway as it's a bit ugly like that.

debfx:
I'm kinda disappointed that you disregard an effort to improve security without even bothering to give a single argument.

There is nothing Debian specific in this change. Debian just happens to be the first distro that makes a significant effort to get packages built reproducible.

I can work on turning this into an optional feature if you'd merge that.

romibi:
I was thinking about this a bit...
Apart from the reason why a build should be reproducible (which I'm not 100% sure about) ... 

Why not instead of removing the (build-)date completely, just replace it with a commit-date?
Most of the time it would be pretty close to each other and then those who want a date there are happy and the people who want reproducible builds are happy...

AlD:
Sounds good to me.

romibi <notifications@github.com> schrieb am 26.10.2015:

CounterPillow:
@romibi I'd be even fine if it just kept the commit hash in the code, since from that it's easy to find out what date the code is from. Build date is just a bad approximation of this anyway; it gives you an upper bound on how fresh the code is, but no actual info on what code the binary is actually built from. What you care about is a lower bound on how fresh the code is though, which the build date does not provide, and thus I fail to see its usefulness.

I'd be interested to hear whether anybody has arguments for keeping the build date, other than "it has always been there". For what it's worth, I also wouldn't describe reproducible builds as "silly reasons" - while Quassel is certainly not a prime target for maliciously altered packages like bitcoin is, it's still a nice to have property of the build system since it could allow for automated independent third party byte-for-byte reproduction of the builds.

Nevcairiel:
@CounterPillow 
The full git version is included already, the build date is just an extra piece of information.

AlD:
Removing the build date entirely would also be fine with me. Just replacing it with a dummy date as proposed in https://github.com/debfx/quassel/commit/d1107f360bb5e97f61c410c364a80eeb37162c2b is a bit too ugly.

rsalinas:
I also think reproducible builds are something great, and that the price we pay by not including the BUILD timestamp is very little.  The build time is actually not even related to the software version itself, so that it is not such a big loss.  Anyway, we might well use the executable's creation timestamp to know the compilation time.

I also believe that hardcoding a false date is ugly.

AlD:
Superseded by #159.

PR #131, opened 2015-05-22
Clang and Qt4 build fix

fuzzball81:
While building on my Mac I found a build issue using clang and Qt4. I made the following fixes and activated the build in travis to make sure there were no regression issues.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,10 +14,10 @@ install:
 language: cpp
 
 compiler:
-  - gcc 
+  - gcc
   - clang
 
-env: 
+env:
   - QT_VERSION=qt4
   - QT_VERSION=qt5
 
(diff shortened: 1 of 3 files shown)


Nevcairiel:
Also, wasn't the consensus that Qt5 works much better on a Mac anyway?

fuzzball81:
Made the change to use auto instead, that was a good catch.

Looking at both the Qt4 and Qt5 documentation of append I do not see where they allow for an initializer_list as an argument. If I am missing something, please point me in the right direction.

Sput42:
The reason for the build failure is the fact that Qt4 does not enable any C++11 features for Clang (i.e. the std::initializer_list ctor is disabled when building with Clang). While this patch works around this issue, there will most likely be more instances of that sort of problem popping up.

Then again, we'll be removing Qt4 support after the next release anyway, it seems.

fuzzball81:
I would suggest that something be added to INSTALL stating that Qt5 needs to be used when compiling with Clang. It might also be a good idea to add some CMake foo to stop the build if Qt5 is not available and the user is compiling with Clang. This would help with the migration away from Qt4.

AlD:
Closing as this seems resolved. Feel free to send a PR to add the suggested hint to INSTALL, though.
Thanks

PR #136, opened 2015-06-16
Add spell-checking support via the Sonnet framework

sandsmark:
I rebased your patch and stuff

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -262,17 +262,22 @@ if (USE_QT5)
 
         if (ECM_FOUND)
             list(APPEND CMAKE_MODULE_PATH ${ECM_MODULE_PATH})
-        endif()
-
-        if (WITH_KDE)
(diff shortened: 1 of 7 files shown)


avih:
ping...

avih:
Should this be working out of the box now on Windows? it seems that the appveyor builds don't include `KF5Sonnet` as far as I can tell (as far as @romibi can tell, actually) and I'm also not seeing any speller option at the input widget config.

sandsmark:
yeah, appveyor needs to be updated. I'm not familiar with it, but it looks like it should be very straightforward, just copying most of the config for e. g. snorenotify and replacing "snorenotify" with "sonnet" (they're both on quickgit.kde.org, for example).

avih:
How would dictionaries be handled (once it works)? Should they be downloaded manually (where from?) and place them at the quassel dir?

PR #137, opened 2015-06-24
Snore 0.6

TheOneRing:
This pull request adds support for the current unstable version of Snorenotify.
With the new support for notification backends for Android, IOS and Windows Phone Snore is also interesting if KDE notifications have been found.

The sound backend could also replace the phonon backend on all systems with QMultimedia support.

A possible problem could be that integration of the settings ui, which has its own strings and misses most  translation.


Sput42:
Hi,

thanks for this! However, I'd appreciate if you could squash the commits somewhat, and also provide proper (capitalized and a bit more verbose) commit messages to keep our git log pretty and useful :)

PR #138, opened 2015-07-02
Performance improvements

HeisSpiter:
Really limited improvements only based on static analysis, but still, can help.

### DIFF
--- a/src/client/bufferviewoverlay.cpp
+++ b/src/client/bufferviewoverlay.cpp
@@ -70,7 +70,7 @@ void BufferViewOverlay::restore()
     currentIds += CoreAccountSettings().bufferViewOverlay();
 
     QSet<int>::const_iterator iter;
-    for (iter = currentIds.constBegin(); iter != currentIds.constEnd(); iter++) {
+    for (iter = currentIds.constBegin(); iter != currentIds.constEnd(); ++iter) {
         addView(*iter);
     }
 }
@@ -144,7 +144,7 @@ void BufferViewOverlay::removeView(int viewId)
         else {
             if (!config->isInitialized())
                 _uninitializedViewCount++;
-            viewIter++;
+            ++viewIter;
         }
     }
 
@@ -211,7 +211,7 @@ void BufferViewOverlay::updateHelper()
     if (Client::bufferViewManager()) {
         BufferViewConfig *config = 0;
         QSet<int>::const_iterator viewIter;
-        for (viewIter = _bufferViewIds.constBegin(); viewIter != _bufferViewIds.constEnd(); viewIter++) {
+        for (viewIter = _bufferViewIds.constBegin(); viewIter != _bufferViewIds.constEnd(); ++viewIter) {
             config = Client::bufferViewManager()->bufferViewConfig(*viewIter);
             if (!config)
                 continue;
--- a/src/client/client.cpp
+++ b/src/client/client.cpp
@@ -535,7 +535,7 @@ void Client::networkDestroyed()
             break;
         }
         else {
-            netIter++;
+            ++netIter;
         }
     }
 }
--- a/src/client/clientirclisthelper.cpp
+++ b/src/client/clientirclisthelper.cpp
@@ -43,7 +43,7 @@ void ClientIrcListHelper::receiveChannelList(const NetworkId &netId, const QStri
         QVariantList channelVar = iter->toList();
         ChannelDescription channelDescription(channelVar[0].toString(), channelVar[1].toUInt(), channelVar[2].toString());
         channelList << channelDescription;
-        iter++;
+        ++iter;
     }
 
     emit channelListReceived(netId, channelFilters, channelList);
--- a/src/client/clientsettings.cpp
+++ b/src/client/clientsettings.cpp
@@ -216,7 +216,7 @@ QSet<int> CoreAccountSettings::bufferViewOverlay()
 {
     QSet<int> viewIds;
     QVariantList variants = accountValue("BufferViewOverlay").toList();
-    for (QVariantList::const_iterator iter = variants.constBegin(); iter != variants.constEnd(); iter++) {
+    for (QVariantList::const_iterator iter = variants.constBegin(); iter != variants.constEnd(); ++iter) {
         viewIds << iter->toInt();
     }
     return viewIds;
--- a/src/client/messagefilter.cpp
+++ b/src/client/messagefilter.cpp
@@ -201,7 +201,7 @@ bool MessageFilter::filterAcceptsRow(int sourceRow, const QModelIndex &sourcePar
             while (idIter != _validBuffers.constEnd()) {
                 if (Client::networkModel()->bufferType(*idIter) == BufferInfo::StatusBuffer)
                     return true;
-                idIter++;
+                ++idIter;
             }
         }
 
@@ -241,6 +241,6 @@ void MessageFilter::requestBacklog()
     QSet<BufferId>::const_iterator bufferIdIter = _validBuffers.constBegin();
     while (bufferIdIter != _validBuffers.constEnd()) {
         Client::messageModel()->requestBacklog(*bufferIdIter);
-        bufferIdIter++;
+        ++bufferIdIter;
     }
 }
--- a/src/client/messagemodel.cpp
+++ b/src/client/messagemodel.cpp
@@ -206,7 +206,7 @@ int MessageModel::insertMessagesGracefully(const QList<Message> &msglist)
     QList<Message>::const_iterator iter;
     if (inOrder) {
         iter = msglist.constEnd();
-        iter--; // this op is safe as we've allready passed an empty check
+        --iter; // this op is safe as we've allready passed an empty check
     }
     else {
         iter = msglist.constBegin();
@@ -229,11 +229,11 @@ int MessageModel::insertMessagesGracefully(const QList<Message> &msglist)
     }
 
     if (!inOrder)
-        iter++;
+        ++iter;
 
     if (inOrder) {
         while (iter != msglist.constBegin()) {
-            iter--;
+            --iter;
 
             if (!fastForward && (*iter).msgId() <= minId)
                 break;
@@ -295,7 +295,7 @@ int MessageModel::insertMessagesGracefully(const QList<Message> &msglist)
                 dupeId = (*iter).msgId();
                 grouplist.prepend(*iter);
             }
-            iter++;
+            ++iter;
         }
     }
 
--- a/src/client/networkmodel.cpp
+++ b/src/client/networkmodel.cpp
@@ -723,7 +723,7 @@ void ChannelBufferItem::addUsersToCategory(const QList<IrcUser *> &ircUsers)
     QHash<UserCategoryItem *, QList<IrcUser *> >::const_iterator catIter = categories.constBegin();
     while (catIter != categories.constEnd()) {
         catIter.key()->addUsers(catIter.value());
-        catIter++;
+        ++catIter;
     }
 }
 
--- a/src/client/selectionmodelsynchronizer.cpp
+++ b/src/client/selectionmodelsynchronizer.cpp
@@ -96,7 +96,7 @@ void SelectionModelSynchronizer::selectionModelDestroyed(QObject *object)
             iter = _selectionModels.erase(iter);
         }
         else {
-            iter++;
+            ++iter;
         }
     }
 }
@@ -252,7 +252,7 @@ void SelectionModelSynchronizer::currentChanged(const QModelIndex &current, cons
     QSet<QItemSelectionModel *>::iterator iter = _selectionModels.begin();
     while (iter != _selectionModels.end()) {
         (*iter)->setCurrentIndex(mapFromSource(current, (*iter)), QItemSelectionModel::Current);
-        iter++;
+        ++iter;
     }
     _changeCurrentEnabled = true;
 }
@@ -267,7 +267,7 @@ void SelectionModelSynchronizer::selectionChanged(const QItemSelection &selected
     QSet<QItemSelectionModel *>::iterator iter = _selectionModels.begin();
     while (iter != _selectionModels.end()) {
         (*iter)->select(mapSelectionFromSource(currentSelection(), (*iter)), QItemSelectionModel::ClearAndSelect);
-        iter++;
+        ++iter;
     }
     _changeSelectionEnabled = true;
 }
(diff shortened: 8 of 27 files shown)


Bombe:
I’m not really sure what centure your C++ compiler was developed in but I’m pretty sure that anything from the last 15 years will already have optimized that without telling you. :)

HeisSpiter:
It's basically not only a matter of compiler:
- http://stackoverflow.com/a/3181240/1913488
- http://programmers.stackexchange.com/a/60002

So, I invite you yo read More Effective C++ from Scott Meyers, and more specifically the Item 6 from the book which should be provide you interesting views about why postfix operator is to be avoided unless necessary. Some points are summed up in the two previous links.

sandsmark:
If this is a performance improvement, do you have any benchmarks (even informal ones)?

sandsmark:
(Modern compilers do in fact optimize postfix operators on iterators, hence why I ask if you have actually benchmarked this.)

HeisSpiter:
Being shared or not, passing by copy will increase reference counter twice (first copy by passing, second copy by assigning) and decrease it once (deleting the copy once the assignment is done).
When passing by reference, you avoid at least the first increase/decrease because not attempt for copy is tried at all (just passing a pointer). Just for the record, an atomic counter has a cost.
If you have a doubt, have a look to this code, I wrote to show what I'm saying: https://gist.github.com/HeisSpiter/a1d61da0674a3e14fc21. Build it with g++ -g -std=c++11 -O3 shared.cpp, and run it with ./a.out. See how many times delete operator is called.

Regarding the iterators, compilers are know to indeed optimize standard iterators, because they are known (part of the stdlib). This is not the case for extra added iterators, especially because the compiler might not know the underlying implementations. We might quote this, for instance: http://fairwaytech.com/2012/03/prefix-vs-postfix-increment-and-decrement-operators-in-c/ (even though the conclusion is wrong though the justifications are right (cf: third comment).

I unfortunately don't have any benchmarks available, this is something really hard to properly evaluate (to provide meaningful benchs). Feel free to close the PR then. I guess I justified enough why there's a room for improvement in spite of general beliefs.

sandsmark:
My point was that creating a reference (a pointer with atomic counter and extra stuff) and increasing/decreasing its atomic counter is more expensive than using the QString's existing atomic counter. As far as I know, at least, feel free to correct me.

As for the iterators; the compiler is separate from the standard library, and the code for iterators in Qt are pretty much the exact same as in STL (and in every other sane implementation of iterators), so the same optimizations apply.

And for benchmarks, if you don't have any benchmarks of the code in use here, at least benchmark your ideas in general? Pass a gazillion QStrings into a function and compare the timing with and without a reference, or run a QList iterator a gazillion times with prefix and postfix operators.

HeisSpiter:
A reference has no atomic counter?! A reference is purely a pointer with syntax sugar. So, obviously, it's faster to use than attempting to deal directly with the QString (hence my example!).

Regarding the iterators, as pointed before with the reference to Scott Meyers' book More Effective C++, it's not only a matter of performance.
And extracting a dummy test just to have it run and measured might not be relevant at all, compiler optimizations being dependent on the whole context. This is why I didn't provide a code snippet here, compared to the reference thingy where it was to demonstrate a C++ behavior.

sandsmark:
Sorry, complete brainfart on my part regarding the atomic counters, thanks for the patience. :-)

But because a reference to a QString is a pointer to a pointer, each time you use it you need to do a double dereferencing. So passing a QString by value is a constant but pretty low overhead (creating a pointer + increasing an atomic int) each time you call the function, while passing by reference leads to a double dereferencing (with clobbering of registers and whatnot) each time you actually use the QString.

And with iterators, I would appreciate it if you could summarize the other reasons for using prefix iterators. The book is already on my reading list, but since you apparently know the relevant information I'd like to learn them sooner rather than later. But I hope you agree now that postfix vs. prefix has no performance benefits for these iterators (this should be easy to verify either with a synthetic test, looking at the source for the STL iterators and these iterators, or just a disassembly of the generated code).

And I agree that synthetic tests aren't very valuable because in the actual code I'm fairly certain that the improvements by your commits here are lower than the measurement error (possibly worse because of the QString references), but it would be better than just guesswork.

Lastly, I don't think this came through clearly earlier; I'm not the one that decides if this goes in or not, and I haven't said your changes shouldn't go through. The passing of things by reference seems to be the standard, so your change makes the code more consistent. And using prefix iterators in general also makes the code more consistent with the cases where you aren't increasing a proper iterator and it therefore incurs an actual overhead. But you marked this as a pure performance improvement, which I found weird.

sandsmark:
After talking with some people, CSE might negate the problem with double dereferencing (if the compiler can show that it doesn't alias), but it still takes real profiling to see if using references here actually improves performance.

HeisSpiter:
> But because a reference to a QString is a pointer to a pointer, each time you use it you need to do a double dereferencing. So passing a QString by value is a constant but pretty low overhead (creating a pointer + increasing an atomic int) each time you call the function, while passing by reference leads to a double dereferencing (with clobbering of registers and whatnot) each time you actually use the QString.

No, you take it wrong. The reference is a pointer to a QString object. Not a pointer. The object itself contains a pointer to a memory zone which is potentially shared amongst objects. This is what I implemented (in a more simplistic way) in my example. So, when you pass your object by value, in every case, you will create a local copy, data being shared or not. And this definitely has a cost. Because you initialize a vtable, a whole object and finally, you dereference the shared data to increment its reference counter.  And this, just to delete it once the call is over, because it was a local copy. So, you dereference again the data pointer just to decrease its reference counter. And in the middle of this, you even dereferenced the data pointer once more, because you copied data to the local variable and thus incremented reference count.
With passing by reference/pointer, you just pass the object QString itself, and you don't need all this stuff. You'll just dereference its pointer, to copy its data address and increment its reference counter. Full stop.
You cannot expect passing by value being faster than passing by reference. Even with CoW mechanisms.

Regarding the iterators, I'll just quote the most valuable part of the Item 6 of the Scott Meyers' book:
> Let us make one more observation about the prefix and postfix incre-
ment operators. Except for their return values, they do the same thing:
they increment a value. That is, they’re supposed to do the same thing.
How can you be sure the behavior of postfix increment is consistent
with that of prefix increment? What guarantee do you have that their
implementations won’t diverge over time, possibly as a result of differ-
ent programmers maintaining and enhancing them? Unless you’ve fol-
lowed the design principle embodied by the code above, you have no
such guarantee. That principle is that postfix increment and decre-
ment should be implemented in terms of their prefix counterparts. You
then need only maintain the prefix versions, because the postfix ver-
sions will automatically behave in a consistent fashion.

HeisSpiter:
For the record, I profiled the connection process of Quassel which I found particularly slow (referring to the moment when you connect to core, gather networks information and backlogs).

Here is what we find out:
 time   seconds   seconds    calls  ms/call  ms/call  name    
 18.63      0.10     0.10  8666116     0.00     0.00  QBasicAtomicInt::deref()
 18.63      0.19     0.10  6034977     0.00     0.00  QBasicAtomicInt::ref()
  3.92      0.21     0.02   485480     0.00     0.00  MessageModel::data(QModelIndex const&, int) const
  1.96      0.22     0.01  7171896     0.00     0.00  QString::~QString()

So, passing value by reference doesn't seem overzealous here, since it appears that atomic counters (QBasicAtomicInt is what is used in QString implementation) are our major vast of time. So avoiding some can be a plus. And avoiding some ~QString() calls too.

For the record, it was done on my branch, so with already my PR applied, so, branch which reduces (a little bit) number of QString construction/destruction.

sandsmark:
A QString has no virtual methods, so no vtable, so the only thing that gets allocated is the pointer to the internal, shared QStringData. When you allocate an object of a class containing just a POD variable that is pretty much the same as allocating just a POD variable.

The QString constructor also increases the atomic counter, and the destructor decreases it, which costs about the same as two L2 cache hits (so approximately 20 clock cycles). So while the initiall call to the function adds a cost of approximately two L2 cache hits, everytime you then use that QString you have one less L1 cache hit (and you don't clobber the L1 cache). The hits to the L2 cache is usually ~10 clock cycles, the L1 cache is ~4 clock cycles.

So lets assume that the *extra* time needed to pass the QString by value is 20 clock cycles, while each time you use the QString you add at least 4 clock cycles plus that it might evict other useful values from the L1 cache leading to slowdown of other operations. So by passing by reference, assuming we only use the reference once, the speedup is less than 16 clock cycles, and probably decrease performance when looking at the overall code because of the extra hits to the L1 cache and evictions from it.

sandsmark:
Ah, didn't see your latest comment. What tool did you profile with?

HeisSpiter:
In your explanation, you just forgot to take into account one major data: the cost of the "LOCK" ASM instruction used for the atomic counter. It's not free. Far from it.

I used gprof (along with GCC 4.9), because callgrind basically just cannot handle quasselclient in decent times.

Anyway, enough of it. I let you with your beliefs.

sandsmark:
No, atomic ints on modern CPUs cost pretty much the same as one L2 cache hit.

Also, gprof isn't a very good tool for profiling, because it modifies the code and therefore changes the performance characteristics. I'd rather use perf, which also can look at cache hits etc.

sandsmark:
*incrementing/decrementing atomic ints.

HeisSpiter:
> No, atomic ints on modern CPUs cost pretty much the same as one L2 cache hit.
Can you provide facts and not generic beliefs? I made my best to document any of the facts I was providing.
http://stackoverflow.com/questions/2538070/atomic-operation-cost
And this is just in the case you don't have to wait. As written in the page itself: "In all cases, a cacheline request can be stalled by other processors already modifying the data."

> Also, gprof isn't a very good tool for profiling, because it modifies the code and therefore changes the performance characteristics. I'd rather use perf, which also can look at cache hits etc.
Top by perf for quasselclient itself:
0.85%  quasselclient  [.] QBasicAtomicInt::deref
0.74%  quasselclient  [.] QBasicAtomicInt::ref
0.12%  quasselclient  [.] MessageModel::data
0.11%  quasselclient  [.] QHash<BufferId, BufferItem*>::findNode
0.09%  quasselclient  [.] QString::~QString

Seems like gprof wasn't that wrong (values are reduced though, because contrary to gprof, perf can profile everything include shared objects).

But once again, I do really give up.

sandsmark:
I was assuming best-case for both using a reference and not using a reference, with everything in the L1 cache. Doing a CAS for the atomic reference counter in the L1 cache is the L1 cache hit + some cycles of actually performing it, which sums to about 10 clock cycles (if you want a reference I can dig up the relevant Intel manuals, I'm basing this mostly on the characteristics of a Xeon processor).

As for the profiling results, can you see the traces for where the atomic int refs and derefs are coming from? Are they from places that could use a reference instead?

HeisSpiter:
Just for the record, a loop with my previous example code.

When using copy (MyFunc):
[schweitzer@leptoquark /tmp]$ time ./a.out > /dev/null 
10.94user 6.34system 0:17.29elapsed 99%CPU (0avgtext+0avgdata 2588maxresident)k
0inputs+0outputs (0major+111minor)pagefaults 0swaps
[schweitzer@leptoquark /tmp]$ time ./a.out > /dev/null 
10.84user 6.35system 0:17.20elapsed 99%CPU (0avgtext+0avgdata 2576maxresident)k
0inputs+0outputs (0major+108minor)pagefaults 0swaps
[schweitzer@leptoquark /tmp]$ time ./a.out > /dev/null 
10.33user 6.74system 0:17.09elapsed 99%CPU (0avgtext+0avgdata 2576maxresident)k
0inputs+0outputs (0major+108minor)pagefaults 0swaps

When using reference (MyFunc2):
[schweitzer@leptoquark /tmp]$ time ./a.out > /dev/null 
7.40user 4.50system 0:11.91elapsed 99%CPU (0avgtext+0avgdata 2572maxresident)k
0inputs+0outputs (0major+108minor)pagefaults 0swaps
[schweitzer@leptoquark /tmp]$ time ./a.out > /dev/null 
7.44user 4.36system 0:11.81elapsed 99%CPU (0avgtext+0avgdata 2692maxresident)k
0inputs+0outputs (0major+111minor)pagefaults 0swaps
[schweitzer@leptoquark /tmp]$ time ./a.out > /dev/null 
7.44user 4.17system 0:11.62elapsed 99%CPU (0avgtext+0avgdata 2648maxresident)k
0inputs+0outputs (0major+110minor)pagefaults 0swaps

Figures talk by themselves.

AlD:
Let's keep this open for now, shall we? Nothing has been decided yet :)

sandsmark:
Yes, as I said, the change is good for consistency, I just questioned the fact that it was actually a performance improvement.

As for the speed of your example code, looking at the disassembly it seems to be able to do CSE in the simple cases you have there, it does the de-referencing before calling your function, as I mentioned earlier.

Sput42:
So, I'm inclined to merge this (after a cursory glance; I'm still working through the open PRs to proper review). I agree with sandsmark about it probably not being very relevant to performance, but at the very least for consistency reasons (there's tons of parts in our codebase that are very old, and it shows sometimes...) Thanks a bunch for going through this!

Now, the QString discussion is an interesting one, and not as trivial as it might seem. In fact, using a const ref may very well be less performant. I recall an old blog post by Thiago Maciera about that; at least on amd64, passing a reference (or a pointer) is *more* expensive than passing a small (up to 16 bytes) object by value. Reason being that the compiler can do certain optimizations involving registers and other nifty things for small objects that are not possible with references. Now, a QString is small enough for that to matter; on the other hand, atomic operations are quite costly on modern processors. I still seem to remember that Thiago's benchmarks came out in favor of passing QString by value, up to the point where changing the APIs for Qt5 accordingly was a very real consideration that was only abandoned due to the amount of effort required (and the fear of introducing sideeffects). Also, consistency again. That latter argument is why I'm in favor of using const refs everywhere in our current codebase.

Now, there's something else to consider: we have C++11 now! If the callee needs to take a copy of the argument, the most efficient way to pass arguments in C++11 is by value in general, assuming that the object to be passed has a move constructor (QString does), and that you use std::move() where appropriate. The reasoning is as follows:

* If the argument given is an rvalue, no extra copy is made (due to the move ctor being invoked)
* If the argument is an lvalue, but the caller knows it won't need it anymore, the caller can avoid that copy using std::move()
* The callee should std::move the argument into its final location if it requires a copy. Otherwise, for non-mutating access only, the API should provide a const ref (under the caveats mentioned above; until we've benchmarked it, we don't know if QString in particular is more efficient to be passed by value anyway).

This is a highly interesting topic, and once we've cleaned up our codebase a bit (I'd say after we've gotten rid of support of Qt4 and KDE4, at least), it would be a good thing to do proper profiling and tackle the hotspots. I'm sure there's some much more relevant issues, such as our abuse of QVariantMap (this is both insanely expensive and memory-fragmenting), but call semantics are surely not on the bottom of the list either.

For now, I think merging this is fine (pending a full review), and thanks again for the effort!

Sput42:
Finally got around to merging this. Thanks again!

For the record: independent of actual performance benefits, I like these changes for consistency reasons. Pre-inkrement is a standard pattern, and currently we pass all QStrings per const ref (the bug fixed was an actual oversight, not intended).

PR #140, opened 2015-07-20
add "flair" to nicks with voice/ops/etc.

crowell:
add + for voice, @ for ops and more.

should this be configurable?
also what about adding this to things like the mobile client?


crowell:
messed up squash, redoing

PR #141, opened 2015-07-20
add "flair" to nicks with voice/ops/etc.

crowell:
add + for voice, @ for ops and more.

### DIFF
--- a/src/uisupport/uistyle.cpp
+++ b/src/uisupport/uistyle.cpp
@@ -21,6 +21,7 @@
 #include <QApplication>
 #include <QIcon>
 
+#include "client.h"
 #include "buffersettings.h"
 #include "qssparser.h"
 #include "quassel.h"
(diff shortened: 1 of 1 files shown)


Zren:
* Whitespace
* Definitely would need to be optional (since I personally don't want them).
* `git commit --amend` (or `git reset --soft HEAD^1`) + `git push -f` instead of making new PRs.

benapetr:
Do you realize that this code is broken for some irc servers? The prefixes for users are defined by IRC server in login messages:
```WALLCHOPS WATCH=128 WATCHOPTS=A SILENCE=15 MODES=12 CHANTYPES=# PREFIX=(qaohv)~&@%+ CHANMODES=beI,kfL,lj,psmntirRcOAQKVCuzNSMTGZ NETWORK=tm-irc CASEMAPPING=ascii EXTBAN=~,qjncrRa ELIST=MNUCT STATUSMSG=~&@%+ are supported by this server```
They can be anything, on some server voiced user might have prefix of * if they like so. IRC client need to respect this and not hardcode symbols that are used elsewhere.

crowell:
Ah I wasn't aware of that. So it seems like the logon message must be parsed for this. Is it stored in quassel somewhere? I'm a bit new to the codebase

Nevcairiel:
I think quassel parses them, in what format it stores them I'm not sure. Could grep for the appropriate tokens, shouldn't be too hard to find.

AlD:
Superseded by #162.

PR #144, opened 2015-09-16
Some changes for OSX

stonerl:
Since icons are uncommon in the Global Menu and usually in the context-menus as well, they won't be shown on OSX. Also the the "Show Menubar" Button has no use on OSX and has therefor been removed.

Furthermore a small workaround. The the inputwidget does not resize in OSX with a font smaller 14pt. So a hint was added to the settingspage and by default the font will be set to 14.

### DIFF
--- a/scripts/build/Info.plist
+++ b/scripts/build/Info.plist
@@ -30,19 +30,6 @@
 	<string>NSApplication</string>
 	<key>NSHighResolutionCapable</key>
 	<true/>
-	<key>CFBundleURLTypes</key>
-	<array>
-		<dict>
-			<key>CFBundleURLName</key>
-			<string>IRC Service URL</string>
-			<key>CFBundleURLIconFile</key>
-			<string>%(ICON_FILE)s</string>
-			<key>CFBundleURLSchemes</key>
-			<array>
-				<string>irc</string>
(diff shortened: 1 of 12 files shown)


Deph0:
You forgot right click toggle for menubar, and could you elaborate on what you mean with the input widget? it resizes just fine with fontsize 9 for me.

-Deas

stonerl:
Since I removed it from the Global menu it also is not available from the right click menu. The problem with the inputwidget is quite simple. It only resizes when I use a font with 14pt or bigger. If I use a smaller font then it only creates multiple lines but it does not resize the input field according to the numbers of max-lines I have set.

Deph0:
Another idea could be to implement the "minimized" function to closeing the window so quassel doesn't quit because you hit X (this feature already exists inside the tray icon).

For example, in Skype, you have to right-click the dock icon to quit skype, if you hit X, you close the window but not the instance, and you can hit cmd+1 to close the window(not the instance) as well.

AlD:
The commits in this PR seem to address different unrelated issues.
Please submit them in individual PRs and if you could add screenshots to describe the UI changes that would help reviewing a lot. Thanks!

PR #145, opened 2015-09-18
Adds nicer ircuser tooltips

Scheirle:
New tooltip:
tooltip

The "names" (Realname, ...) may have to be adjusted.

What do you think?


digitalcircuit:
Ack, I didn't realize every force-push would put a note here, too.  I apologize for the spam!

PR #147, opened 2015-09-23
Performance tweak: rem. pointless recursive calls

benapetr:
For some reason there was an extra call to IrcChannel::joinIrcUser from IrcUser::joinIrcChannel. That means that in past we did:

* Call to                                 IrcChannel::joinIrcUsers(...)
* From there we call              IrcUser::joinIrcChannel(this)
* And from there we called    IrcChannel::joinIrcUser(this)
* Which again called             IrcChannel::joinIrcUsers(list that contained only this 1 user)
* Which figured out this user is already in this channel and exits

That is a pointless overhead, so now we call the second function with extra parameter that tells it not to call joinIrcUser because we know this user doesn't need to be added there, thus saves few thousands of pointless function calls and user list contructions, especially while joining large channels.

### DIFF
--- a/src/common/ircchannel.cpp
+++ b/src/common/ircchannel.cpp
@@ -183,7 +183,7 @@ void IrcChannel::joinIrcUsers(const QList<IrcUser *> &users, const QStringList &
         }
 
         _userModes[ircuser] = modes[i];
-        ircuser->joinChannel(this);
+        ircuser->joinChannel(this, true);
         connect(ircuser, SIGNAL(nickSet(QString)), this, SLOT(ircUserNickSet(QString)));
 
         // connect(ircuser, SIGNAL(destroyed()), this, SLOT(ircUserDestroyed()));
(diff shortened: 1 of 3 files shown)


benapetr:
Just to elaborate on why this saves "thousands of function calls":

In past there was an overhead of at least 2 internal quassel functions that didn't need to be called, these however contains code that calls at least another 7 functions because of this code:

```
    QList<IrcUser *> users; // calls ctor of QList
    users << ircuser; // calls insert() or append()
    QStringList modes; // calls ctor of QStringList
    modes << QString(); // calls ctor of QString and append()
    // calls dtor of QList and QStringList
```
That is about 10 extra function calls per 1 channel user. Given that there are channels with about ~1000 users, that is extra 10 000 function calls that are just not needed per 1 such a channel. Of course, you won't feel this on last Intel i7 processors, but it's still code that smells... :P

Sput42:
It smells indeed :) Thanks for fixing, merged for the next release.

PR #149, opened 2015-09-24
Store the channel keys in CoreNetwork again

Tucos:
The refactoring from 3146ad0 (and e0781ee) was not a true refactoring as
it changed behaviour: CoreIrcChannels (as opposed to CoreNetworks) are
ephemeral, which meant disconnecting or parting lost the channel key.
Keys are still kept by the channel's cipher, but synced from/to its network when
the constructor/destructor is called.

Fixes #1248


sjefen6:
Thanks!

sjefen6:
I (attempted to) reopen #1248 as the issue still persist for queries.

PR #154, opened 2015-10-31
quasselcore: fixed return values for --add-user and --change-userpass

rsalinas:
Maybe it is not a common use case to include a user addition in a script, but in such case we need this.

### DIFF
--- a/src/core/core.h
+++ b/src/core/core.h
@@ -533,7 +533,7 @@ private slots:
     void socketError(QAbstractSocket::SocketError err, const QString &errorString);
     void setupClientSession(RemotePeer *, UserId);
 
-    void changeUserPass(const QString &username);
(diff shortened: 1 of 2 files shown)


src/core/core.h
```diff
@@ -533,7 +533,7 @@ private slots:
     void socketError(QAbstractSocket::SocketError err, const QString &errorString);
     void setupClientSession(RemotePeer *, UserId);
 
-    void changeUserPass(const QString &username);
+    bool changeUserPass(const QString &username);
```
rsalinas: May I add a return value to a slot without breaking existing code?

sandsmark: I don't see why it would break anything (it's not like Quassel has a exported stable ABI anyways).

PR #157, opened 2015-11-12
Fix SSL detection

wengxt:
CMAKE_POSITION_INDEPENDENT_CODE will always append fPIE flag as last compiler flag when calling try_compile which makes check_cxx_source_compiles fails. Move the SSL check before set CMAKE_POSITION_INDEPENDENT_CODE and append Qt5Core_EXECUTABLE_COMPILE_FLAGS to CMAKE_CXX_FLAGS solve the problem.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -413,6 +413,20 @@ if (NOT WIN32)
     )
 endif()
 
+# Check for SSL support in Qt
+# As there's no easy way to get Qt's configuration in particular for Qt5, let's just compile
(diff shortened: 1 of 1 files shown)


sandsmark:
This seems wrong, on cmake >= 2.8.12 Qt5CoreConfigExtras.cmake doesn't add -fPIC to Qt5Core_EXECUTABLE_COMPILE_FLAGS, it only sets Qt5_POSITION_INDEPENDENT_CODE to ON.

The check is broken here now at least, with Qt 5.6.

PR #159, opened 2015-11-13
Replace build date with commit date (Alternative to #127)

romibi:
Alternative to and based on PR #127

This makes it possible to create reproducible builds.

Here using the git-commit date instead of a dummy date

See also:
https://wiki.debian.org/ReproducibleBuilds/About

Because I'm not yet very familiar with cmake and c++ please double check:
* that I didn't overlook something and _buildInfo.commitDate was really not used! (changed from uint to QString)
* that the change to CMakeLists.txt is ok...

### DIFF
--- a/src/common/quassel.cpp
+++ b/src/common/quassel.cpp
@@ -261,15 +261,16 @@ void Quassel::setupBuildInfo()
     _buildInfo.baseVersion = QUASSEL_VERSION_STRING;
     _buildInfo.generatedVersion = GIT_DESCRIBE;
 
-    // This will be imprecise for incremental builds not touching this file, but we really don't want to always recompile
-    _buildInfo.buildDate = QString("%1 %2").arg(__DATE__, __TIME__);
-
     // Check if we got a commit hash
-    if (!QString(GIT_HEAD).isEmpty())
+    if (!QString(GIT_HEAD).isEmpty()) {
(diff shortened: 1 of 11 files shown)


src/common/quassel.cpp
```diff
@@ -259,15 +259,16 @@ void Quassel::setupBuildInfo()
     _buildInfo.baseVersion = QUASSEL_VERSION_STRING;
     _buildInfo.generatedVersion = GIT_DESCRIBE;
 
-    // This will be imprecise for incremental builds not touching this file, but we really don't want to always recompile
-    _buildInfo.buildDate = QString("%1 %2").arg(__DATE__, __TIME__);
-
     // Check if we got a commit hash
-    if (!QString(GIT_HEAD).isEmpty())
-        _buildInfo.commitHash = GIT_HEAD;
+	if (!QString(GIT_HEAD).isEmpty()) {
```
sir-ragna: It seems that you used tabs here instead of spaces contrary to the rest of the project. (See also src/qtui/aboutdlg.cpp line 39)

romibi: Arg … Yes, sorry … when I did this PR i didn't notice yet that there aren't tabs …
I really should search where to set what to use per project in Visual Studio…
I'll fix it as soon as I have time! (And I better check my other PRs too)

PR #160, opened 2015-11-20
Add Core (dis-)connect Button to Main-Toolbar (+ enable Nick-Toolbar)

romibi:
Added "Connect to Core" and "Disconnect from Core" buttons to the Main-Toolbar in Quassel-Client Mode. The Network buttons were misleading because they had the same icon as the Core (dis-)connect Options in File-Menu.
Because of that I changed the Icons for  Core (dis-)connect.
**I'm pretty sure someone can come up with better ones**

~~While doing this I found the Nick-Toolbar in the Code. Enabled it and tested it. Works fine at least for me. Because the Nick-Toolbar has a lot of Buttons I set it as invisible by default.~~ (Split away to PR #243)

[https://gyazo.com/e91e2d6016303f3d302f0fee4971ab69](https://gyazo.com/e91e2d6016303f3d302f0fee4971ab69)

Added a "really disconnect" and "really connect" popup for the main buttons:
[![https://gyazo.com/fb9575c3608272999cfa946f14e37cf1]
(https://gyazo.com/fb9575c3608272999cfa946f14e37cf1.png)]
(https://gyazo.com/fb9575c3608272999cfa946f14e37cf1)

Note: If this and "Improve MainPage" #155 get merged it should be checked if every "Connect to core" icon is the same... (at the moment it wouldn't be the case)


AlD:
First of all, thanks for the PR :) I'm not really using any of the bars myself, but making them more useful if they're there anyway (and even by default) absolutely makes sense.

- Could you please split the core (dis)connect buttons change from the nick toolbar into separate pull requests?
- I think we actually shouldn't use .svgz, but .svg instead. Could you please drop the commit where you converted this?
- Can you also clean up the commit messages a bit? Ideally shouldn't contain questions like "So is it really the case that for each different looking button there needs to be an ActionType?", but instead provide answers.

  Also 9147b194 is the only commit that shows up a authored by romibi <info@romibi.ch>; not sure if you want that changed.

romibi:
Thank you very much for your Input …
Yes i can do all this … as soon as I find time to do it … (the next few weeks I'm having exams)

about svgz … the other 2 files in icons/hicolor/scalable are svgz too … that's why i changed it …
but yes i can drop that commit if svg is wanted for new icons …

And I'm still not 100% sure if those icons fit for the Core (dis-)connect Buttons

romibi:
Found time to do it now :)

PR #161, opened 2015-11-21
Fix OIdentdConfigGenerator double-printing on ssl

esainane:
Examining ~/.oidentd.conf after quassel core has been running with the --oidentd option for a while:

lport 56590 { reply "quassel" } # stanza created by Quassel
lport 56590 { reply "quassel" } # stanza created by Quassel
lport 56597 { reply "quassel" } # stanza created by Quassel
lport 56597 { reply "quassel" } # stanza created by Quassel

connected() is always connected to socketInitialized().
encrypted() is also connected to socketInitialized() when ssl is enabled.

socketInitialized() fires socketOpen(...) unconditionally, but does not fire socketInitialized(...) if we're using SSL without encryption yet - see line 464.

With socketOpen(...) connected to OIdentdConfigurationGenerator::addSocket, OIdentdConfigurationGenerator writes to .oidentd.conf twice for every connection.

Changing this to have socketInitialized(...) connected instead prevents this duplication.


AlD:
LGTM, thanks!

PR #162, opened 2015-11-25
Nick flair prefix

esainane:
This builds upon #141.

 - This uses flairs from the PREFIX capability reported by the server
 - This is configurable from the chat view settings, and is disabled by default


justjanne:
Code duplication:

A large part of your code duplicates https://github.com/quassel/quassel/blob/master/src/common/network.cpp#L141-L147

AlD:
Sorry, haven't reviewed the code, but please respond to the previous comment if you care about this PR.

RoboMWM:
Would be appreciated if this feature (which I'd imagine is a "standard" for most clients) could be added.

justjanne:
@AlD : You might want to reopen, and then use the reviewer-feature to block the merge – if the request is closed, it might end up being forgotten about.

esainane:
This should not currently be considered mergeable. There is a lot more to do in order to achieve this properly, and most of that comes under the daunting task of implementing backwards-compatible serialization.

PR #164, opened 2015-11-29
Add strict-oidentd mode

esainane:
Provides strict identity checking, useful for shared cores.

This mode is disabled by default.

When specified, this restricts authoratative ident responses to be either
the authuser account name, or custom mappings stored in the database.
When not specified, the previous behaviour (allowing users to return any ident,
including those of other users) is preserved.

No interface for editing these entries is currently exposed.

This increments the database version, adding a new table to store the
userid -> sysident mappings.


AlD:
Thanks a lot for the patches!
I like the forced-response flag, but I'm a bit undecided about the DB changes.
Could you maybe create separate pull requests for each feature, i. e. have one that only adds the flag, but doesn't require the DB changes?

esainane:
(Partly carried over from IRC)

When restoring a previous core session (eg. if the core is restarted), the usual login process is bypassed, but the core still needs to know what to tell oidentd to respond with. The CoreSession always gets a UserId, but this comes in from:
```c++
CoreSettings s;
s.corestate().toMap()["activeSessions"];
```
 which is only a list of UserId.

So, we'd either need to change the CoreSettings format to include usernames as well, or add in logic so that we'd sometimes ask the database anyway.

Is there a reason not to ask the database anyway? As far as I can tell, quassel deliberately uses the UserId everywhere and the only place a human readable auth username (rather than nick) would be required is during the initial connection. So holding on to that information in a CoreSession could be a bit unnecessary. The cost of an additional query per new connection to a server is negligible compared to all of the queries storing messages from the motd alone, for instance.

AlD:
I seem to recall that we followed up on this on IRC and Sput also commented on it, with the conclusion that passing the username along with the userId is preferred over adding the DB query.
Please correct me if I'm wrong.

digitalcircuit:
If storing and passing the username along with the userId, how would that handle username changes?  Quassel doesn't provide a manner to do so, but (*as far as I can tell*) editing the username in the database currently just works.

IrcCloud [generates a specific user-id for ident purposes](https://www.irccloud.com/abuse ).  Quassel technically has a userId already, but I'm not sure it's a good idea to expose the database layout to the public Internet.  Hashing would work, but needs a salt generated by the core in order to avoid trivial rainbow-table lookup.

Alternatively, would it work to have ```strict-oidentd``` mode only enforce if a mapping exists in the ```sysident``` table?  That would allow some users to use any ident, still let accounts be locked down, and would remove the need to look up the username via query.

(*In the future, the Quassel client interface could have a way to show what, if any, available idents are allowed, but that doesn't need accounted just yet*)

AlD:
> If storing and passing the username along with the userId, how would that handle username changes? Quassel doesn't provide a manner to do so, but (as far as I can tell) editing the username in the database currently just works.

This is simply not a supported use case right now. I'm not even sure it is a use case. Why would you want to do this except for very infrequent occasions?

> IrcCloud generates a specific user-id for ident purposes. Quassel technically has a userId already, but I'm not sure it's a good idea to expose the database layout to the public Internet. Hashing would work, but needs a salt generated by the core in order to avoid trivial rainbow-table lookup.
> 
> Alternatively, would it work to have strict-oidentd mode only enforce if a mapping exists in the sysident table? That would allow some users to use any ident, still let accounts be locked down, and would remove the need to look up the username via query.

The way I see it, our supported use cases should be:

1. Let the user choose an ident response, i. e. `strict-oidentd=false`. Use whatever the user provided in the identity settings.
2. Let the administrator choose an ident response, i. e. `strict-oidentd=true`. Use the quassel username for this. If you want to change it, the admin changes the username. I don't think using some kind of numeric ID as ident is a popular request.

Introducing a new table for user ↔ ident mapping would mean to just duplicate the already existing identities, with the difference that the user doesn't have access to them. I don't like that much.

digitalcircuit:
> This is simply not a supported use case right now. I'm not even sure it is a use case. Why would you want to do this except for very infrequent occasions?

It's infrequent, yes, but potentially worth considering for larger hosting, especially if the Quassel username becomes the ident in strict mode.

> 1. Let the user choose an ident response, i. e. strict-oidentd=false. Use whatever the user provided in the identity settings.
> 2. Let the administrator choose an ident response, i. e. strict-oidentd=true. Use the quassel username for this. If you want to change it, the admin changes the username. I don't think using some kind of numeric ID as ident is a popular request.

Is there an approach that would allow some users to have custom idents while some would be limited to their Quassel username?

E.g. in my case, I have a few names I go by on different networks, and I'd prefer to keep them separate.  However, I'd like to let others use my Quassel core without letting them have free reign over ident due to concerns of handling banning.

The table method @esainane implemented seems like a compromise offering the most control - any user can be given any number of allowed idents.  I do agree that it is more complex.

AlD:
OK, I understand this use case, but IMHO it's not worth the maintenance cost of an additional table and partially duplicating user identities.

I think the proper solution here would be to introduce user roles, so that the core can restrict access to certain attributes. For this use case e. g. it would make the ident part of the identity read-only.

digitalcircuit:
I fully agree - user roles would better serve the shared Quassel core use case and it'd be a good time to implement settings profiles, too.. though all this is certainly out of reach of my capability (for now) :)

User roles probably warrant their own discussion separate from this pull request.  Returning to the topic at hand, @esainane 's ```strict-oidentd``` mode should be fine to merge if it restricted ident replies to the Quassel username rather than using a separate table?

AlD:
Yep, without the database changes would be fine by me.

AlD:
Closing for now. Please reopen when the issues mentioned were addressed.

PR #165, opened 2015-11-30
Include dirty status in core information

esainane:
If a core was built with a dirty working tree, -dirty will be appended
to the version string. Interestingly, we already test for this in
quassel.cpp (line 292), but this was not being added in the cmake phase.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -561,7 +561,7 @@ endif()
 
 include(GetGitRevisionDescription)
 get_git_head_revision(GIT_REFSPEC GIT_HEAD)
-git_describe(GIT_DESCRIBE --long)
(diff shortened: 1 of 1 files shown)


AlD:
Thanks!

AlD:
Can you please take another look?

This seemed to introduce two bugs:
1) The arguments need to be quoted, so "--long --dirty".

2) GetGitRevisionDescription.cmake(100) ends up running this command:
git describe 441d136436d670f520ab351dd8fbf97363de713b --long --dirty

Which results in this git error:
fatal: --dirty is incompatible with commit-ishes

I suspect there's no easy way to solve this, but wanted to give you the chance to take a look before reverting it.

PR #168, opened 2015-12-19
Make "/ " use the rest of the message as a literal

esainane:
"/ /asdf" now works the same as "//asdf". This is just a convenience for
those of us used to irssi habits, which tend to die hard. :)


AlD:
Thanks!

PR #169, opened 2015-12-24
Added support for authentication providers; wrote an LDAP auth provider.

TC01:
This branch is a fairly substantial change: it adds support for selecting an auth provider (also referred to as an authenticator or auth backend in the code) to quassel during core setup, in addition to selecting a storage backend.

This allows external auth providers to be written that can do e.g. LDAP, PAM, Kerberos, etc. authentication in a relatively modular way. As an example of what can be done (and also because I suspect it's the most important use case), I took [this pull request by abustany](https://github.com/quassel/quassel/pull/4) and refactored most of his LDAP code to write a new LdapAuthenticator class. If cmake fails to find the LDAP libraries then this part of the core will not be built.

This does not break protocol compatibility; it adds optional parameters to the ClientInitAck/ClientRegistered handshake (for passing a QVariantList of auth providers and their setup information) and two optional parameters to the SetupData handshake (for passing the configuration information about auth providers). I added a new CoreFeature, ```AuthBackends = 0x0020```, that the client uses to determine whether or not it should display the authentication selection page in core setup if it's configuring an older core. On the other hand, if the newer core is being configured by an older client, then the core simply defaults to using the "SqlAuthenticator" provider as its auth backend: this is a class that is simply a wrapper for the existing auth in quassel (try to check the password against the one stored in the database). Because new client-side code was required to display the auth provider setup page in the core setup dialog, older clients cannot configure newer cores to use anything but local authentication.

Authentication now works as follows: the core auth handler calls ```Core::authenticateUser```, which calls the current authenticator's ```validateUser()``` method. The naming is now slightly odd, but this is done so that the SqlAuthenticator class can simply call ```Core::validateUser``` without doing anything more complicated. Also note that if the call to ```Core::authenticateUser``` fails, then the core auth handler will simply call ```Core::validateUser``` to check if the account only exists locally and not in the external auth provider. (See below for some more commentary on this).

In addition to what I've written above I'm more than willing to contribute documentation on my changes and how to add new auth providers to e.g. the quassel wiki, but I haven't gotten around to it yet. (I will likely also implement some kind of Kerberos auth provider too now that the framework to do so exists).

It may not be quite done (see below for issues I have yet to tackle), but I figured I should at least open a public issue / pull request about my work (rather than just leaving the odd comment in #quassel every now and then), especially now that the core functionality works.

Things that should possibly be addressed before merging the branch in its current state:

1. It's currently still possible to make local users on the command line, and as mentioned above, you can still authenticate to local users even when an auth provider is enabled. I think this is mostly fine because core setup does create at least one local user, after all (the "admin user"). But I'm not sure, possibly adding new users manually should be forbidden if the auth provider isn't SqlAuthenticator?

2. Password changes. I don't think a user using an external auth provider should be allowed to change their password, but at present there is no way of tracking whether or not a user was set up through an external auth provider. (As done in abustany's LDAP branch, I store the empty string as a LDAP user's password, but I'm not sure we can check this from inside the core?). This might require adding an entry to the table?

3. A stopgap solution to prevent password changes would be to set a flag in an auth provider called "canChangePassword" or something like that, and in all but the SqlAuthenticator, set this flag to 0. Then the password change routine just needs to check whether or not the authenticator allows password changes, and if it doesn't, fail to change passwords. But this will not work if a core has local users *and* external users.

There may be other issues that I have not thought about, too. Please let me know if you have feedback, suggestions on dealing with the above, or if you find issues, etc!


kalaksi:
Very nice, I bet many people (including me) are eager to use the modular auth backend and LDAP. 
Forgive me, but I'm currently posting from my phone and have no time to look at the code, but wanted to check out this PR right away so I'm just shooting some ideas after reading the description.

- description: "this is a class that simply", probably missing something here. The older clients can configure the core for LDAP backend, I think? 

1. - I think having local users is handy, e.g. for admin users when LDAP is unavailable.
- I don't see a reason for disabling local user creation. It's up to the admin to configure it as he wants anyways.
- I'd prefer a common linux/PAM-like order of preference where local users are tried first. IMO, it's more intuitive and external auth couldn't override admin user password.

2. - Having a column or whatever for storing the user's auth method sounds good to me

3. - Maybe expanding this solution a bit would make it more useful? Like store info about what backends the core can modify and if users auth method is something the core can't touch (atm, anything but local), then disable modifying password

TC01:
@kalaksi: thanks for the comments!

> description: "this is a class that simply", probably missing something here. The older clients can configure the core for LDAP backend, I think?

In fact, they can't; I decoupled authentication from the storage backends, or at least I started to, so there's a new page in the core setup dialog (if the client supports it) analogous to storage setup but for authentication. This required new code client side.

I can make the description a bit more clear about that, sure.

> I'd prefer a common linux/PAM-like order of preference where local users are tried first. IMO, it's more intuitive and external auth couldn't override admin user password.

Oh, you're right, that's definitely better. I can change that.

> Maybe expanding this solution a bit would make it more useful? Like store info about what backends the core can modify and if users auth method is something the core can't touch (atm, anything but local), then disable modifying password

Sure, that sounds reasonable to me. For certain auth backends it might make sense to have this be a parameter in auth setup ("can quassel change your password for you?") although no example is really coming to mind at the moment. For others, it can just be hard-coded at compile time.

Doing it this way would let us move to having multiple auth backends per core in the future too, besides just local (which we get for free) and a single external one.

TC01:
As suggested, local auth against the database is now attempted first, and if it fails quassel tries the auth provider.

I will probably take a look at modifying the database schema over the next few days.

TC01:
I have pushed more changes and made new versions of the SQL schemas.

The authenticator each user was configured with is now a string stored in the quasseluser table. This string defaults to "Database" and is checked twice: once when a user logs in (to verify they are logging in with the auth provider they were configured with) and once when a user tries to change their password (or the admin does, from the command line).

The password changing logic now works as follows:

* If the user was configured with the Database authenticator, allow them to change their password always. (This is a crutch because we always need to be capable of doing database auth but the core doesn't always have a database authenticator stored, because the core can currently only remember one auth provider at a time).
* If the user was configured with a different authenticator, and it's *not* the one authenticator the core currently knows about, do not allow password changes. This is currently a nonissue because there aren't any other auth providers yet.
* If the user was configured with an authenticator that the core currently knows about, check what ```authenticator->canChangePassword()``` returns. For ldapauthenticator, this is a hardcoded function that always returns false.

Although there are definitely future improvements that could be made (moving to having many auth providers on one core at the same time, writing new auth providers for PAM, GSSAPI, etc.), I believe *this* pull request is now ready for review. Happy new year, quassel. :)

TC01:
Okay, I'm not quite sure why that broke. :( But what I was attempting to do is pull in the latest commits to master and add the authenticator column to the migration queries, too.

AlD:
Instead of merging the updated master branch, try rebasing onto the master branch.
If you have the official repo added as `origin`, you can do so via `git rebase origin/master`.

The output will also point you at some whitespace errors. See https://github.com/AlD/quassel/tree/jhuacm/auth-provider for an example of the rebased branch.

As for the PR, unfortunately I haven't yet had a chance to try or review it, but thanks a lot for your contribution. This will allow frequently requested features to be implemented.

TC01:
Thanks for the advice; I'm going to close this PR and re-open it after having done a git rebase to hopefully clean that up. :)

kalaksi:
> Authentication now works as follows: the core auth handler calls Core::authenticateUser, which calls the current authenticator's validateUser() method. The naming is now slightly odd, but this is done so that the SqlAuthenticator class can simply call Core::validateUser without doing anything more complicated. Also note that if the call to Core::authenticateUser fails, then the core auth handler will simply call Core::validateUser to check if the account only exists locally and not in the external auth provider. (See below for some more commentary on this).

Didn't you change it in the previous PR so that the local user's are always checked first, similar to e.g. common linux setup with LDAP?

Also, could you provide some short explanation / command reference on how to setup the core from the command line and use LDAP? Right after installation, I try to do --add-user to add the local admin-user and it says that "Unable to add user". Nothing on the debug log.

PR #170, opened 2016-01-04
Added support for authentication providers; wrote an LDAP auth provider.

TC01:
(Sorry for re-opening this and closing [the other one](https://github.com/quassel/quassel/pull/169), but I managed to break the previous branch through a couple of silly operations and wanted to clean things up, so here we are, after @AlD  suggested I do a rebase. Anyway, nothing has changed really but here's the description from before):

This branch is a fairly substantial change: it adds support for selecting an auth provider (also referred to as an authenticator or auth backend in the code) to quassel during core setup, in addition to selecting a storage backend.

This allows external auth providers to be written that can do e.g. LDAP, PAM, Kerberos, etc. authentication in a relatively modular way. As an example of what can be done (and also because I suspect it's the most important use case), I took [this pull request by abustany](https://github.com/quassel/quassel/pull/4) and refactored most of his LDAP code to write a new LdapAuthenticator class. If cmake fails to find the LDAP libraries then this part of the core will not be built.

This does not break protocol compatibility; it adds optional parameters to the ClientInitAck/ClientRegistered handshake (for passing a QVariantList of auth providers and their setup information) and two optional parameters to the SetupData handshake (for passing the configuration information about auth providers). I added a new CoreFeature, `AuthBackends = 0x0020`, that the client uses to determine whether or not it should display the authentication selection page in core setup if it's configuring an older core. On the other hand, if the newer core is being configured by an older client, then the core simply defaults to using the "SqlAuthenticator" provider as its auth backend: this is a class that is simply a wrapper for the existing auth in quassel (try to check the password against the one stored in the database). Because new client-side code was required to display the auth provider setup page in the core setup dialog, older clients cannot configure newer cores to use anything but local authentication.

Authentication now works as follows: the core auth handler first calls `Core::validateUser` as before, attempting to do local authentication. If local authentication fails and an external provider is configured, the auth handler calls  `Core::authenticateUser`, which calls the current authenticator's `validateUser()` method. The naming is now slightly odd, but this is done so that the SqlAuthenticator class can simply call `Core::validateUser` without doing anything more complicated. 

In addition to what I've written above I'm more than willing to contribute documentation on my changes and how to add new auth providers to e.g. the quassel wiki, but I haven't gotten around to it yet. (I will likely also implement some kind of Kerberos auth provider too now that the framework to do so exists).

Other comments on password changing:

The authenticator each user was configured with is now a string stored in the quasseluser table. This string defaults to "Database" and is checked twice: once when a user logs in (to verify they are logging in with the auth provider they were configured with) and once when a user tries to change their password (or the admin does, from the command line).

The password changing logic now works as follows:
- If the user was configured with the Database authenticator, allow them to change their password always. (This is a crutch because we always need to be capable of doing database auth but the core doesn't always have a database authenticator stored, because the core can currently only remember one auth provider at a time).
- If the user was configured with a different authenticator, and it's _not_ the one authenticator the core currently knows about, do not allow password changes. This is currently a nonissue because there aren't any other auth providers yet.
- If the user was configured with an authenticator that the core currently knows about, check what `authenticator->canChangePassword()` returns. For ldapauthenticator, this is a hardcoded function that always returns false.

There may be other issues that I have not thought about, too. Please let me know if you have feedback, suggestions on dealing with the above, or if you find issues, etc!


Nevcairiel:
For easier review, maybe you could squash a bunch of these commits into logical units, instead of partial work units how it happened to be developed?

Its just a whole bunch of commits, many of those fix issues in earlier commits etc. It would be neat to have a list of X commits, each adding a separate part (ie. the auth provider abstraction, the LDAP implementation, whichever other distinct components there may be)

TC01:
Sure, I'll take care of that tonight. I think the main logical units of this are:

* Initial auth provider abstraction / core setup dialog (this will be most of the commits, since I was admittedly committing haphazardly until I was able to compile).
* LDAP authentication support.
* Backwards compatibility to older clients and cores.
* Changing authentication to be tried in the order local -> external.
* Adding the authenticator column to the database / checks to prevent password changing.

Does this sound like a reasonable way to break this up?

sir-ragna:
> This allows external auth providers to be written that can do e.g. LDAP, PAM, Kerberos, etc. authentication in a relatively modular way

With modular, do you mean at runtime (versus compile time)? Will this open way to create a backend option in the configuration file? 

I really appreciate your work on this. I really like the direction this is going.

TC01:
@Nevcairiel: rebased into ten commits. I think it's much cleaner now; an argument could be made that c2b02cf and 79a779a are still superfluous, but they didn't really belong in the directly previous commit (unless the entire last four commits should be collapsed into one?)

I'll be happy to adjust things further, but each commit should "work"; it should be able to be compiled, ran, and tested if necessary.

@sir-ragna: no, unfortunately, not yet, it's only modular in the sense that authentication is now decoupled from storage. But future work could be done to make the core load Authenticator classes at runtime and add them when gathering a list of available authenticators. I don't think that would be too difficult now.

And thanks. :)

Nevcairiel:
Definitely looks much better, thanks!

kalaksi:
Wanted to start testing this and noticed that cmake warns that:

> CMakeLists.txt:463 (if)
> closes on the line
> CMakeLists.txt:465 (endif)
> with mis-matching arguments.

It indeed does and it's on the ldap-related code. It's mainly cosmetic but you might want to fix it.

TC01:
Oh, I see, whoops! Going to fix and rewrite the commit,

kalaksi:
>Authentication now works as follows: the core auth handler calls Core::authenticateUser, which calls the current authenticator's validateUser() method. The naming is now slightly odd, but this is done so that the SqlAuthenticator class can simply call Core::validateUser without doing anything more complicated. Also note that if the call to Core::authenticateUser fails, then the core auth handler will simply call Core::validateUser to check if the account only exists locally and not in the external auth provider. (See below for some more commentary on this).

Didn't you change it in the previous PR so that the local user's are always checked first, similar to e.g. common linux setup with LDAP?

Also, could you provide some short explanation / command reference on how to setup the core from the command line and use LDAP? Right after installation, I try to do --add-user to add the local admin-user and it says that "Unable to add user". Nothing on the debug log.

kalaksi:
Also, it doesn't help if I select the DB first after which it automatically asks for new user.

TC01:
> Didn't you change it in the previous PR so that the local user's are always checked first, similar to e.g. common linux setup with LDAP?

Yes; whoops, forgot to update that part of the text. Fixed.

> Also, could you provide some short explanation / command reference on how to setup the core from the command line and use LDAP? Right after installation, I try to do --add-user to add the local admin-user and it says that "Unable to add user". Nothing on the debug log.

Oh... I forgot to write this part! Command line setup slipped my mind entirely, my bad. :(

What I'll do is (over the course of the next week, I expect) add a ```--select-authenticator`` option that, analogous to the storage select option, will let you pick an auth provider. I'll also investigate why creating users would fail if an authenticator isn't currently configured.

TC01:
@kalaksi: I just pushed a new commit that implements the --select-authenticator switch. Using this I was able to configure the core for LDAP and Database auth using something like this:

```
quasselcore -c confdir --select-backend=SQLite --select-authenticator={Database,LDAP}
```

Let me know how it goes. :)

TC01:
Interesting. I just tried a build on Unix without the LDAP libraries to link against, and it appeared to work fine, so I'm not sure why that failed.

kalaksi:
Compiled fine for me and it seems to be working. I have now tested that it correctly binds with given account and password, finds the correct branch and applies the search filter.

I seem to be getting SQL error in the debug log with error number -1. Those seem to be related to searching for userid or quasseluser, such as:

...
>   last Query: SELECT userid, password, hashversionFROM quasseluserWHERE username = :username
>  executed Query: SELECT userid, password, hashversionFROM quasseluserWHERE username = ? 
>    bound Values: :username='kalaksi' 
>    Error Number: -1 
>   Error Message:   
>  Driver Message:  
>      DB Message:  
>      last Query: SELECT userid, password, hashversionFROM quasseluserWHERE username = :username 
>  executed Query: SELECT userid, password, hashversionFROM quasseluserWHERE username = ? 
>    bound Values: :username='kalaksi' 
>    Error Number: -1 
>   Error Message:   
>  Driver Message:  
>      DB Message:  
>      last Query: SELECT authenticatorFROM quasseluserWHERE userid = :userid 
>  executed Query: SELECT authenticatorFROM quasseluserWHERE userid = ? 
>    bound Values: :userid=2 
>    Error Number: -1 
...

kalaksi:
Meant to add that maybe those are just debug messages since I was running it in debug-mode.
Anyhow, I did some migrating by hand and as you probably noticed, I commented on a possible mistake in the SQL.

TC01:
@kalaksi; sorry, not seeing your comment (and github may have eaten it when I rebased the commits against current master); what do you think the issue is?

TC01:
Nevermind, I saw your feedback. That could be the issue, yeah-- let me take a look.

TC01:
@kalaksi: fixed that; possibly that will stop those errors (if you were using the postgresql bindings when you got those error messages, anyway)?

kalaksi:
Yes, I am, but I'm starting to be pretty sure those -1 errors are just debug messages since I got plenty of them and with sane parameters and without any actual error message.

Also, I'm pretty sure that the SQL would have produced errors in some case (indeed, when trying to handle authenticator-field). In my case, I faced the issue indirectly by trying to modify DB manually because I migrated data manually from the ldap-wrapper-sqlite-thing. Adding authenticator-column as unique was the first way I tried to modify the DB and noticed that ofc the "authenticator"-field can't be unique, so I instead added the authenticator-columns properly as in other sql-files.

TC01:
I reread the build output and figured out why it's failing on MSVC; we need to link against ```winldap.h``` and not ```ldap.h``` if we're on Windows and building with MSVC. I'll push a commit with that fix and see if it clears things up momentarily.

@kalaksi: well, let me know if you run into any other issues. :)

TC01:
Alright, so, per discussion in IRC I removed the special cases for winldap.h in both the code and the cmake script, so we should *attempt* to link against openldap when compiling with MSVC too. If it turns out to be desirable to try and add some kind of compatibility shim for winldap.h instead that can be done later (and ideally by someone familiar with it).

I should also disclose that this code uses a new CoreFeature to tell clients whether or not to display the auth selection page in core setup; currently that's 0x0020 but since this conflicts with @esainane's branch I'll push a commit moving that to 0x0040 (also as discussed in IRC).

TC01:
Done. Once again, I believe this branch is now ready for review.

If I haven't mentioned this already somewhere, I'll probably have a go at writing a Kerberos auth provider (that takes a Kerberos realm as a config option and tries to get tickets for username@realm, or something) in the near future but won't be adding that to *this* pull request. That can be a separate one if/when this features gets merged into the core.

TC01:
Rebased against latest quassel master, for anyone who happens to be running my branch and wants all the latest fixes in master.

Any word on whether or not this will be reviewed? I'm more than happy to keep working on improving it (I'd really like to see this feature get into Quassel, ideally in time for 0.12.3), but at the moment I'm not sure what I should be doing.

digitalcircuit:
Quassel 0.12.3 was already released, and 0.12.4 is out by now.  I don't know for sure, but I suspect the project maintainers wished to put this into a full point release instead of a maintenance release, similar to IRCv3 support.

I'd suggest stopping by **```Freenode/#quassel```** and asking for feedback - ```Sput``` has been around a bit more often the past few days.  You can also mention ```@Sput42``` in comments here, too.

This is a lot of work for a feature useful to both large personal servers and corporations, and it'd be a shame to lose it.

src/core/SQL/PostgreSQL/18/setup_000_quasseluser.sql
```diff
@@ -2,5 +2,6 @@ CREATE TABLE quasseluser (
        userid serial NOT NULL PRIMARY KEY,
        username varchar(64) UNIQUE NOT NULL,
        password TEXT NOT NULL,
-       hashversion integer NOT NULL DEFAULT 0
+       hashversion integer NOT NULL DEFAULT 0,
+       authenticator varchar(64) NOT NULL DEFAULT "Database"
```
AlD: Would it make sense to use an Enum here?

CMakeLists.txt
```diff
@@ -116,6 +116,10 @@ if (LINK_EXTRA)
 endif()
 
 
+# LDAP Authentication (and other authentication backends).
+####################################################################
+option(WITH_LDAP     "Enable LDAP authentication support if present on system"  ON)
```
AlD: Since this is not LDAP specific, please use a different string throughout the code. How about something to the effect of s/LDAP/AUTHENTICATORS/?

I realize this may be tedious, especially for filenames, but if we don't change this now, it never will ;)

AlD: Oh, not sure if I misunderstood the "(and other authentication backends)" part and I haven't read the full patch yet. Does WITH_LDAP enable just LDAP support or does it support authentication backends support, but LDAP is the only backend?

In the latter case, my original comment does apply. In any case, please clarify the comment :)

cmake/FindLdap.cmake
```diff
@@ -0,0 +1,44 @@
...
+endif(LDAP_INCLUDE_DIR AND LDAP_LIBRARIES)
+
+#if(UNIX)
+# Attempt to link against ldap.h regardless of platform!
+FIND_PATH(LDAP_INCLUDE_DIR ldap.h)
+FIND_LIBRARY(LDAP_LIBRARIES NAMES ldap)
+FIND_LIBRARY(LBER_LIBRARIES NAMES lber)
+
+#else(UNIX)
+#   FIND_PATH(LDAP_INCLUDE_DIR winldap.h)
```
AlD: Why is this commented? Either remove it or explain in a comment, why that's here.

src/client/clientauthhandler.cpp
```diff
@@ -303,6 +303,7 @@ void ClientAuthHandler::handle(const ClientRegistered &msg)
 {
     _coreConfigured = msg.coreConfigured;
     _backendInfo = msg.backendInfo;
+    _authBackendInfo = msg.authBackendInfo;
```
AlD: Isn't "authBackend" equivalent to what's up until here has been referred to as "authenticator"? Can we use either "authenticator" or "authBackend" everywhere, but not both?

src/common/main.cpp
```diff
@@ -157,6 +157,7 @@ int main(int argc, char **argv)
 #endif
     cliParser->addOption("logfile", 'l', "Log to a file", "path");
     cliParser->addOption("select-backend", 0, "Switch storage backend (migrating data if possible)", "backendidentifier");
+	cliParser->addOption("select-authenticator", 0, "Switch auth backend", "authidentifier");
```
AlD: whitespace

AlD: "Select authentication backend(s)"

Using "backend" in the doc string here instead of "authenticator" makes sense I think.

src/common/protocols/datastream/datastreampeer.cpp
```diff
@@ -201,6 +202,11 @@ void DataStreamPeer::dispatch(const SetupData &msg)
     map["Backend"] = msg.backend;
     map["ConnectionProperties"] = msg.setupData;
 
+    // Auth backend properties.
+    // XXX: make these optional using core features.
```
AlD: Please use TODO (missing features) or FIXME (known issues) instead of XXX.

src/common/protocols/legacy/legacypeer.cpp
```diff
@@ -264,6 +265,11 @@ void LegacyPeer::dispatch(const SetupData &msg)
     map["AdminPasswd"] = msg.adminPassword;
     map["Backend"] = msg.backend;
     map["ConnectionProperties"] = msg.setupData;
+    
+    // Auth backend properties.
+    // XXX: make these optional using core features.
```
AlD: see above

src/core/authenticator.cpp
```diff
@@ -0,0 +1,26 @@
...
+ *   Free Software Foundation, Inc.,                                       *
+ *   51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.         *
+ ***************************************************************************/
+
+#include "authenticator.h"
+
+Authenticator::Authenticator(QObject *parent)
+    : QObject(parent)
+{
+}
```
AlD: add newline

src/core/authenticator.h
```diff
@@ -0,0 +1,95 @@
...
+    //! Check if the authenticator type is available.
+    /** An authenticator subclass should return true if it can be successfully used, i.e. if all
+     *  prerequisites are in place.
+     * \return True if and only if the authenticator class can be successfully used.
+     */
+    virtual bool isAvailable() const = 0;
+
+    //! Returns the display name of the authenticator backend
+    /** \return A string that can be used by the client to name the authenticator backend */
+    virtual QString displayName() const = 0;
```
AlD: "displayName" is really misleading. I know this comes from storagebackend, but I think it should be called "backendId" or something.

AlD:
Sorry, ran out of time while reviewing this, so didn't make it to the bottom and I'm not sure if I'll find the time to continue tomorrow or in the next days.
I hope the comments so far give you something to work on.

Modelling this after the storagebackends was a good idea for consistency, although it shows in several places that it's not a perfect fit.

I think we can make it work, though.

Huge thanks for the work so far!

CMakeLists.txt
TC01: Oh, I see. WITH_LDAP is supposed to enable *just* WITH_LDAP support. If it's disabled, the LDAP auth backend is not built, but the auth backends system as a whole is still compiled in. So yeah, the comment should be clarified (something like "Authentication backend options (currently, that's just LDAP, but more can be added here)" instead).

src/core/SQL/PostgreSQL/18/setup_000_quasseluser.sql
TC01: I'm not sure. The main problem I see with that is that it's theoretically possible for me to setup LDAP on a core with LDAP support compiled in and then upgrade to a core that doesn't have LDAP support compiled in, for whatever reason. At the moment this is not a huge issue because there are only two auth backends, but in the future we could run into a situation where we have, say:

* Storage
* LDAP
* Kerberos

And then suddenly, after rebuilding the core, we have:

* Storage
* Kerberos

So if the enum is generated at core runtime based on the number of loaded auth backends, things will get confusing. Thus, auth backends would probably need a hardcoded ID (e.g. LDAP is always 1, Kerberos is always 2, etc). But we already have a hardcoded identifier for each auth backend-- their "display name" (though as you mention below, "ID" would be a better descriptor). So it seems to me the only reason to use an integer ID instead of the string ID is to save space in the database-- is this worth it? (It may very well be, I'm not sure).

Unless by "enum" you meant creating a new table in the DB that serves as a lookup for the integer auth backend IDs. I guess that would work, because it would solve the core upgrade problem; those tables might not be consistent from system to system but behavior wouldn't change by adding and disabling auth backends at compile time.

src/client/clientauthhandler.cpp
TC01: Originally I was using the term "authenticator" as the object that does authentication for a user or users. I then started using the term "authBackend" after I'd already written the authenticator classes and was trying to add them to the protocol; so I copied the "backend" and "backendInfo" objects in the messages sent by the protocol and renamed them authBackend and authBackendInfo analogously. Things sort of spiralled out of control from there.

I think I prefer the term authBackend, but authenticator is used in more places in the code and I used authenticator in the database. I know that at least @kalaksi (I think?) is using this branch of the core (and its database schema), so it would probably be better to stick with authenticator-- I'll standardize on that.

TC01:
I've pushed a commit fixing what was pointed out above (minus the database enum thing, which I am not sure about-- see my comments on the diff).

* Standardized on "authenticator" rather than "authBackend"
* Removed XXXs and replaced with TODO or FIXME comments appropriately
* Fixed a couple whitespace errors
* Cleared up some comments in the ldap cmake
* In the authenticator code, renamed displayName to backendId
* Added missing newline at the end of authenticator.cpp

I think this should be ready for continued review.

TC01:
I have a rebase commit that brings this branch up to current master, but I'd need to force push and I think that will wipe out @AlD's comments on the diff above, and I'm hesitant to do that at the moment. (Also I think I messed up the manual merge of the schema changes, as I needed to reapply some of my database changes in another commit afterwards). 

But, I tested authentication (both normal and LDAP) and it still works!

Let me know if/when I should push that here.

TC01:
I'll again fix these merge conflicts (ideally by sometime this weekend) and rebase up to master, but other than that, is there anything that is holding this up? What's the current "more work" that is believed needed?

I would really like to see this feature, or a modified version of it, land in 0.13.

Sput42:
The "needs more work" was set before you did your recent changes, mostly the db schema stuff again. I'll make sure to review this before we do the next release (and of course I welcome others to review too!)

digitalcircuit:
@TC01 Is there a "simple" way to get the minimum of LDAP server necessary to test this?  Time permitting, I'll try testing if it breaks existing setups without LDAP, but a way to verify with LDAP would be useful, too.

Sput42:
Rebased on top of current master and added bunch of new stuff on top of this, including a rework and modernization of backend handling and core configuration.

jaakristioja:
I'm getting the following SQL errors with a05e435 which also seem to result in low performance logging in (at least the first time). I don't know if these are relevant or not. Attaching the [core log](https://github.com/quassel/quassel/files/565117/quassel-core-stdout.txt).

src/core/sql.qrc
```diff
@@ -1,260 +1,264 @@
 <!DOCTYPE RCC><RCC version="1.0">
 <qresource>
```
digitalcircuit: Though it's not required, it'd make it much easier to read ```git diff``` if the list is kept alphabetized, e.g. [how it currently is in ```master```]( https://github.com/quassel/quassel/blob/5221b0b4d044d214539a1932a7a2eefbe88e4a66/src/core/sql.qrc ).

In the course of troubleshooting what went wrong, [I've made a sorted copy of ```sql.qrc``` here](https://dl.dropbox.com/u/839888/Hosting/Utilities/Quassel/Development/pr/others/jhuacm/auth_providers_rebase/sql.qrc-alphabetical.txt ) - feel free to use it.

TC01: Included with a fix for at least one of your reported issues. :)

TC01:
I believe I've fixed the Postgres issues above; going to just copy-paste the last commit message:

"It seems that at some point the DEFAULT option for a varchar()
column has become sensitive to whether or not single or double quotation
marks are used. In particular when ""s were used, core setup failed
with the message 'ERROR:  column "Database" does not exist'. Changing
to single quotation marks (''s) resolved the issue."

The schema upgrade error seems to have been caused by the same thing: the upgrade went from 17 -> 20, failed on 20 because of the "" vs '' thing, and then failed when restarting because the upgrade from 17 -> 18 had already been done but the core didn't think so. Ideally the core should have known it got to version 19 before trying 19 -> 20. :/

@jaakristioja: neither @digitalcircuit or I saw the SQL errors in your log when testing. (I've additionally tested connecting to LDAP too). What version of Qt are you linking Quassel against? When googling the issue I found a [few StackOverflow references](https://stackoverflow.com/questions/24684550/qt-5-with-sqlite-bindvalue-results-in-parameter-count-mismatch-error) to similar problems, but like I said I've never seen them myself.

I have seen the "core doesn't respond to SIGTERM/SIGINT" issue as a result of this branch, and at this time have no idea why. It seems to only occur after first setting up a database, as you said.

jaakristioja:
@TC01 I built Quassel from git on a fairly normal Ubuntu Xenial machine. Looking at the binaries, it seems to link against "Qt 5.5.1 (x86_64-little_endian-lp64 shared (dynamic) release build; by GCC 5.4.0 20160609)" which seems to correspond to what [packages.ubuntu.com](http://packages.ubuntu.com/) lists for "qt5" or "libqt5" package searches on Xenial.

TC01:
@jaakristioja : do you get the same errors on master, by any chance? (They don't immediately seem to be related to any of the auth changes here). Also, do you still get them from 4cd7615?

wrouesnel:
So a problem with deploying this via ansible or docker or other automated means - the config files generated on first run serialize as a form like `AuthenticationSettings=QVariant("<a bunch of octets>")`.

It would help a lot for the (I suspect fairly common) use case of wanting to run a quasselcore server if the config files serialized more nicely so they could be templated as text.

digitalcircuit:
@wrouesnel I fully agree that human-readable/templating-compatible configuration files would be wonderful.  Currently, to automate with configuration management, I'm [having to make use of Bash `here documents`](https://github.com/digitalcircuit/salt-box-quassel/blob/master/salt/files/quassel/quassel-setup.sh ) - a brittle, inelegant approach.

That said, I personally think modifying Quassel's configuration format should be addressed in a follow-up pull request.  Quassel has used serialized `QVariant` and other data within its configuration files for many releases, so this pull request isn't doing anything out of the ordinary.  Given the size and scope of the LDAP authenticator, getting the current code polished up and merged might be the best use of resources.

As far as I know, there's only [one reproducible crash bug](https://github.com/quassel/quassel/pull/170#discussion_r87688657 ) and [one UI glitch](https://github.com/quassel/quassel/pull/170#pullrequestreview-10222726 ) remaining to fix, asides whatever might come up when the project maintainers finish their review.

TC01:
@wrouesnel: I agree with you that it's an issue, but unfortunately, it's how quassel configuration files are currently written. The storage server configuration looks like that as well, sadly, so I figured I would stick with that convention for at least the time being.

(Granted, maybe there's something to be said for taking this opportunity to *also* fix the configuration file syntax before this change enters master-- we've already bolted improvements to the storage (and authentication) backend selection code onto this PR. But I'm 

For the mean time, I started working on [a tool to read](https://gitlab.com/TC01/quasselconf/) quassel configuration files and parse them into something more human readable. When I get around to it, it'll also support writing them too, by taking the human readable form and turning it into the quassel-readable form.

src/qtui/coreconfigwizard.cpp
```diff
@@ -221,121 +403,41 @@ int StorageSelectionPage::nextId() const
...
-    ui.description->setText(_backends[backend].toMap()["Description"].toString());
+    ui.description->setText(_backendProperties[index]["Description"].toString());
 
-    if (_connectionBox) {
-        layout()->removeWidget(_connectionBox);
-        _connectionBox->deleteLater();
-        _connectionBox = 0;
+    if (_fieldBox) {
+        layout()->removeWidget(_fieldBox);
+        _fieldBox->deleteLater();
```
TC01: > Note: I encountered this crash on the storage selection screen. As I don't have the LDAP backend, I could not verify the crash happens here, too, but as the code seems identical, it's possible.

I can confirm that this also happens on the authentication selection page (on Fedora 25, Qt 5.7.1; didn't check other platforms); after toggling back and forth from LDAP to Database a few times in the dropdown, my client segfaulted.

I barely understood the core setup dialog code *before* it got rewritten. :( @Sput42, when you have time, any chance you can investigate this crash?

TC01:
I received a patch from @Tuetuopay fixing a case sensitivity issue with LDAP that I've merged into this PR:

> This patches a case sensitivity issue in the core.

> LDAP itself is case insensitive, but the LDAP module of the core will keep the case of the usename when creating a local Quassel account. Thus, if you login with the same username but using a different casing, a new internal Quassel account will be creating, resulting in an apparent settings loss (networks, ...).

> The patch converts the username to lowercase before handing it to the core.

TC01:
I have force pushed another (hopefully final) rebase; the PR now uses the new SQL schema style and has no conflicts with master.

I tested basic functionality, but more thorough testing would (as always) be greatly appreciated.

Tuetuopay:
For what it is worth : I have been running https://github.com/jhuacm/quassel/commit/3b842b1415d96a9d928f803f90465d7bf88d6b44 for a few months without a single problem. LDAP works reliably with 5-10 daily users on CentOS machine.
I will upgrade as soon as I have time to this new one :)

Tuetuopay:
Okay so I just upgraded from https://github.com/jhuacm/quassel/commit/3b842b1415d96a9d928f803f90465d7bf88d6b44 to https://github.com/quassel/quassel/pull/170/commits/d9266496ec207e263b431dcfb6b3d6aab60adca1 and everything broke.

Quassel core cannot even start anymore with the following in the logs :

    quasselcore[1773]: 2017-05-31 23:38:02 Warning: Switching Postgres to legacy mode. (set standard conforming strings to off)
    quasselcore[1773]: Installed Schema (version 20) is not up to date. Upgrading to version 22...

And then it crashes with status 1. But it actually broke and migrated some stuff as rolling back to the previous version throws an insane amount of SQL errors :

    quasselcore[1619]: 2017-05-31 23:36:54 Error: unhandled Error in QSqlQuery!
    quasselcore[1619]: 2017-05-31 23:36:54 Error:                   last Query:
    quasselcore[1619]: EXECUTE quassel_insert_message ('2017-5-31 21:36:54.754', 46, 1024, 0, 1, 'Disconnecting. (Core Shutdown)')
    quasselcore[1619]: 2017-05-31 23:36:54 Error:               executed Query:
    quasselcore[1619]: 2017-05-31 23:36:54 Error:                 bound Values:
    quasselcore[1619]: 2017-05-31 23:36:54 Error:                 Error Number: -1
    quasselcore[1619]: 2017-05-31 23:36:54 Error:                Error Message: ERROR:  column buffer.lastmsgid does not exist
    quasselcore[1619]: LIGNE 4 :             AND buffer.lastmsgid < new.messageid
    quasselcore[1619]: ^
    quasselcore[1619]: [26B blob data]
    quasselcore[1619]: SET lastmsgid = new.messageid
    quasselcore[1619]: WHERE buffer.bufferid = new.bufferid
    quasselcore[1619]: AND buffer.lastmsgid < new.messageid
    quasselcore[1619]: CONTEXTE : PL/pgSQL function backlog_lastmsgid_update() line 3 at SQL statement
    quasselcore[1619]: (42703) QPSQL: Unable to create query
    quasselcore[1619]: 2017-05-31 23:36:54 Error:               Driver Message: QPSQL: Unable to create query
    quasselcore[1619]: 2017-05-31 23:36:54 Error:                   DB Message: ERROR:  column buffer.lastmsgid does not exist
    quasselcore[1619]: LIGNE 4 :             AND buffer.lastmsgid < new.messageid
    quasselcore[1619]: ^
    quasselcore[1619]: [26B blob data]
    quasselcore[1619]: SET lastmsgid = new.messageid
    quasselcore[1619]: WHERE buffer.bufferid = new.bufferid
    quasselcore[1619]: AND buffer.lastmsgid < new.messageid
    quasselcore[1619]: CONTEXTE : PL/pgSQL function backlog_lastmsgid_update() line 3 at SQL statement
    quasselcore[1619]: (42703)

I'll try digging into this

TC01:
@Tuetuopay; unfortunately, directly upgrading from the previous LDAP branch won't work, because the SQL versions have changed under the hood. Namely the revision of the schema used for the authenticator changes in the previous PR was used in quassel master for something else.

See this conversation I just had on IRC with @digitalcircuit:

```
<TC01> digitalcircuit: am I crazy, or does it make sense that https://github.com/quassel/quassel/pull/170#issuecomment-305326911 would break horribly? (i.e. upgrading from previous version of LDAP branch to the new, rebased version)
<TC01> Since the schema versions between the old branch and new branch aren't consistent...
<TC01> s/new branch/new branch and current quassel master/
<TC01> my instinct is to say it makes sense that that didn't work, but I'd like someone to confirm before I say that on the thread :)
<digitalcircuit> TC01: Yes, it will (sadly) break.  However...
<digitalcircuit> It could be possible to manually revert the database changes you first introduced, downgrade the stored database schema version number, then let it upgrade again.
<digitalcircuit> TC01: Alternatively, a hybrid approach would be to manually increment the schema version number, manually apply all the other upgrade scripts, and really hope nothing's inconsistent now.
<digitalcircuit> (I would not do that latter thing if possible...)
<TC01> digitalcircuit: hm. :( well, thanks for confirming
<digitalcircuit> Well.. hm.
<digitalcircuit> You could also lower the schema version number to what it was before your original PR, let it partially upgrade, then manually bump the schema version number so it doesn't try to re-apply the LDAP upgrade (this would not be a valid upgrade test, though, it just restores database usability).
<digitalcircuit> If there's no backups.. this gets more complicated.
```

Tuetuopay:
Thanks for the quick update. Well at least it is confirmed the database update is broken !

wrouesnel:
Well at least with a clean database this is working great - just spent some time baking this into a self-configuring docker container: https://github.com/wrouesnel/docker.quassel-ldap (basically just some Python magic to let you specify config via env vars) and its working great on my home server.

TC01:
As requested by Zhoriel a few weeks ago on IRC, I've added an info message to the LDAP connection that prints the address and port of the server being connected to, in the hopes of making debugging a bit easier.

@wrouesnel: great! Glad to hear it's working well for you.

@Tuetuopay: just to be clear, database updates from master should work just fine-- it's only automatic database updates from older versions of the PR that sadly won't work.

digitalcircuit:
# Overview - 2017-8-18
## Summary
* SQLite and Postgres setup/migration success
* **New setup crashing bug** during storage selection
* Setup wizard widget layout sometimes squashed
* LDAP untested

## Steps for a successful test
1.  Initialize or upgrade Quassel core (*completing first run wizard if needed*)
2.  Verify networks intact
3.  Connect to a network, verify message sends
4.  Restart the core
5.  Verify networks intact

*Note: steps 4-5 were skipped for most tests.*

## Notes
* **LDAP functionality was not tested** as I do not have an easily-used LDAP setup
* There's still a [**new crashing bug in initial setup flow**, see review comment](https://github.com/quassel/quassel/pull/170#discussion_r87688657 )
* When choosing storage backend, [the configuration widgets are squashed, see review comment](https://github.com/quassel/quassel/pull/170#pullrequestreview-10222726 )
* Windows has a release-blocking bug where OpenSSL is compiled with MSVC 2013 while Quassel is compiled with MSVC 2015
  * For the sake of testing, I manually installed the MSVC 2013 runtime
  * `TheOneRing` is currently investigating using KDE's Craft system to fix this
* I did not test Postgres on Windows as I don't know how to easily set it up
* Restarting core on Windows did not always automatically reconnect networks
* Upgrades cause the new buffer search bar to show
  * This is unrelated to this pull request and should be fixed in a separate branch

# Test results - 🗹 pass
## Linux - 🗹 pass
### Fresh configuration - 🗹 pass
*Quassel core from this PR set up without any initial data*

Test conditions  | Result  | Remarks
-----------------|---------|-------------
SQLite, new core, new client | 🗹 *pass* |
SQLite, new core, old client | 🗹 *pass* |
SQLite, old core, new client | 🗹 *pass* |
SQLite, new monolithic | 🗹 *pass* |
Postgres, new core, new client | 🗹 *pass* |
Postgres, new core, old client | 🗹 *pass* |
Postgres, old core, new client | 🗹 *pass* |

### Migration - 🗹 pass
*Quassel core from this PR set up without any initial data, then migrated*

Test conditions  | Result  | Remarks
-----------------|---------|-------------
SQLite → Postgres, new core, new client | 🗹 *pass* |

### Upgrading existing - 🗹 pass
*Quassel core 0.12.2 set up (0.13-git-master for monolithic), then upgraded to this PR*

Test conditions  | Result  | Remarks
-----------------|---------|-------------
SQLite, old → new core | 🗹 *pass* |
SQLite, old monolithic → new monolithic | 🗹 *pass* |
Postgres, old → new core | 🗹 *pass* |

## Windows - 🗹 pass
### Fresh configuration - 🗹 pass
*Quassel core from this PR set up without any initial data*

Test conditions  | Result  | Remarks
-----------------|---------|-------------
SQLite, new core, new client | 🗹 *pass* |
SQLite, new core, old client | 🗹 *pass* |
SQLite, old core, new client | 🗹 *pass* |
SQLite, new monolithic | 🗹 *pass* |

### Migration - ? N/A
*Without Postgres on Windows, migration could not be tested*

### Upgrading existing - 🗹 pass
*Quassel core 0.12.4 set up, then upgraded to this PR*

Test conditions  | Result  | Remarks
-----------------|---------|-------------
SQLite, old → new core | 🗹 *pass* |
SQLite, old monolithic → new monolithic | 🗹 *pass* |

## Failed cases
None found!

Sput42:
Thanks for the very thorough testing! I'll look into the remaining issues, then I think this can be merged soon.

Sput42:
Fixed the crash in 06cf760

src/qtui/coreconfigwizard.cpp
Sput42: Yeah, it's always a good idea to set a pointer to nullptr when deleting the pointee...

Sput42:
Layout issues in CoreConfigWizard were fixed in 05249bc.

wrouesnel:
Is this still going to get merged? I've been using this branch consistently and it's excellent.

digitalcircuit:
@wrouesnel It's merged!

> Sput42 closed this in 05249bc on Aug 30

However, it was done via a local rebase/fix-up, so Github interpreted it as closing without merging.  You can check the [Quassel commit history](https://github.com/quassel/quassel/commits/master?after=c0b6aaa15a39ddd1d4ad3e2dcbee8e03e41ae59b+0 ) to find all these commits in it :)

PR #174, opened 2016-01-17
Disable SSLv3

esainane:
irssi has done this since late 2014, shortly after POODLE disclosure:
  https://github.com/irssi/irssi/commit/8bd575df2ec0d4a17b6f59116d555b64f5bb1312

See https://bugreports.qt.io/browse/QTBUG-43168

### DIFF
--- a/src/common/main.cpp
+++ b/src/common/main.cpp
@@ -20,6 +20,8 @@
 
 #include <cstdlib>
 
+#include <QSslCipher>
+#include <QSslConfiguration>
 #include <QTextCodec>
 
 #ifdef BUILD_CORE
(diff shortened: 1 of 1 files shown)


esainane:
Brief summary from discussion on IRC at the time:
- QSsl::SecureProtocols isn't actually secure in Qt 4.8. In Qt 5.5, where TlsV1_0OrLater is introduced, SecureProtocols was adjusted to the saner behaviour of disallowing Sslv3 that does not upgrade to TLS anyway. Since SecureProtocols should be updated from Qt release to release, it would be better to stay with SecureProtocols for now.
- Removing ciphers, the "last resort" from https://www.tinfoilsecurity.com/blog/how-to-fix-poodle-and-why-you-are-probably-still-vulnerable can't apply here. While the advice is dated (don't use RC4!), GSM doesn't have enough support yet to be reasonably required either.

As a workaround, I've made an interposer (which should work for anything that dynamically links against openssl): https://github.com/esainane/sanetls

Since there's nothing further we can do here, I'll close this pull request.

PR #175, opened 2016-01-18
Fix the SSL check on systems that don't use OpenSSL

ayonix:
OSX uses another backend than OpenSSL according to someone on #qt, so we shouldn't check for it in cmake. Instead we should only check if `QT_NO_SSL` is defined and do runtime checks with `QSslSocket::supportsSsl()` (which I still have to do in all the right places).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -421,7 +421,7 @@ set(CMAKE_REQUIRED_INCLUDES ${QT_INCLUDES} ${Qt5Core_INCLUDE_DIRS})
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${Qt5Core_EXECUTABLE_COMPILE_FLAGS}")
 check_cxx_source_compiles("
     #include \"qglobal.h\"
-    #if defined QT_NO_OPENSSL || defined QT_NO_SSL
+    #if defined QT_NO_SSL
     #  error \"No SSL support\"
     #endif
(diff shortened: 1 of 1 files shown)


ayonix:
Having the HAVE_SSL and runtime checks asks for duplicate code:
Example in pseudocode:
```
check_ssl:
#ifdef HAVE_SSL
  if (runtime_check) block1
  else block2
#else
  block2
#endif
```

Tucos:
> Having the HAVE_SSL and runtime checks asks for duplicate code:
> Example in pseudocode:
>
> check_ssl:
> #ifdef HAVE_SSL
>   if (runtime_check) block1
>   else block2
> #else
>   block2
> #endif

Actually if `block2` is truly duplicate, you can avoid this in a
manner also used at ircparser.cpp:233. It does decrease legibility and
becomes even messier when `block2` is longer though (as you'd need to
conditionally end the scope). Your example would then become something
like:

   #ifdef HAVE_SSL
    if (runtime_check) block1
    else
   #endif
    block2

I suppose whether this is better or not depends on how often it'd be
used and what the result would look like (and whether anyone actually
wants non-ssl builds).

Sput42:
I agree we should consider making compile-time support of SSL mandatory. Given that we conditional compile for HAVE_SSL in more than a hundred places, adding the runtime checks properly would likely end up in a nightmare.

For now, I think just fixing the compile-time check is fine though.

PR #176, opened 2016-01-27
Add an edit menu on Mac

ayonix:
Adds an empty edit menu on macs, because this adds the `Start dictation...` and `Emoji&Symbols` entries that allow users to use those as input.

 screen shot 2016-01-27 at 14 26 48 screen shot 2016-01-27 at 14 25 49

### DIFF
--- a/src/qtui/mainwin.cpp
+++ b/src/qtui/mainwin.cpp
@@ -536,7 +536,35 @@ void MainWin::setupMenus()
 
 #ifdef Q_OS_MAC
     _editMenu = menuBar()->addMenu(tr("&Edit"));
-    _editMenu->addAction("")->setEnabled(false);
+
+    QAction *undoAct = new QAction(tr("&Undo"), this);
+    undoAct->setShortcuts(QKeySequence::Undo);
+
+    QAction *redoAct = new QAction(tr("&Redo"), this);
+    redoAct->setShortcuts(QKeySequence::Redo);
+
+    QAction *cutAct = new QAction(tr("Cu&t"), this);
+    cutAct->setShortcuts(QKeySequence::Cut);
+    cutAct->setStatusTip(tr("Cut the current selection's contents to the "
+                            "clipboard"));
+
+    QAction *copyAct = new QAction(tr("&Copy"), this);
+    copyAct->setShortcuts(QKeySequence::Copy);
+    copyAct->setStatusTip(tr("Copy the current selection's contents to the "
+                             "clipboard"));
+
+    QAction *pasteAct = new QAction(tr("&Paste"), this);
+    pasteAct->setShortcuts(QKeySequence::Paste);
+    pasteAct->setStatusTip(tr("Paste the clipboard's contents into the current "
+                              "selection"));
+
+    _editMenu->addAction(undoAct);
+    _editMenu->addAction(redoAct);
+    _editMenu->addSeparator();
+    _editMenu->addAction(cutAct);
+    _editMenu->addAction(copyAct);
+    _editMenu->addAction(pasteAct);
+
 #endif
 
     _viewMenu = menuBar()->addMenu(tr("&View"));
--- a/src/qtui/mainwin.cpp
+++ b/src/qtui/mainwin.cpp
@@ -534,6 +534,11 @@ void MainWin::setupMenus()
     _fileMenu->addSeparator();
     _fileMenu->addAction(coll->action("Quit"));
 
+#ifdef Q_OS_MAC
+    _editMenu = menuBar()->addMenu(tr("&Edit"));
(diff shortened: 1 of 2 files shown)


carrotengineer:
:+1: for this - I have to go to a different window to get the "Emoji & Symbols" menu, and copy and paste emoji back in to Quassel, right now. Tiny addition for a big UX win.

@ayonix what's the behaviour if you add the parent menu, but don't add a blank item (line 539)? (got a screenshot?)

ayonix:
If you add only the parent menu it gets weird:
* When you start the client the menu bar has no edit menu in it and you can't click on it.
* When you then focus something else and then the client again the edit menu gets added with the Emoji & Symbols item, but not the Start Dictation...

So yeah, I guess I haven't understood the menu well enough to do it properly :)

AlD:
Thanks for your contribution.
I think the empty entry is a bit too hacky, though and we'll need to find a more proper way before this could be merged.
Additionally, on my El Capitan machine with Qt 4.8.7 I just get an empty menu:
image

Please reopen if you find a fix for these issues.

ayonix:
I've added some basic actions to the menu, so it won't look broken/unintended, but I can't reopen the pull request.
No idea why the entries don't get added with Qt4 though, everything points to the fact that it _should_ be added.

Qt5.6: 
screen shot 2016-05-17 at 17 15 32

Qt4.8.7: 
screen shot 2016-05-17 at 17 16 12

Deph0:
with this implementation, does this fix itself ?
screen shot 2016-06-07 at 6 43 47 pm

AlD:
OK, since this appears to work with Qt >= 5, could you please #ifdef it accordingly?

ayonix:
@Deph0 No, it doesn't touch the chat view.

@AlD I'd still like to confirm if this is only for me or if other people with Qt4 don't have the items either. I couldn't find anything in the documentation why this doesn't work with Qt4 but does with Qt5. If I don't find anything by this evening I will add the `ifdef`.

ayonix:
I noticed the entries actually don't do anything, but get highlighted when you use the keyboard shortcut. Implementing a copy/paste wasn't successful so far, since we have several widgets you can select from without giving them focus, which is the easiest way I've found.

So we could either have entries that do nothing or someone with more knowledge about Qt has to implement the actions correctly.

Regarding Qt4 vs. Qt5: Naming the menu `Edit` instead of `&Edit` adds the desired entries on Qt4 *and* Qt5 for some reason...

CarrotEngineer:
I made these changes on a fork of mine and they seem to work fine. I'm curious what didn't work with respect to copying and pasting, because adding menu items shouldn't actually change anything, and as far as I can tell everything works as it did before, but now I can use the emoji menu and the Edit menu highlights briefly when using any of the actions, as expected.

I also removed the &s from menu items because I'm not convinced they actually did anything. Same with the tooltips - I'm not sure I've ever seen those menu items actually display a tooltip on macOS.

I've attached screenshots to show it looks sane on my machine, and I'm happy for @ayonix to sync up, or make a PR for them separately. Are there other concerns? I suspect it's specific enough with Qt 5 and macOS for the risk to be pretty minimal if it works on my machine now. I'd like to try and push this forwards.

https://github.com/quassel/quassel/compare/master...CarrotEngineer:feature/edit-qt5-macos

screen shot 2016-12-27 at 11 50 01

screen shot 2016-12-27 at 12 00 17

CarrotCodes:
A summary of where this PR stands:

* Qt does not have a way to bring up the "emoji picker" manually (by listening to the standard Ctrl+Cmd+Space key combo)
* Searching Google suggests you can't meaningfully implement Cut/Copy/Paste with widgets because they might do something different / you have to hook them all up manually
* Doing stuff for Qt 4 is basically not worth it at this point if it's getting dropped soon - I'm not even convinced you can get Qt 4 on recent versions of macOS properly
* Implementing "functionless" menu items prevents Cut/Copy/Paste from working *only when using the mouse* - they work and get highlighted as expected when using the keyboard
* The problem this is solving remains really annoying - to use emoji at all in Quassel on macOS you have to write it somewhere else and copy it in

@AlD @ayonix Is there a contextually relevant action we can implement in the Edit menu that actually works? Is doing so even worth it (the blank menu item looks a bit weird, but solves a much more annoying problem on this platform)?

AlD:
Did you test this with Qt 5.6.1? https://bugreports.qt.io/browse/QTBUG-53085 / https://bugreports.qt.io/browse/QTBUG-52931 seem related.

CarrotCodes:
@justjanne suggested moving the View->Search Bar Edit->Find which seems more correct anyway, and would solve the empty item issue.

PR #177, opened 2016-02-02
Fixed a tiny bug in aliasmanager

justjanne:
The `/wait` check is to test if the first command is a `/wait`, which would operate synchronously.
But `/wait` matches also on other commands – so, if, for example, we’d have a command like `/waiting`, it would be matched.
Instead we actually want to match on `/wait_ ` (with space at the end, the `_` is supposed to represent a space).

### DIFF
--- a/src/common/aliasmanager.cpp
+++ b/src/common/aliasmanager.cpp
@@ -198,7 +198,7 @@ void AliasManager::expand(const QString &alias, const BufferInfo &bufferInfo, co
 
     while (!expandedCommands.isEmpty()) {
         QString command;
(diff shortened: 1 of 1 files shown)


sir-ragna:
I don't know the `/wait` command, but does it always take parameters? Because if so the `.trimmed()` function might prevent that last white space to get matched when no parameters are given.

justjanne:
Good argument, @sir-ragna didn’t notice that.

Maybe we should add a special case for that? if ([...]startsWith("/wait ") || […] == "/wait")?

PR #179, opened 2016-02-04
Implement IRCv3 CAP LS negotiation, away-notify

digitalcircuit:
Change IRC connection flow to unconditionally send ```CAP LS``` as part of the connection.  ```IrcEventCap``` now checks the capability list before requesting anything (currently ```sasl```, ```away-notify```) as ```CAP REQ``` only succeeds on an all-or-nothing basis.  If sasl is available and enabled, authentication continues as normal.

Add support for away-notify, including a new event type to parse any away events.  As nicks change status, update the ```ircuser``` objects, including away messages.  Any changes to away can be immediately seen without needing to ```/whois``` for the message.

Modify automatic WHO polling when away-notify is enabled to poll all channels ONCE regardless of polling settings and auto-who nick-limit, as directed by IRCv3 specs.  This is necessary to capture away state for any existing users in the channel.  Any new channels joined after this will queue for a separate one-off WHO poll.

See: http://ircv3.net/specs/core/capability-negotiation-3.1.html
And: http://ircv3.net/specs/extensions/away-notify-3.1.html

### DIFF
--- a/src/common/eventmanager.h
+++ b/src/common/eventmanager.h
@@ -88,6 +88,7 @@ public :
 
         IrcEvent                    = 0x00030000,
         IrcEventAuthenticate,
+        IrcEventAway,
         IrcEventCap,
         IrcEventInvite,
         IrcEventJoin,
--- a/src/core/corenetwork.cpp
+++ b/src/core/corenetwork.cpp
@@ -298,11 +298,7 @@ void CoreNetwork::putCmd(const QString &cmd, const QList<QList<QByteArray>> &par
 
 void CoreNetwork::setChannelJoined(const QString &channel)
 {
-    _autoWhoQueue.prepend(channel.toLower()); // prepend so this new chan is the first to be checked
-    if (_useCapAwayNotify) {
-        // When away-notify is active, the timer's stopped.  Start a new cycle to who this channel.
-        setAutoWhoEnabled(true);
-    }
+    queueAutoWhoOneshot(channel); // check this new channel first
 
     Core::setChannelPersistent(userId(), networkId(), channel, true);
     Core::setPersistentChannelKey(userId(), networkId(), channel, _channelKeys[channel.toLower()]);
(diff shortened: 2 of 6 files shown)


digitalcircuit:
@Nevcairiel wrote on line 494:
> "should ignore" seems like a dangerous word. IIRC there were servers that crapped out on you when you turned on SASL, so doing this entirely unconditionally may cause the same problem.
>
> Also, are we sure there are no servers that support SASL now via CAP REQ, but may not like CAP LS?

@digitalcircuit replied:
> Before, enabling SASL immediately tried to request the capability without checking if the server had it.  Merely requesting capabilities _should_ be less risky.
>
> Unfortunately, beyond adding another setting to the network configuration to turn off all IRCv3 capabilities (then, what if client/core versions mismatch so no setting is shown?), I don't see a way around this risk.  The [client capability negotiation specs](http://ircv3.net/specs/core/capability-negotiation-3.1.html) hint at the same issues.  Likewise for the uncertainty with SASL via CAP REQ versus CAP LS.
>
> It appears that both [HexChat](https://github.com/hexchat/hexchat/blob/4362085847f359ed13df6f8f488a06eb52ecd767/src/common/proto-irc.c#L51 ) and [ZNC](https://github.com/znc/znc/blob/1f226d2adeb2cfa08fa54d360c110c16fd361f18/src/IRCSock.cpp#L1184 ) (when acting as client) take the same approach of unconditionally trying ```CAP LS```.  ZNC [additionally notes that error 451](https://github.com/znc/znc/blob/1f226d2adeb2cfa08fa54d360c110c16fd361f18/src/IRCSock.cpp#L1184 ) should be returned by servers that don't support ```CAP LS```.  Right now [in Quassel](https://github.com/quassel/quassel/blob/17311326016f3654fd33a656b2a0c4ac95776db6/src/core/eventstringifier.cpp#L127 ) that would just print an error message in the status buffer yet still carry on.

Further discussion on IRC suggested always doing ```CAP LS``` after releasing 0.12.3.  If wider testing reveals issues, I can implement a per-network setting to revert to old behavior.

(_And now I know that a ```git rebase``` and ```git push --force``` will eat line comments.  Sorry for the spam!_)

digitalcircuit:
Turns out I should add support for ```account-notify``` and ```extended-join``` at the same time, otherwise user/host information might not be updated if an IRCop uses ```/chghost``` and friends.

I'm working on implementing those right now, and I'll re-open when done.

PR #181, opened 2016-02-07
System proxy support for quassel-client

javierllorente:
Quassel client changes:
- Proxy settings refactored
- Added support for system proxy

### DIFF
--- a/src/client/clientauthhandler.cpp
+++ b/src/client/clientauthhandler.cpp
@@ -75,10 +75,21 @@ void ClientAuthHandler::connectToCore()
     QTcpSocket *socket = new QTcpSocket(this);
 #endif
 
-// TODO: Handle system proxy
 #ifndef QT_NO_NETWORKPROXY
-    if (_account.useProxy()) {
-        QNetworkProxy proxy(_account.proxyType(), _account.proxyHostName(), _account.proxyPort(), _account.proxyUser(), _account.proxyPassword());
+    QNetworkProxy proxy;
+    proxy.setType(_account.proxyType());
+    if (_account.proxyType() == QNetworkProxy::Socks5Proxy ||
+            _account.proxyType() == QNetworkProxy::HttpProxy) {
+        proxy.setHostName(_account.proxyHostName());
+        proxy.setPort(_account.proxyPort());
+        proxy.setUser(_account.proxyUser());
+        proxy.setPassword(_account.proxyPassword());
+    }
+
+    if (_account.proxyType() == QNetworkProxy::DefaultProxy) {
+        QNetworkProxyFactory::setUseSystemConfiguration(true);
+    } else {
+        QNetworkProxyFactory::setUseSystemConfiguration(false);
         socket->setProxy(proxy);
     }
 #endif
--- a/src/qtui/qtuiapplication.cpp
+++ b/src/qtui/qtuiapplication.cpp
@@ -214,7 +214,7 @@ bool QtUiApplication::migrateSettings()
     //
     // NOTE:  If you increase the minor version, you MUST ALSO add new version upgrade logic in
     // applySettingsMigration()!  Otherwise, settings upgrades will fail.
-    const uint VERSION_MINOR_CURRENT = 6;
+    const uint VERSION_MINOR_CURRENT = 7;
     // Stored minor version
     uint versionMinor = s.versionMinor();
 
--- a/src/qtui/qtuiapplication.cpp
+++ b/src/qtui/qtuiapplication.cpp
@@ -279,6 +279,17 @@ bool QtUiApplication::applySettingsMigration(QtUiSettings settings, const uint n
     // saved.  Exceptions will be noted below.
     // NOTE:  If you add new upgrade logic here, you MUST ALSO increase VERSION_MINOR_CURRENT in
     // migrateSettings()!  Otherwise, your upgrade logic won't ever be called.
+    case 7:
+    {
+        // New default changes: ProxyType=3 (no proxy) now means QNetworkProxy::HttpProxy
+        // So we have to change it to ProxyType=2 (QNetworkProxy::NoProxy)
+        const QString proxyType = "ProxyType";
+        if (settings.valueExists(proxyType) && settings.value(proxyType)=="3") {
+            settings.setValue(proxyType, 2);
+        }
+        // Migration complete!
+        return true;
+    }
     case 6:
     {
         // New default changes: sender colors switched around to Tango-ish theme
--- a/src/qtui/settingspages/coreaccountsettingspage.cpp
+++ b/src/qtui/settingspages/coreaccountsettingspage.cpp
@@ -250,12 +250,40 @@ CoreAccountEditDlg::CoreAccountEditDlg(const CoreAccount &acct, QWidget *parent)
     ui.user->setText(acct.user());
     ui.password->setText(acct.password());
     ui.rememberPassword->setChecked(acct.storePassword());
-    ui.useProxy->setChecked(acct.useProxy());
-    ui.proxyHostName->setText(acct.proxyHostName());
-    ui.proxyPort->setValue(acct.proxyPort());
-    ui.proxyType->setCurrentIndex(acct.proxyType() == QNetworkProxy::Socks5Proxy ? 0 : 1);
-    ui.proxyUser->setText(acct.proxyUser());
-    ui.proxyPassword->setText(acct.proxyPassword());
+
+    ui.buttonGroupProxyType->setId(ui.radioButtonNoProxy, 0);
+    ui.buttonGroupProxyType->setId(ui.radioButtonSystemProxy, 1);
+    ui.buttonGroupProxyType->setId(ui.radioButtonManualProxy, 2);
+
+    bool manualProxy = false;
+    switch (acct.proxyType()) {
+    case QNetworkProxy::NoProxy:
+        ui.buttonGroupProxyType->button(0)->setChecked(true);
+        break;
+    case QNetworkProxy::DefaultProxy:
+        ui.buttonGroupProxyType->button(1)->setChecked(true);
+        break;
+    case QNetworkProxy::Socks5Proxy:
+        ui.buttonGroupProxyType->button(2)->setChecked(true);
+        ui.proxyType->setCurrentIndex(0);
+        manualProxy = true;
+        break;
+    case QNetworkProxy::HttpProxy:
+        ui.buttonGroupProxyType->button(2)->setChecked(true);
+        ui.proxyType->setCurrentIndex(1);
+        manualProxy = true;
+        break;
+    default:
+        break;
+    }
+
+    if (manualProxy) {
+        ui.proxyHostName->setText(acct.proxyHostName());
+        ui.proxyPort->setValue(acct.proxyPort());
+        ui.proxyType->setEnabled(true);
+        ui.proxyUser->setText(acct.proxyUser());
+        ui.proxyPassword->setText(acct.proxyPassword());
+    }
 
     if (acct.accountId().isValid())
         setWindowTitle(tr("Edit Core Account"));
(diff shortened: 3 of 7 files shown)


AlD:
Thanks for your contribution!

The patch looks good in general. To prevent misunderstandings and make the review a bit easier, could you please add some more details:
- What use cases are now supported that didn't work before?
- Which platforms does this work on?
- Which platforms has this been tested on?
- How can this be tested?

javierllorente:
Hello!
Here are my answers:

- Now Quassel can use whatever proxy is used system-wide, and therefore the user wouldn't need to configure the proxy settings for Quassel alone.
- It works on Linux and Windows (I have tested this code on these two platforms)

According to the [Qt5 docs](http://doc.qt.io/qt-5/qnetworkproxyfactory.html), there are some limitations

> 
> These are the limitations for the current version of this function. Future versions of Qt may lift some of the limitations listed here.
> 
>  - On MacOS X, this function will ignore the Proxy Auto Configuration settings, since it cannot execute the associated ECMAScript code.
>  - On Windows platforms, this function may take several seconds to execute depending on the configuration of the user's system.

- The easiest way to test this is to check "system proxy" on Quassel's proxy settings and run Quassel on an environment that requires the use of a proxy. Note: the system-wide proxy must be configured in the appropiate place (ie: KDE systemsettings, Windows Internet settings). Quassel would get the proxy settings from there.

winks:
@javierllorente sorry to put the workload now on you, but could you quickly check if the mouseover popup window from links will also use this proxy setting (easy to check if you just paste a link to a webserver that you control where you can check with which source ip address the request fires). If yes, then #18 can stay closed :)

javierllorente:
@winks I have tested the mouseover window popups on Linux and they work fine - they use the system proxy :)

This is what I did: 
1. I set/export the http_proxy and https_proxy. 
2. I ran quasselclient from the command line
3. I set the Quassel proxy settings to system proxy.
4. I placed my mouse over an URL to see the web page preview popup

There is a problem regarding upgrades of Quassel on Linux (and probably on other platforms). If you don't use a proxy, Quassel's proxy settings are set to "manual proxy" after the upgrade. On new installs, "system proxy" is the default setting.

winks:
Thanks a lot! (as you can see in the other PR that I wanted such a feature back then :P)

javierllorente:
You're welcome! :)

AlD:
> There is a problem regarding upgrades of Quassel on Linux (and probably on other platforms). If you don't use a proxy, Quassel's proxy settings are set to "manual proxy" after the upgrade.

That doesn't sound good. Why exactly is that? Isn't it possible to detect if the settings were set manually in the old version and override with the new defaults (once) otherwise?

javierllorente:
It's because 
- not using a proxy on current version → ProxyType=3 is written to quasselclient.conf
- with this patch, ProxyType=3 means that the client is using a proxy (QNetworkProxy::HttpProxy has a value of 3)

Sput42:
With #185 being merged, we now have a way to migrate settings to newer versions... that could/should be used maybe?

AlD:
Hey Javier, any chance you could add a fix for the settings issue using https://github.com/quassel/quassel/pull/185/commits/150eac5b52438ec207e5035edf92b6f0f7d1f09f?

javierllorente:
Hello Daniel,
I have created a new case for migrating the proxy settings. Fix is untested.

src/qtui/qtuiapplication.cpp
```diff
@@ -270,6 +270,18 @@ bool QtUiApplication::applySettingsMigration(QtUiSettings settings, const uint n
     // Each missed version will be called in sequence.  E.g. to upgrade from '1' to '3', this
     // function will be called with '2', then '3'.
     // Use explicit scope via { ... } to avoid cross-initialization
+
+    case 4:
```
digitalcircuit: Almost got it!  You also need to change [```VERSION_MINOR_CURRENT```](https://github.com/quassel/quassel/pull/181/files#diff-880688f2885d4ebfc2e07927657df789R213 ) from ```3``` to ```4``` in ```QtUiApplication::migrateSettings()``` so your migration code gets run.

```cpp
// Check minor settings version, handling upgrades/downgrades as needed
// Current minor version
const uint VERSION_MINOR_CURRENT = 3;  // ← THIS VALUE - change from 3 to 4
// Stored minor version
```

I know this isn't ideal, and if anyone has ideas on improving the migration logic, I'd be happy to try doing the work.  At minimum I'll add some comments to here in my next pull-request :)

src/qtui/qtuiapplication.cpp
```diff
@@ -270,6 +270,18 @@ bool QtUiApplication::applySettingsMigration(QtUiSettings settings, const uint n
...
+    {
+       // New default changes: ProxyType=3 (no proxy) now means QNetworkProxy::HttpProxy
+      // So we have to change it to ProxyType=2 (QNetworkProxy::NoProxy)
+        const QString proxyType = "ProxyType";
+        if (settings.valueExists(proxyType) && settings.value(proxyType)=="3") {
```
digitalcircuit: Really minor thing, you might want to re-align the spaces for the comment and ```settings.value``` check.

src/qtui/qtuiapplication.cpp
javierllorente: Done. Perhaps a comment pointing to that const in applySettingsMigration() helps.

src/qtui/qtuiapplication.cpp
javierllorente: Done for the comments. I've seen that the `settings.value` check is fine regarding spaces.

src/qtui/qtuiapplication.cpp
digitalcircuit: Good idea - I'll open a new small pull request for that soon.

digitalcircuit: And done - see [pull request #269](https://github.com/quassel/quassel/pull/269 ).  I apologize for the confusion over this.

javierllorente: Excellent. Thanks for your comments!

digitalcircuit:
Alas, you'll need to rebase your pull request since another one that modifies settings was merged ([pull request #266](https://github.com/quassel/quassel/pull/266)).  You should just need to increase `VERSION_MINOR_CURRENT` again, and make sure to use this new number for the settings migration.

```diff
-    const uint VERSION_MINOR_CURRENT = 4;
+    const uint VERSION_MINOR_CURRENT = 7;
```

```diff
-    case 4:
+    case 7:
    {
        // New default changes: ProxyType=3 (no proxy) now means QNetworkProxy::HttpProxy
        // So we have to change it to ProxyType=2 (QNetworkProxy::NoProxy)
        const QString proxyType = "ProxyType";
        if (settings.valueExists(proxyType) && settings.value(proxyType)=="3") {
            settings.setValue(proxyType, 2);
        }
        // Migration complete!
        return true;
}
```

Any Quassel client configuration files you previously upgraded to this pull request's version 4 might be inconsistent.  You can just delete it and start fresh if desired, or manually set `VersionMinor` to 3 in the file to force the new migrations to apply.

If you need further help, let me know!

digitalcircuit:
Ack, I apologize, I missed another comment change on the `master` branch.  If you update your local Git tree and rebase this on `master`, you should only see one minor change.

[Here](https://github.com/quassel/quassel/pull/181/commits/a54935583462fc959650a439cb4b6f524f173223#diff-880688f2885d4ebfc2e07927657df789R273), Git views your pull request commit as adding these...
```diff
-
-    case 4:
+    //
+    // In most cases, the goal is to preserve the older default values for keys that haven't been
+    // saved.  Exceptions will be noted below.
+    case 7:
```

When the `master` branch [now has a different comment](https://github.com/quassel/quassel/blob/master/src/qtui/qtuiapplication.cpp#L277-L280 ).
```cpp
    // In most cases, the goal is to preserve the older default values for keys that haven't been
    // saved.  Exceptions will be noted below.
    // NOTE:  If you add new upgrade logic here, you MUST ALSO increase VERSION_MINOR_CURRENT in
    // migrateSettings()!  Otherwise, your upgrade logic won't ever be called.
    case 6:
{
```

You should be able to just `git rebase master` on this branch, and if that has merge conflicts, change it to
```diff
-    case 4:
+    case 7:
```
I.e. dropping the newline and comment change.  If this doesn't work, let me know.  I haven't tested it locally since I've got tasks to due pretty soon, but I should be around later.  You can also reach me as `digitalcircuit` in `Freenode/#quassel`!

javierllorente:
Yes, I did it wrong 0:-). It's now merged.

digitalcircuit:
Looks good here!  I'll try to give this an actual test in the near future, but nothing stands out as concerning from reading through your changes :)

javierllorente:
Excellent! Thanks! :)

src/qtui/settingspages/coreaccountsettingspage.cpp
```diff
@@ -266,18 +294,37 @@ CoreAccountEditDlg::CoreAccountEditDlg(const CoreAccount &acct, QWidget *parent)
...
+        break;
+    case SystemProxy: // QNetworkProxy::DefaultProxy:
+        QNetworkProxyFactory::setUseSystemConfiguration(true);
+        _account.setProxyType(QNetworkProxy::DefaultProxy);
+        break;
+    case ManualProxy: // QNetworkProxy::Socks5Proxy || QNetworkProxy::HttpProxy
+        proxyType = ui.proxyType->currentIndex() == 0 ?
+                    QNetworkProxy::Socks5Proxy : QNetworkProxy::HttpProxy;
+        QNetworkProxyFactory::setUseSystemConfiguration(false);
+        _account.setAccountName(ui.accountName->text().trimmed());
```
Sput42: I don't think this is correct - you need to set the non-proxy-related things regardless of the proxy settings (i.e., move out of the switch)

src/qtui/settingspages/coreaccountsettingspage.h
```diff
@@ -102,12 +103,14 @@ private slots:
...
     void on_accountName_textChanged(const QString &);
     void on_user_textChanged(const QString &);
+    void on_radioButtonManualProxy_toggled(bool checked);
 
     void setWidgetStates();
 
 private:
     Ui::CoreAccountEditDlg ui;
     CoreAccount _account;
+    enum ProxyType { NoProxy, SystemProxy, ManualProxy };
```
Sput42: I don't think you're using this enum, are you?

src/qtui/settingspages/coreaccountsettingspage.h
```diff
@@ -22,6 +22,7 @@
 #define COREACCOUNTSETTINGSPAGE_H_
 
 #include <QSortFilterProxyModel>
+#include <QNetworkProxy>
```
Sput42: I'd prefer alphabetic ordering in new code

digitalcircuit:
@javierllorente Adding on to Sput's review, if you don't have time in the next week or so, I'd be willing to adopt your changes while still giving you credit (*via setting `GIT_AUTHOR`*), similar to what I did with the "no sender brackets" pull request :)

src/qtui/settingspages/coreaccountsettingspage.h
javierllorente: I am using it in CoreAccountEditDlg::account()

javierllorente:
Sorry, I was on holidays. Done with the requested changes :)

src/qtui/settingspages/coreaccountsettingspage.cpp
javierllorente: Oops. Corrected.

src/qtui/settingspages/coreaccountsettingspage.h
javierllorente: Done :)

src/client/clientauthhandler.cpp
```diff
@@ -75,10 +75,21 @@ void ClientAuthHandler::connectToCore()
...
+        proxy.setPort(_account.proxyPort());
+        proxy.setUser(_account.proxyUser());
+        proxy.setPassword(_account.proxyPassword());
+    }
+
+    if (_account.proxyType() == QNetworkProxy::DefaultProxy) {
+        QNetworkProxyFactory::setUseSystemConfiguration(true);
+    } else {
+        QNetworkProxyFactory::setUseSystemConfiguration(false);
         socket->setProxy(proxy);
```
digitalcircuit: Just to confirm, you're not needing to call `socket->setProxy(proxy)` when using the system configuration?  If so, I'm confused as to why you'd create the proxy at all when using system proxy, then ignore it down here.

I've not actually tested this; pardon if I misunderstand.

src/qtui/qtuiapplication.cpp
digitalcircuit: The `master` branch is now up to [settings minor version 6](https://github.com/quassel/quassel/blob/0260fbe00ddbd1ce78b1870f40fce969b8f34ee5/src/qtui/qtuiapplication.cpp#L215-L217 ), so you'd need to rebase this to [add a version 7 at the top here](https://github.com/quassel/quassel/blob/0260fbe00ddbd1ce78b1870f40fce969b8f34ee5/src/qtui/qtuiapplication.cpp#L282-L332 ).

It looks like you did this in the past [based on an older comment]( https://github.com/quassel/quassel/pull/181#issuecomment-294010867 ); maybe a commit got dropped?

src/client/clientauthhandler.cpp
javierllorente: It's not needed. According to the [Qt5 docs](http://doc.qt.io/qt-5/qnetworkproxyfactory.html#setUseSystemConfiguration), setUseSystemConfiguration() 
>Enables the use of the platform-specific proxy settings, and only those. 
>[...]
>Calling this function with enable set to true resets any proxy or QNetworkProxyFactory that is already set.

javierllorente:
@digitalcircuit The migration logic is now up-to-date :)

src/client/clientauthhandler.cpp
digitalcircuit: Noted!  Thank you for the quick reply (and pardon my long delay).

In this case, would it work to reorganize the `if` to avoid creating a new proxy object when it's not needed?  E.g.
```cpp
#ifndef QT_NO_NETWORKPROXY
    if (_account.proxyType() == QNetworkProxy::DefaultProxy) {
        QNetworkProxyFactory::setUseSystemConfiguration(true);
    } else {
        // If not using the default proxy, still create the NetworkProxy object in order to
        // override the default use of "QNetworkProxy::DefaultProxy".
        QNetworkProxyFactory::setUseSystemConfiguration(false);
        QNetworkProxy proxy;
        proxy.setType(_account.proxyType());
        if (_account.proxyType() == QNetworkProxy::Socks5Proxy ||
                _account.proxyType() == QNetworkProxy::HttpProxy) {
            proxy.setHostName(_account.proxyHostName());
            proxy.setPort(_account.proxyPort());
            proxy.setUser(_account.proxyUser());
            proxy.setPassword(_account.proxyPassword());
        }
        socket->setProxy(proxy);
    }
#endif
```

**Note, I am not sure if this is any "better" or more efficient**, it is just something that came to mind.  Feel free to ignore it :)

javierllorente:
@digitalcircuit I think that it should work but I don't know if it's more efficient. ATM I don't have a proxy to play with.

Sput42:
I had to redo the settings migration logic - not sure what the previous implementation was supposed to do, but the UseProxy flag was used to indicate if a proxy was to be used, nothing to do with a ProxyType == 3. Also, that information is stored in [CoreAccounts], so migration needs to account for that.

javierllorente:
@Sput42 It was a translation of a settings value (a proxy being used in a previous version).

PR #183, opened 2016-02-14
Enable Link-Preview on appveyor builds

romibi:
Pros:
- Those who want it don't have to build quassel themselves anymore

Cons (what I heard):
- Adds about 23MB to the Download
- Whole Link Preview should be redone anyways

Comments:
- Those who don't want it or have problems with it can disable it in the Chat View Settings
- I don't think the Cons are a problem for this


TheOneRing:
You need to ship icu too with webkit.

romibi:
they were already in it … but some others were missing

TheOneRing:
Seems to work :+1: :smiley:

AlD:
Thanks for the patch, romibi, and thanks for testing, TheOneRing!

PR #184, opened 2016-02-15
Warn when IRC server SSL validation fails

digitalcircuit:
During connection, warn if the SSL validation fails, then continue connecting as before.  This isn't ideal, but it's better than silently accepting the error without any indication.

In the future, Quassel should stop connecting entirely and require toggling an override in the IRC server network settings.  That will require a database schema change, a new checkbox in the UI, CoreFeature, etc.

Example output:
```
[6:04:32 pm] * Connecting to irc.rizon.net:6697...
[6:04:32 pm] * WARNING: Encrypted connection is insecure and might not be trustworthy
[6:04:32 pm] * Reason: The certificate is self-signed, and untrusted
```

### DIFF
--- a/src/core/corenetwork.cpp
+++ b/src/core/corenetwork.cpp
@@ -1040,9 +1040,15 @@ void CoreNetwork::sendAutoWho()
 #ifdef HAVE_SSL
 void CoreNetwork::sslErrors(const QList<QSslError> &sslErrors)
 {
-    Q_UNUSED(sslErrors)
+    displayMsg(Message::Error, BufferInfo::StatusBuffer, "",
+               tr("WARNING: Encrypted connection is insecure and might not be trustworthy"));
+    if (!sslErrors.empty()) {
+        displayMsg(Message::Error, BufferInfo::StatusBuffer, "",
+                   tr("Reason: %1").arg(sslErrors.first().errorString()));
+    }
     socket.ignoreSslErrors();
-    // TODO errorhandling
+    // TODO Implement an off-by-default "Allow insecure connections" setting for networks.
+    // Don't ignore SSL errors unless that setting is enabled.
 }
 
 


kalaksi:
As discussed on IRC, I'd prefer if the message said "the encrypted connection is insecure" so it's clear that the connection is still encrypted. And I don't think there would be a need to fall back to unencrypted connection automatically.

And for the record about the override mechanism which was also discussed: it'd be probably fine if a checkbox would be added next to the "Use SSL" checkbox saying "Allow insecure SSL connections". Greyed out when "Use SSL" is not selected. "Allow insecure..." instead of something like "Ignore errors" since it better conveys the meaning. (Should we already use "TLS" or language that wouldn't use the protocol names? "Use encrypted connection" and "Allow insecure encrypted connection"?)

The DB schema change could just simply be the addition of "allowinsecure"-column to ircserver table. This could be backwards compatible if the column was made "true" by default. The quassel client  could have it "false" by default, so it would be "false" by default when supported by the client and "true" when not (the current behaviour). Not sure if it's a good idea, though, considering that quassel could have core clients that don't support the option and would therefore allow insecure connection by default, but the situation wouldn't be any worse from the current.

Some of this was probably rather obvious but you requested feedback so there you go :)

digitalcircuit:
@kalaksi I think I need to check for [```QSslError::NoSslSupport```](https://doc.qt.io/qt-4.8/qsslerror.html ) before claiming the connection's still encrypted.  I agree that this should be fixed for the long-term override-forced behavior.  Perhaps some combination of your message and "may be insecure."

I agree with the override mechanism in the network configuration screen, including changing the strings (perhaps as a different translation pull-request).

As to the DB schema, I think that's only part of the story.  E.g. when SASL authentication was added, the database schema changed, migration scripts changed, and the core added a new feature [```SaslAuthentication```](https://github.com/quassel/quassel/blob/17311326016f3654fd33a656b2a0c4ac95776db6/src/common/quassel.h#L71 ) so the client could [show and hide the UI as needed - full commit here](https://github.com/quassel/quassel/commit/e4972a0ebce0d555d752ab2f34ffd6a6dcdef1dd ).  I aim to do this, but I'm not certain of the best approach yet and wish to study more of Quassel first.  Naturally, anyone more familiar with Quassel's codebase can take over this effort if they'd like.

Erroring out enabled by default should be reasonable trade-off between security and usability.  Usually the client is easier to upgrade than the core, and the error message would make it easy to point folks to upgrading their client to revert to the old behavior.

Thank you for your feedback!

digitalcircuit:
I've changed the wording of the message as suggested by @kalaksi.  Still need to do the "Allow insecure connections" toggle in here or a separate pull request, time permitting.

digitalcircuit:
Superseded by [pull request #217](https://github.com/quassel/quassel/pull/217 ), a much more thorough, proper implementation.

PR #80, opened 2014-07-01
Nobrackets

AlD:
Removes brackets around sender names.

PR #185, opened 2016-02-16
Optionally remove <nick> brackets, sender/nick colors by default

digitalcircuit:
## In short
* Allow turning on and off the <nick> brackets with a setting in ```Interface``` → ```Chat View```
 * This does not affect notifications; see comments below
 * Copying text will include the brackets like before, even when they're hidden
* Allow coloring the nickname in ```Action``` messages
 * Improves readability for long action messages, such as on story-telling channels
 * Fixes stylesheets not being able to reference sender hashes
* Implement settings upgrade logic
 * New setting ```VersionMinor``` for compatible changes, such as changing defaults
 * For existing installs, changed settings and old defaults kept as in *Classic*
 * For fresh installs, sender/timestamp brackets removed, sender colors enabled as in *Modern*
* Move ```Chat View``` colors to new settings page ```Chat View Colors```
 * Reduces visual clutter, improves organization, and saves vertical space, good for small screens
 * Modify layout to handle wider Settings pages, e.g. when maximized (*also see [pull request #232](https://github.com/quassel/quassel/pull/232)*)

*Previously, old defaults weren't kept in a few known cases.*  ***This should be fixed*** *as of ```2016-6-27```.*

Criteria | Rank | Reason
Impact | ★★☆ *2/3* | Major change to default Quassel look, often more readable
Risk | ★★☆ *2/3* | Protest of new defaults, minor theme conflicts, upgrade edge cases
Intrusiveness | ★☆☆ *1/3* | Style-sheet handling, translations, shouldn't interfere with other PRs

This expands upon @AlD's pull request #80 (thank you!), making the choice of brackets or no brackets configurable, among other things.

## Rationale
Modern chat clients bias towards a more colorful, minimal approach, such as [Quasseldroid's no-bracket look](https://i.imgur.com/GSvIT65.png ), or [IrcCloud's](https://www.irccloud.com/) [mobile interface](https://static.irccloud-cdn.com/static/landing/imgs/ipad-white.png ).

As Quassel can inject the brackets when copying text, there's no need to always display them.  And for those who prefer the older style, it can always be turned back on.

*Note: Changes to defaults can be reverted if required to merge.  The* ***classic style still exists***, *it just is no longer default.*

### Modern (default for new installs)
*Sender brackets hidden, sender colors enabled and nicks colored in action messages*
Show brackets hidden

### Classic (default for upgrades)
*Sender brackets shown, sender colors disabled and nicks not colored in action messages*
Show brackets shown

### Rearranged Settings interface
*Quassel 0.12.4 on top, Quassel 0.13 + this pull request on bottom - less clutter despite more options*
Changes to Chat View settings pane for configuring colors

### Quasseldroid Alpha (for comparison)
*Sender brackets hidden, sender colors enabled and nicks colored in action messages.  Screenshot trimmed to take up less space.*
Quasseldroid showing bracket-less, colored nicks look

### Settings maximized - Chat View Colors
*Settings showing the Chat View Colors pane, maximized on a 1920×1080 monitor*
Before and after Settings maximized -> Chat View Colors
**Other settings panes are handled in [pull request #232](https://github.com/quassel/quassel/pull/232 ).**


TheOneRing:
Keep in mind that there are more places with the brackets https://github.com/quassel/quassel/blob/master/src/qtui/knotificationbackend.cpp#L67

digitalcircuit:
Good catch!  I'll think on this some more for I don't know how all notification frameworks will interact.  For example, on Ubuntu, KNotify defaults to using ```libnotify-osd``` which doesn't have support for formatting in notifications.  Removing the brackets might make it harder to read.

**Notification *with* brackets around sender**
Ubuntu, KNotify, and libnotify-osd bubble without formatting

I certainly agree with your concern of keeping consistency.  However, if this pull request is sufficiently useful now, perhaps the remaining 80% of effort (investigating notification frameworks) for 20% of functionality should be addressed with a follow-up pull request.

----
*As an aside, I think the KNotify layout could be improved.  A long while before getting into Quassel development I wrote a program taking advantage of KNotify's "log to file" option.  It puts the sender in the title and removes it from the message except for ```/me``` notifications.  I'm guessing KNotify might allow a similar approach, including support for the ```x-canonical-append``` flag.*

Ubuntu, KNotify, and libnotify-osd bubble using custom NotifyMerged program

digitalcircuit:
In ```#quassel```, @romibi suggested using a ```switch``` fall-through for the settings upgrade logic.  It may look more clean, but it might also complicate printing error messages and tracking successful upgrades.  We weren't able to determine which coding style would be best, considering [some languages don't support it](https://blogs.msdn.microsoft.com/abhinaba/2005/11/15/c-i-miss-case-fall-through/ ), and [others argue for and against this style choice](https://programmers.stackexchange.com/questions/116221/appropriate-uses-of-fall-through-switch-statements ).

I'd be happy to either leave it as is (*for loop calling a dedicated function*), or try to change it (*switch fall-through*).

TheOneRing:
I'd prefer more intense default colours on a white background. The colors may look great on a dark background but I think they might be hard to read.

TheOneRing:
https://github.com/quassel/quassel/blob/master/src/qtui/snorenotificationbackend.cpp#L78 there are still some <> :)

TheOneRing:
But I really like it!
It makes it more "human" and feels better to read.

digitalcircuit:
### Colors
I'd be happy to change the colors; I agree that they could use some more care.  However I have no idea what to pick - perhaps a poll in ```#quassel```, or something?  I know ```justJanne``` was looking into color themes for [QuasselDroid](https://github.com/sandsmark/QuasselDroid ).

If need be they can always be changed in a follow-up pull request once this is merged.  Likewise with the color for self, for I don't what to choose for it, either.

### Notifications
Good catch with Snore too.  Though as mentioned above, I'd need to investigate each notification system more thoroughly since it looks like some backends don't use formatting (*e.g. [Windows 10 toasts](https://techbase.kde.org/images.techbase/6/64/Windows-toast-win10.png ) - maybe using the title?*) and Quassel needs a way to distinguish sender from message.  I'm not sure if changing sender brackets in the Quassel UI should also change the (*usually*) system-themed notifications.

Thank you for reviewing this and for the compliment!

PR #187, opened 2016-02-19
Basic Touch Support

romibi:
Improved scrolling with a Touch Screen: (Before/After)
https://gyazo.com/33fefa848189711ac68db43c471a35c3
(It was possible before with the Scrollbar but difficult)

Works in BufferView, ChatView and NickView:
https://gyazo.com/ae8a81ce26cb71ab123e174534fce856

Old Behaviour (Selecting/Dragging) if movement starts horizontal:
https://gyazo.com/0a6fec318587581b261c8a403292bc32

Mouse Behaviour shouldn't have Changed…

I hope the new TreeViewTouch class is ok …


TheOneRing:
Code looks reasonable and touch works fine.
I can finally stop using the slider with touch which was quite nasty.

CounterPillow:
I just want to comment on how nice this PR is presentation-wise, with the animated images demonstrating it. Even as somebody who doesn't use a touchscreen, I understand exactly what the interface changes are that this patchset makes through this. Definitely going to remember this if I ever submit a PR myself.

src/qtui/chatview.cpp
```diff
@@ -108,7 +109,20 @@ bool ChatView::event(QEvent *event)
         }
     }
 
-    if (event->type() == QEvent::Wheel) {
+    if (event->type() == QEvent::TouchBegin && _lastTouchStart < QDateTime::currentMSecsSinceEpoch() - 1000) { //(slow) double tab = normal behaviour = select text. 1000 ok?
```
AlD: I think 1s a bit too much.
Does Qt provide a setting we could delegate this to?

src/uisupport/treeviewtouch.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+TreeViewTouch::TreeViewTouch(QWidget *parent)
+	: QTreeView(parent)
+{
+	setAttribute(Qt::WA_AcceptTouchEvents);
+}
+
+
+bool TreeViewTouch::event(QEvent *event) {
+	if (event->type() == QEvent::TouchBegin && _lastTouchStart < QDateTime::currentMSecsSinceEpoch() - 1000) { //(slow) double tab = normal behaviour = select multiple. 1000 ok?
```
AlD: I think 1s a bit too much.
Does Qt provide a setting we could delegate this to?

AlD:
Thanks for this very nicely presented PR.

Code looks nice as well, one minor comment above.

romibi:
Thanks :)
About the 1s … That line is already gone.
With the last commit i changed from slow double tab to directional behaviour:
Before last commit:
- Touch and Drag -> Scroll
- (slow) Double Touch and Drag -> Select

After last commit:
- Touch and Drag Vertically -> Scroll
- Touch and Drag Horizontally -> Select

The Direction only gets checked in the beginning … so if you start Horizontally but you want to select more above or below that is no problem.

(Fast Double Touch = Double Click -> Select 1 Word is unchanged Behaviour in both…)

TheOneRing:
Works fine with qt5 on Windows

TheOneRing:
Ping

Sput42:
Very nice PR! I would appreciate if you could squash the last three fixup commits into their respective "parents" (using interactive rebase, just order the fixes behind the original commits they're fixing and squash them in). That way we won't needlessly end up with non-building commits in the history.

PR #186, opened 2016-02-18
Update all channel/nick/query tooltips to new-style

digitalcircuit:
## In short
* Modify channel, network, query and nick tooltips to use a clean, consistent style
 * More padding trades some efficiency for significant readability improvements on tool-tips with lots of information - *see examples below*
 * Simplifies expanding tooltips in the future, such as after [pull request #198](https://github.com/quassel/quassel/pull/198 ) is merged
* Unifies HTML escaping
 * Reduces clutter of ```#if QT_VERSION```
 * Fixes crashes and weird formatting by escaping HTML where reasonable, such as nick ident information
 * Fixes titles line-wrapping by replacing spaces with non-breaking spaces
* Shows multiple prefixes in channel modes if available
 * Gives ```multi-prefix``` a user-visible benefit
 * Finishes up work in [pull request #188](https://github.com/quassel/quassel/pull/188 )

Criteria | Rank | Reason
---------|---------|-------------
Impact | ★★☆ *2/3* | User-facing look and consistency, eases adding more information
Risk | ★☆☆ *1/3* | Possible theming or stylistic concerns (*risk accepted in #145*)
Intrusiveness | ★★☆ *2/3* | Translation change, needed for other improvements including IRCv3

Expands upon @Scheirle's pull request https://github.com/quassel/quassel/pull/145.  Thanks!

# Examples (old on left, new on right)
*Note: Ubuntu's Ambience theme doesn't mesh well with Qt's tooltips; it'll look nicer with other themes*
### Topic shown
Channel - name, users, mode, topic

### Topic hidden
Channel - name, users, mode

## Networks
Network - name, server, users, lag

### Real name, hostmask, etc
Nick - realname, hostmask, idle, login, server
### Away message, real name, hostmask etc
Nick - away, away message, realname, hostmask, idle, login, server

### Away (no known message), real name, hostmask etc
*Note: on modern networks [IRCv3 support for ```away-notify```](https://github.com/quassel/quassel/pull/180 ) should make this a rare occurrence*
Nick - away, realname, hostmask, server

### No information
Nick - no information

## Queries (PMs)
*Note: this is mostly identical to Nicks above*
### Real name, etc
Query - realname, hostmask, operator, server

### No information (e.g. offline)
Query - no information

PR #188, opened 2016-02-29
Don't update IrcUser modes with channel-specific modes

digitalcircuit:
Channel-specific user modes in ```IrcChannel``` should not be applied to ```IrcUser``` objects shared between channels.  Otherwise, this results in getting ```+o``` operator status in one channel implying ```+o``` in other channels.  Fortunately, only tooltips seem affected.

User modes shown in the nick list of Quassel should instead check the current channel user modes in ```IrcUserItem::toolTip()```, perhaps as part of pull-request #186.

### DIFF
--- a/src/core/coresessioneventprocessor.cpp
+++ b/src/core/coresessioneventprocessor.cpp
@@ -942,8 +942,22 @@ void CoreSessionEventProcessor::processIrcEvent352(IrcEvent *e)
             // Some IRC servers decide to not follow the spec, returning only -some- of the user
             // modes in WHO despite listing them all in NAMES.  For now, assume it can only add
             // and not take away.  *sigh*
-            if (!validModes.isEmpty())
-                ircuser->addUserModes(validModes);
+            if (!validModes.isEmpty()) {
+                if (channel != "*") {
+                    // Channel-specific modes received, apply to given channel only
+                    IrcChannel *ircChan = e->network()->ircChannel(channel);
+                    if (ircChan) {
+                        // Do one mode at a time
+                        // TODO Better way of syncing this without breaking protocol?
+                        for (int i = 0; i < validModes.count(); ++i) {
+                            ircChan->addUserMode(ircuser, validModes.at(i));
+                        }
+                    }
+                } else {
+                    // Modes apply to the user everywhere
+                    ircuser->addUserModes(validModes);
+                }
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/core/coresessioneventprocessor.cpp
```diff
@@ -942,8 +942,22 @@ void CoreSessionEventProcessor::processIrcEvent352(IrcEvent *e)
...
             // and not take away.  *sigh*
-            if (!validModes.isEmpty())
-                ircuser->addUserModes(validModes);
+            if (!validModes.isEmpty()) {
+                if (channel != "*") {
+                    // Channel-specific modes received, apply to given channel only
+                    IrcChannel *ircChan = e->network()->ircChannel(channel);
+                    if (ircChan) {
+                        // Do one mode at a time
+                        // TODO Better way of syncing this without breaking protocol?
```
AlD: What is this TODO about? This change doesn't seem to break protocol.

digitalcircuit: This change doesn't break protocol, but it's deliberately inefficient for the sake of maintaining backwards compatibility.  [```IrcChannel::addUserMode```](https://github.com/quassel/quassel/blob/0cdc3bdc397e5d9b45fecb682f0de73d47f9e228/src/common/ircchannel.cpp#L289 ) calls ```IrcChannel::isValidChannelMode```, which [expects only one channel mode at a time](https://github.com/quassel/quassel/blob/0cdc3bdc397e5d9b45fecb682f0de73d47f9e228/src/common/ircchannel.cpp#L75 ), and otherwise rejects it entirely.  Rather than extend the protocol and break functionality with older clients, I opted for this alongside a ```TODO``` so it could be reconsidered in the future.

As a future example, @esainane has a series of commits on [their Quassel fork](https://github.com/esainane/quassel/commits/master ) which makes it possible to change the protocol in a backwards-compatible way.  Among other more substantial things, it'd allow optimizing this to a single ```IrcChannel::addUserModes``` call should their ```ClientFeatures``` system ever be proposed and merged.

I [followed the same pattern](https://github.com/quassel/quassel/pull/180/commits/4aea5769f4c155234394957aa55360f111c12cbb#diff-96a0a2f0c28dbfd6b88fc7d083fc5fbaL180 ) in the ```multi-prefix``` portion of my original IRCv3 pull request.

Only risk seems to be older desktop clients logging warnings in ```Help → Debug → Debug Log```.  See the [tests described in one of the original IRCv3 commits](https://github.com/quassel/quassel/pull/180/commits/4aea5769f4c155234394957aa55360f111c12cbb ).

PR #189, opened 2016-02-29
Minor string cleanup

phuzion:
For my first PR to the Quassel project, I here's a handful of minor improvements in the capitalization and formatting of a handful of strings that I figured could use some cleanup.


AlD:
LGTM, thanks!
Changes will show up in https://github.com/quassel/quassel-i18n/commits/tx-sync and subsequently Transifex after the nightly sync.

PR #193, opened 2016-03-23
Handle STATUSMSG messages

mamarley:
Previously PRIVMSGs to channels prefixed with certain characters
(usually "+" or "@") (example "@#quassel") would appear in new
query buffers instead of in the channel buffer.  People were using
that problem to broadcast query SPAM in active channels such as

This patch fixes the issue by detecting the condition, first in
ircparser.cpp where it will use the channel name instead of the
sender's nick as the target.  ctcpparser.cpp then sees the prefixed
channel name and reacts by truncating the prefix(es) and setting a
new flag so that the client can tell the message was a STATUSMSG.
A look-ahead system is used to ensure that the channel prefix
character is not stripped even when the channel prefix characters
and STATUSMSG prefix characters overlap.

The server-provided STATUSMSG prefixes are used if available,
otherwise "@" and "+" are used.

This patch doesn't make the client display the STATUSMSGs any
differently than regular messages.  I omitted this because there
was some discussion on the channel about the best way to display
these messages, but no conclusion was ever reached.  This patch
will at least stop the query SPAM and also provides the
infrastructure necessary to change the display format later.

### DIFF
--- a/src/common/network.h
+++ b/src/common/network.h
@@ -131,6 +131,7 @@ public :
     inline bool isMe(IrcUser *ircuser) const { return (ircuser->nick().toLower() == myNick().toLower()); }
 
     bool isChannelName(const QString &channelname) const;
+    bool isStatusMsg(const QString &target) const;
 
     inline bool isConnected() const { return _connected; }
     //Network::ConnectionState connectionState() const;
--- a/src/common/message.h
+++ b/src/common/message.h
@@ -61,6 +61,7 @@ public:
         Highlight = 0x02,
         Redirected = 0x04,
         ServerMsg = 0x08,
+        StatusMsg = 0x10,
         Backlog = 0x80
     };
     Q_DECLARE_FLAGS(Flags, Flag)
--- a/src/common/network.cpp
+++ b/src/common/network.cpp
@@ -77,6 +77,18 @@ bool Network::isChannelName(const QString &channelname) const
 }
 
 
+bool Network::isStatusMsg(const QString &target) const
+{
+    if (target.isEmpty())
+        return false;
+
+    if (supports("STATUSMSG"))
+        return support("STATUSMSG").contains(target[0]);
+    else
+        return QString("@+").contains(target[0]);
+}
+
+
 NetworkInfo Network::networkInfo() const
 {
     NetworkInfo info;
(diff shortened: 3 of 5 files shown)


AlD:
Excuse my ignorance, I missed the discussions on the channel.

What's the purpose of a STATUSMSG? Is this a freenode thing?
How does it currently look like in Quassel? (I think you may have meant to mention that in the first paragraph, but it was cut off.)
Is there an easy way I can reproduce this?

mamarley:
Sure.  STATUSMSG is a way for people to communicate with people with specific modes (usually voiced or op) in a channel.  Freenode does support it but is not the only network.  In Quassel, STATUSMSGs currently appear in a query buffer to the user who sent the original message, which can be used to broadcast query SPAM to many users at once.  Several people complained about it in #quassel, so I figured I should probably take a stab at fixing it.  A quick way to reproduce it in Quassel is to join two instances of Quassel into the same channel and make sure they are both opped.  Then send "/msg +#channel test" from one.  The "test" message will appear in a query buffer on the other instance.

AlD:
Got it, thanks. Now the usual question - how do other IRC clients handle this? :)

mamarley:
Quassel is really the only client that I have used very much, but the impression I get from talking to other people in #quassel is that most other clients display the messages in the channel buffer but display them differently in some way so that the user can tell the message is a STATUSMSG.  In http://bugs.quassel-irc.org/issues/1399, it is mentioned that at least one other client displays the nick as "<User:#channel>".  My patch doesn't make the message be displayed any differently, but it sets a flag on the message so that someone with a better eye for design can implement that part.

src/common/network.h
```diff
@@ -131,6 +131,7 @@ public :
     inline bool isMe(IrcUser *ircuser) const { return (ircuser->nick().toLower() == myNick().toLower()); }
 
     bool isChannelName(const QString &channelname) const;
+    bool isStatusMsg(const QString &target) const;
```
digitalcircuit: I'd suggest adding some brief documentation here, akin to the style in [aboutdata.h](https://github.com/quassel/quassel/blob/master/src/uisupport/aboutdata.h#L30-L54 ).  See [the Doxygen guide](http://www.stack.nl/~dimitri/doxygen/manual/docblocks.html ) and Sput's remarks from #quassel:
```
<Sput> we use @ instead of \; no @brief but a blank line after the subject (like in a git commit); @param should have [in] or [out] as appropriate
<Sput> and for attributes, ///< is the preferred way for a same-line comment
```
Not required, of course!

AlD: Good call, great collaboration :)

AlD:
Merged, thanks!

PR #194, opened 2016-04-03
Implement Highlight-Only ignore

CounterPillow:
This new strictness allows for only ignoring highlights caused by messages matching a certain ignore rule, not the actual messages themselves.

This can be useful to ignore highlights produced by bots.

### DIFF
--- a/src/common/ignorelistmanager.h
+++ b/src/common/ignorelistmanager.h
@@ -44,7 +44,8 @@ public:
     enum StrictnessType {
         UnmatchedStrictness = 0,
         SoftStrictness = 1,
-        HardStrictness = 2
+        HardStrictness = 2,
+        HighlightOnlyStrictness = 3
     };
 
     enum ScopeType {
@@ -97,7 +98,7 @@ public:
     /** This method checks if a message matches the users ignorelist.
       * \param msg The Message that should be checked
       * \param network The networkname the message belongs to
-      * \return UnmatchedStrictness, HardStrictness or SoftStrictness representing the match type
+      * \return UnmatchedStrictness, HardStrictness, SoftStrictness or IgnoreOnlyStrictness representing the match type
       */
     inline StrictnessType match(const Message &msg, const QString &network = QString()) { return _match(msg.contents(), msg.sender(), msg.type(), network, msg.bufferInfo().bufferName()); }
 
--- a/src/qtui/settingspages/ignorelisteditdlg.ui
+++ b/src/qtui/settingspages/ignorelisteditdlg.ui
@@ -7,8 +7,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>389</width>
-    <height>368</height>
+    <width>390</width>
+    <height>400</height>
    </rect>
   </property>
   <property name="windowTitle">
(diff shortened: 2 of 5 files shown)


CounterPillow:
Apparently, this currently has a case where crashes happen due to a protocol incompatibility or something.

src/common/ignorelistmanager.h
```diff
@@ -44,7 +44,8 @@ class IgnoreListManager : public SyncableObject
     enum StrictnessType {
         UnmatchedStrictness = 0,
         SoftStrictness = 1,
-        HardStrictness = 2
+        HardStrictness = 2,
+        HighlightOnlyStrictness = 3
```
digitalcircuit: The problem arises with modifying this ```StrictnessType``` enum.  ```IgnoreListManager``` is synchronized over the network between the client and core as a ```SyncableObject```, and it doesn't know how to handle the same enum with different entries (newer client and older core, or vice-versa).

Unfortunately, in the current version of the Quassel protocol, you can add new synced functions, but adding to enums breaks protocol compatibility.

You could try synchronizing the strictness in another way, but it may be easier to wait until the next protocol break.  There's consideration of breaking protocol once v0.13 is released in order to implement a system where these sorts of changes don't cause problems.

AlD:
I wonder if it would be a better fit to implement this as a property of the highlights instead of an ignore rule.

Highlights are already in an ordered lists. We could make it more obvious in the UI that this is the case, allow reordering of highlights and never highlight a message once an "ignore" highlight is matched.

Alternatively split the highlights section into two lists, one for highlights, one for ignores and always process ignores first.

Any thoughts?

AlD:
Regarding the protocol break, just bump the protocol version number I guess. Then we'll merge once there's enough justification to actually bump the protocol.

CounterPillow:
I decided to implement this as an ignore rule because it can easily make use of the various matching mechanisms present. Some people want to ignore highlights by message content, some others by the nickname it originates from, etc.

src/qtui/settingspages/ignorelisteditdlg.ui
```diff
@@ -25,6 +25,8 @@
       <widget class="QGroupBox" name="strictnessGroupBox">
        <property name="toolTip">
         <string>&lt;p&gt;&lt;b&gt;Strictness:&lt;/b&gt;&lt;/p&gt;
+&lt;p&gt;&lt;u&gt;Highlight only:&lt;/u&gt;&lt;/p&gt;
+&lt;p&gt;Do not ignore messages, simply ignore highlights caused by them.&lt;/p&gt;
```
AlD: How about: "Messages matching this type of rule don't trigger highlights, but are still shown to the user."

AlD:
Good point.

@Sput42, since you flagged this nmw, is there anything else required in this PR except more changes to justify a protocol bump?

digitalcircuit:
@CounterPillow Good news, @justjanne is including this functionality in the [core-side highlight pull request](https://github.com/quassel/quassel/pull/303 )!

Right now it's called an "inverse" highlight rule, taking priority over other highlight rules (*including nicknames*).  The discussion's still open for making a better UI (*e.g. two panes, one for "highlight" rules", one for "don't highlight" rules*), separate from ignore rules (*those hide messages*).

digitalcircuit:
Belated follow-up for anyone stumbling across this - this functionality is merged and available in Quassel 0.13 (albeit in a slightly different form) via the new core-side highlight/highlight ignore system.

Highlight ignores are configured in `Remote Highlights` instead of the `Ignore List` section, since it affects highlighting instead of removing the message entirely.  (It's actually for technical reasons of new protocol additions, but that's one possible logical explanation.)

CounterPillow:
alright, guess I can close this then

PR #196, opened 2016-05-18
simplify checking if CMake policies need to be set

DerDakon:
Remove the magic knowledge about when a policy was introduced, simply check if the policy exists.


Sput42:
I didn't even know that checking for policy existence is possible, thanks :)

PR #199, opened 2016-05-28
Fix the SSL check with Qt 5.6 and GCC 5

sandsmark:
The check in qglobal.h requires _only_ -fPIC, if -fPIE is defined the check fails for GCC 5 and above. To avoid linking against QtCore we also need to define QT_NO_VERSION_TAGGING.


seezer:
+1

PR #198, opened 2016-05-27
Sync IRCv3, account tracking, CAP NEW/DEL, polish, and more!

digitalcircuit:
## In short
* Support ```CAP NEW``` and ```CAP DEL``` as [required by IRCv3 specs](http://ircv3.net/specs/extensions/cap-notify-3.2.html )
 * Needed to remain [listed as IRCv3.2 compliant](http://ircv3.net/software/clients.html ), used to handle services restarting, etc
* Synchronize IRCv3 capabilities between core and client
 * Backwards-compatible, allows for smarter client suggestions
* Synchronize nick services accounts (NickServ/SASL), part two of [```account-notify```](http://ircv3.net/specs/extensions/account-notify-3.1.html )
 * Now that [pull request #186](https://github.com/quassel/quassel/pull/186 ) is merged, display client-side when available
*  Add Feature flag ```CapNegotiation = 0x0020```
* Support [the ```CHGHOST``` capability](http://ircv3.net/specs/extensions/chghost-3.2.html )
* Polish and fixes

Criteria | Rank | Reason
---------|---------|-------------
Impact | ★★★ *3/3* | User-facing features, foundation for IRCv3 compliance and future
Risk | ★★☆ *2/3* | Difficult to test, many possible edge cases (*risk accepted in #180*)
Intrusiveness | ★★★ *3/3* | Modifies protocol, required for other improvements, may conflict PRs

## Rationale
The reason for ```CAP NEW``` and ```CAP DEL``` are well explained on [the IRCv3 specs page](http://ircv3.net/specs/extensions/cap-notify-3.2.html).  In relation to Quassel, it's necessary to implement in order to avoid breaking spec-compliant servers correctly assuming IRCv3.2-compliant clients support ```CAP NEW``` and ```CAP DEL```.

The approach in #180 of making IRCv3 invisible to clients works, but leaves room for improvement.  Right now, if the client knows what capabilities are supported by the server, it could guide users in setting up matters like account registration, suggesting SASL over NickServ if supported, and warning if ```PLAIN``` or ```EXTERNAL``` is not supported by a given server.

The ```CapNegotiation``` Feature flag technically isn't needed as newer clients can connect to older cores just fine.  However, it may be useful to know if the core supports cap negotiation or account-tracking in the future, e.g. distinguishing between account tracking being unsupported by the core, versus the IRC server simply not providing the needed information.  Most importantly, if 0.13 is released without this flag, there's no way to reliably determine if negotiation is supported.

## Implementation
[...]

PR #202, opened 2016-06-01
Add support for self messages

digitalcircuit:
## In short
* Check if incoming message sender is our nick; if so, treat
as if sent by the core
 * Message decryption is still tried in case the user manually set up identical keys on different clients,
but key negotiation isn't allowed with self-messages
 * Carry over "Self" flag from ```EventManager``` to ```Message```, so incoming
self-messages are processed as if sent by client
* Enable self message support with [ZNC](http://znc.in/)

Criteria | Rank | Reason
Impact | ★★☆ *2/3* | User-facing, better support other chat systems, part of IRCv3 future
Risk | ★☆☆ *1/3* | Messages might possibly be wrongly interpreted as from self

## Rationale
Modern chat systems allow sending messages from multiple clients while keeping chat history in sync.  While Quassel handles this for messages sent via the Quassel core, it currently does *not* handle messages sent from elsewhere.

For example, [Bitlbee](https://www.bitlbee.org/ ) serves as a bridge between IRC and other chat services, and [with self message support](https://wiki.bitlbee.org/SelfMessages ) it can relay conversations held over Jabber, Skype, etc with other clients (*web, mobile, desktop*) back into Quassel.

[ZNC](http://wiki.znc.in/ZNC ) also supports self messages, but requires requesting the capability ```znc.in/self-message```.  (*Included now that [pull request #198](https://github.com/quassel/quassel/pull/198 ) is merged*)

Self message support is required for [IRCv3 ```echo-message```](http://ircv3.net/specs/extensions/echo-message-3.2.html ), which would allow Quassel to determine when messages are received by the server, and how the server decided to present them (*e.g. if color formatting was stripped or words were censored*).

## Example
If the core's nick is ```quasseluser```, then
```
:quasseluser!ex@example.com PRIVMSG dcircuit :hi
```
is handled as if ```quasseluser``` had sent ```hi``` to nick ```dcircuit```.

### DIFF
--- a/src/common/irccap.h
+++ b/src/common/irccap.h
@@ -102,6 +102,24 @@ namespace IrcCap {
     const QString USERHOST_IN_NAMES = "userhost-in-names";
 
     /**
+     * Vendor-specific capabilities
+     */
+    namespace Vendor {
+
+        /**
+         * Self message support, as recognized by ZNC.
+         *
+         * Some servers (e.g. Bitlbee) assume self-message support; ZNC requires a capability
+         * instead.  As self-message is already implemented, there's little reason to not do this.
+         *
+         * More information in the IRCv3 commit that removed the 'self-message' capability.
+         *
+         * https://github.com/ircv3/ircv3-specifications/commit/1bfba47843c2526707c902034b3395af934713c8
+         */
+        const QString ZNC_SELF_MESSAGE = "znc.in/self-message";
+    }
+
+    /**
      * List of capabilities currently implemented and requested during capability negotiation.
      */
     const QStringList knownCaps = QStringList {
@@ -112,7 +130,8 @@ namespace IrcCap {
             EXTENDED_JOIN,
             MULTI_PREFIX,
             SASL,
-            USERHOST_IN_NAMES
+            USERHOST_IN_NAMES,
+            Vendor::ZNC_SELF_MESSAGE
     };
     // NOTE: If you modify the knownCaps list, update the constants above as needed.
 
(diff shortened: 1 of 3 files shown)


digitalcircuit:
Updated to fix case-insensitivity checking by using ```Network::isMyNick()``` instead of a custom comparison.

arcadedroid:
Using self-built core at d48c2331, self-messages from Bitlbee still do not seem to be detected and placed properly.

Example with Bitlbee `self_message` set to `true`, message was placed in `&bitlbee`:
```
<Arcade> arcade__: test
```

Example with Bitlbee `self_message` set to `prefix`, message was placed in `&bitlbee`:
```
<Arcade> set self_messages prefix
<root> self_messages = `prefix'
<arcade__> Arcade: -> test
```

dequis:
Why would you report issues by posting comments in a PR from 2016. It's not even a quassel issue.

arcadedroid:
@dequis

I'm not sure if it's actually an issue, maybe there's some flag that has to be set. But how does it not concern Quassel? This pull request claims to fix a problem Quassel didn't have?

digitalcircuit:
Bitlbee was my test case for `self-message`, and in the `0.13.1` core, it still seems to work today.  If I edit or send a message on Discord's web UI, Bitlbee via `bitlbee-discord` shows the edit as from myself (`digitalcircuit` with the correct black "self" nickname color).

It might be worthwhile testing a Quassel 0.13.1 core, and if that works, there may be a regression.  If that doesn't work, maybe Bitlbee isn't configured correctly, or there's an issue in the self-message implementation.

I do appreciate the effort to raise awareness of the problem.  I'm around in freenode/`#quassel` as `digitalcircuit`, and I think `dx` is there, too.  As @dequis notes, it might be better to troubleshoot there instead of commenting here, as discussion here will email `Sput` and any others who subscribed to this pull request back in 2016.

PR #203, opened 2016-06-06
Fix sending quit messages on core shutdown

digitalcircuit:
## In short
* When shutting down the core, disconnect all sessions' networks first
 * Signal sent via core, session threads, to networks, calling disconnect on each
 * When sockets closed, signal sent back up to ```[Core|Mono]Application``` as appropriate
* Modify monolithic to gracefully shutdown when main window calls quit
 * Disconnect happens in background, window still disappears nearly instantly
* If disconnect not finished within 10 seconds, Quassel immediately quits
 * Same behavior as before, should reduce risk of getting stuck

Criteria | Rank | Reason
Impact | ★★☆ *2/3* | User-facing polish, visible to other clients
Risk | ★★☆ *2/3* | Edge-cases might block normal shutdown
Intrusiveness | ★★☆ *2/3* | Many files, but shouldn't interfere with other pull requests

```
> From this:
<-- dcircuit_dev (quasseldev@hostmask.IP) has quit (Input/output error)
> To this:
<-- dcircuit_dev (quasseldev@hostmask.IP) has quit (Quit: My Message!)
```

## Rationale
This addresses two small matters of polish.

First, Quassel should send a ```QUIT``` when shutting down to make use of the "Quit Reason" in the identity settings.  Having the core shutdown may be one of the most common methods of quitting, whether via closing the monolithic client or ```upstart```, ```systemd```, ```sysvinit```, etc managing the ```quasselcore``` service.  Simply dropping the socket as done now seems somewhat haphazard.

Second, by properly waiting to ```QUIT``` and disconnect, Quassel now logs disconnects in the chat history.  This reduces confusion over what happened, and simplifies distinguishing between a controlled shutdown and a power outage or crash.

## Implementation
* Send ```QUIT``` when core is shutting down, treating it as a non-requested
disconnect (*core will reconnect on next launch*)
 * If sockets aren't closed within 10 seconds, Quassel quits immediately like before
* Add the necessary ```SLOT``` and ```SIGNAL``` calls throughout the chain of ```[Core|Mono]Application``` ↔ ```CoreApplicationInternal``` ↔ ```Core``` ↔ ```SessionThread``` ↔ ```CoreSession``` ↔ ```CoreNetwork```
 * Lets networks be centrally disconnected, then the status signaled back
* Redirect ```SIGTERM``` and ```SIGINT``` to a virtual function in ```quassel.h```, so
CoreApplication and MonolithicApplication can implement custom
clean-up, i.e. calling Core::quitSessions().
* Modify ```mainwin``` in monolithic mode to emit a signal to request quit
 * Window will hide while core shuts down in background
 * Classic client immediately quits as before

*Other approaches are possible; constructive criticism is welcome!*

### Test results

**Success** (*properly sends ```QUIT```*)
* ```Ctrl-C``` ```quasselcore``` in terminal
* ```Upstart```/```systemd``` shutting down ```quasselcore```
* Closing monolithic client with ```Ctrl-Q```/close button/etc (*Win/Linux*)

**Failure** (*behaves as before, dropping socket*)
* Shutting down desktop session (*Win/Linux*) with monolithic client without first closing the window
* Shutting down standalone core on Windows

### Examples
**Unreal 3.2**
```
> Before
<-- dcircuit_dev (quasseldev@hostmask.IP) has quit (Input/output error)
> After
<-- dcircuit_dev (quasseldev@hostmask.IP) has quit (Quit: My Message!)
```
**Freenode**
```
> Before
<-- dcircuit_dev (~quasselde@hostmask) has quit (Remote host closed the connection)
> After
<-- dcircuit_dev (~quasselde@hostmask) has quit (Quit: My Message!)
```

Where "My Message!" is specified in Configure Quassel → IRC → Identities → Advanced → Quit Reason

*Note: Freenode hides quit messages from clients that disconnect soon after connecting.  Stay connected ~10 minutes before testing quit.*

### DIFF
--- a/src/core/coresession.cpp
+++ b/src/core/coresession.cpp
@@ -75,7 +75,8 @@ CoreSession::CoreSession(UserId uid, bool restoreState, QObject *parent)
     _ircParser(new IrcParser(this)),
     scriptEngine(new QScriptEngine(this)),
     _processMessages(false),
-    _ignoreListManager(this)
+    _ignoreListManager(this),
+    _globalDisconnectActive(false)
 {
     SignalProxy *p = signalProxy();
     p->setHeartBeatInterval(30);
@@ -502,6 +503,8 @@ void CoreSession::createNetwork(const NetworkInfo &info_, const QStringList &per
             SLOT(recvMessageFromServer(NetworkId, Message::Type, BufferInfo::Type, const QString &, const QString &, const QString &, Message::Flags)));
         connect(net, SIGNAL(displayStatusMsg(QString)), SLOT(recvStatusMsgFromServer(QString)));
         connect(net, SIGNAL(disconnected(NetworkId)), SIGNAL(networkDisconnected(NetworkId)));
+        // Keep track of network disconnect and clean-up
+        connect(this, SIGNAL(networkDisconnected(NetworkId)), this, SLOT(markNetworkDisconnected(NetworkId)));
 
         net->setNetworkInfo(info);
         net->setProxy(signalProxy());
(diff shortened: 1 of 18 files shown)


seezer:
First things first:
Tremendous effort in all your PRs, props for that alone! :trophy:
You really make it hard for anyone to decline them :smile:

But... Here i suspect that there exists a simpler solution. All untested - so i might just be wrong.

I assume the actual problem is that we end the event loop immediately when we still need it.
The destructors seem to actually do all that is needed for a clean shutdown (~CoreNetwork, ~CoreSession) internally.
There is even a 10s socketClose timer, started in CoreNetwork::disconnectFromIrc (unused in the destructor case) and there are all the lists needed for housekeeping.

Bold, wild *guess*: socket.flush() in/after disconnectFromIrc is enough in most cases

The sockets are buffered and operate async, so CoreNetwork::writeToSocket() doesn't ensure anything is written at all upon returning.
See documentation on [QAbstractSocket::flush()](http://doc.qt.io/qt-4.8/qabstractsocket.html#flush) - it references the event loop. And that is ended as the first step of the current signal handler implementation (QCoreApplication::quit()).

Summing up I see two and a half *possible* solutions not requiring new lists/timers:
* flush the sockets after sending the irc-QUIT and hope for the best (that is: not much has been written to the socket right before so we have a huge msgQueue - will almost always be true)
 * ensure everything is written by blocking APIs(?!)
* Refactor a little and induce destruction of all CoreSessions before actually calling QCoreApplication::quit or more exactly ~SessionThread (which spins its own loop).
The event loop also has to run between disconnectFromIrc and ~CoreNetwork (which destructs the sockets).

I prefer the last one.

Now tell me I'm wrong ;)

src/core/coresession.cpp
```diff
@@ -642,6 +645,47 @@ void CoreSession::globalAway(const QString &msg)
...
+{
+    if (!_globalDisconnectActive) {
+        return;
+    }
+    // Note: Qt automatically handles concurrency via slots and signals
+    // See https://doc.qt.io/qt-4.8/threads-qobject.html#signals-and-slots-across-threads
+    if (_disconnectingNetworks.contains(networkId)) {
+        // Network was disconnected, remove it from the pending list
+        _disconnectingNetworks.removeAll(networkId);
+    }
```
seezer: Just fyi: That is actually the same as just calling removeAll (or probably better: removeOne), as there is no else path anyway, so just remove it if it's there. Saves one iteration of the list. removeOne is worst-case O(n), removeAll is always O(n).

digitalcircuit: Thanks for the heads-up!

I was thinking along the lines of the [IRCv3 handling where I needed to do something before removing it](https://github.com/digitalcircuit/quassel/blob/cc2e93f84316566252a61a6aabe697dd7835e99b/src/common/network.cpp#L734-L743 ) (*incidentally, there's a case of unneeded contains checking right below that*).  For this commit, simply ```.removeOne()``` makes the most sense.

Granted, thanks to your suggestion of a different approach, this entire section of code will most likely be removed.  But it's still nice to learn :)

PR #206, opened 2016-06-08
Try to reconnect to the database multiple times when disconnected

sandsmark:
In case we lose contact with the database, e. g. when it is restarting
during an upgrade, we can't assume it is available immediately, so loop
for a minute trying to reconnect before giving up.

**Not tested much yet.** So should probably not be merged immediately.


romibi:
Building with AppVeyor would work now with msvc2015
(it's merged in this build: https://ci.appveyor.com/project/romibi/quassel/build/37)
But building with Qt4 does not work because of QThread::sleep() beeing "protected"
(see https://travis-ci.org/romibi/quassel/jobs/142208675#L1507)

AlD:
Please reopen when ready for review :)

sandsmark:
if this gets closed I'm going to forget it :-)

PR #213, opened 2016-06-11
Fix a crash with MSVC debug builds.

TheOneRing:
When staticMetaObject.className() was called before qApp->exec(),
the application crashed. This happend only in debug builds.


TheOneRing:
Just a ci problem, the build passed.

PR #214, opened 2016-06-13
Hide network tooltip info when disconnected, simplify markup

digitalcircuit:
## In short
* Remove network information from tooltip when disconnected; show *Not connected* instead
 * Fixes confusing state where tooltip implies network's connected when it's not
* Don't show the server name if empty
* Simplify QML markup
 * Clean up style declaration, redundant ```<qt>``` tags
 * Use *italics* for when away message is unknown, distinguishing from literally being ```(unknown)```

Criteria | Rank | Reason
Impact | ★☆☆ *1/3* | User-facing interface polish
Risk | ★☆☆ *1/3* | Might possibly show less information while connecting to a network
Intrusiveness | ★☆☆ *1/3* | New translation string, shouldn't interfere with other pull requests

### Network tooltips
**Before: additional, confusing information**
Network tooltip showing confusing empty server
**After: simple "Not connected" message**
Network tooltip showing simple Not connected message
### Nick tooltips
**Before: indistinguishable from normal away messages**
Nick tooltip showing regular unknown message
**After: italics used to signify unknown**
Nick tooltip showing italicized unknown message

### DIFF
--- a/src/client/networkmodel.cpp
+++ b/src/client/networkmodel.cpp
@@ -76,7 +76,7 @@ QVariant NetworkItem::data(int column, int role) const
 
 QString NetworkItem::escapeHTML(const QString &string, bool useNonbreakingSpaces)
 {
-    // QString.replace() doesn't guarentee the source string will remain constant.
+    // QString.replace() doesn't guarantee the source string will remain constant.
     // Use a local variable to avoid compiler errors.
 #if QT_VERSION < 0x050000
     QString formattedString = Qt::escape(string);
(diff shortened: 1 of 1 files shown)


src/client/networkmodel.cpp
```diff
@@ -224,7 +224,8 @@ QString NetworkItem::toolTip(int column) const
     Q_UNUSED(column);
     QString strTooltip;
     QTextStream tooltip( &strTooltip, QIODevice::WriteOnly );
-    tooltip << "<qt><style>.bold { font-weight: bold; }</style>";
+    tooltip << "<qt><style>.bold { font-weight: bold; }</style>"
+            << "<qt><style>.italic { font-style: italic; }</style>";
```
Scheirle: * Opens the `<qt>` tag twice
* One `<style>` tag would be cleaner (e.g. `<style>.bold { font-weight: bold; } .italic { font-style: italic; }</style>`)

digitalcircuit: Turns out I made the mistake of double-opening the ```<qt>``` tag in one other area.. which happened to be the style I copy-pasted for this.  I'll need to review more closely in the future.

In the latest update, I've fixed both the ```<qt>``` tag issue, and simplified the ```<style>``` tag in every tooltip.

Thank you very much for catching this!

PR #216, opened 2016-06-15
Add AppData metadata for quassel, quasselclient

TC01:
The freedesktop.org AppData standard defines metadata for GUI applications that wish to show up in various graphical software centers, such as GNOME Software. See the specification for more details: https://people.freedesktop.org/~hughsient/appdata/

Applications without metadata will not show up in GNOME Software, and possibly other graphical package managers. As GNOME Software is the recommended way on Fedora Workstation to install software, this means that without these files users won't be able to find quassel or quasselclient.

This was a first pass at writing this metadata for inclusion in the downstream Fedora packages; let me know if something is not correct or should be modified!

### DIFF
--- a/data/quassel.appdata.xml
+++ b/data/quassel.appdata.xml
@@ -1,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2016 Quassel Project <devel@quassel-irc.org> -->
+<component type="desktop">
+	<id>quassel.desktop</id>
+	<metadata_license>CC0-1.0</metadata_license>
+	<project_license>GPL-2.0 or GPL-3.0</project_license>
+	<name>Quassel</name>
+	<summary>Modern IRC client</summary>
+	<description>
+		<p>
+Quassel IRC is a modern, cross-platform, distributed IRC client, meaning that
+one (or multiple) client(s) can attach to and detach from a central core: much
+like the popular combination of screen and a text-based IRC client such as
+WeeChat, but graphical. In addition to this unique feature, we aim to bring a
+pleasurable, comfortable chatting experience to all major platforms, making
+communication with your peers not only convenient, but also available everywhere.
+		</p>
+		<p>
+And the best of all: It's free - as in beer and as in speech.
+		</p>
+		<p>
+This application is the "monolithic" build of Quassel; it can be used on its own
+without setting up a core.
+		</p>
+	</description>
+	<screenshots>
+		<screenshot type="default">
+			<image>http://quassel-irc.org/files/images/20080914-011743-quasselkde4.preview.png</image>
+			<caption>Quassel, with a dark theme enabled</caption>
+		</screenshot>
+		<screenshot type="default">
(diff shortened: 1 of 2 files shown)


TC01:
Oh, may be worth mentioning this will close bug [1400](http://bugs.quassel-irc.org/issues/1400).

AlD:
Thanks for the initiative, I didn't know about this stuff.

Can you please add information how this can be tested?

The description text is apparently also present in dev-notes/quassel.desc. Could we de-duplicate these descriptions (also data/quassel{,client}.desktop etc.) into a single place? My idea would be to generate the individual files via cmake.

How do translations of these files work?

TC01:
> Can you please add information how this can be tested?

Oh, yes, sorry. You can validate them using the command ```appstream-util``` (which is a package available in at least Fedora and Debian). The documented command to do validation according to the [Fedora packaging guidelines](https://fedoraproject.org/wiki/Packaging:AppData) is:

```appstream-util validate-relax --nonet *.appdata.xml```

They are meant to be installed into ```/usr/share/appdata/```. After installing them (and the .desktop files they depend on for things like the application icon) the applications should appear in GNOME Software (if installed) with the specified metadata.

> The description text is apparently also present in dev-notes/quassel.desc. Could we de-duplicate these descriptions (also data/quassel{,client}.desktop etc.) into a single place? My idea would be to generate the individual files via cmake.

This should probably be possible. The documentation on AppData appears to be a little sparse, unfortunately (```man appstream-util``` is extraordinarily unhelpful), but it looks like there's support for generating AppData from various other formats (possibly including ```*.xml.in``` files; see below), including desktop files. But the generated AppData would be missing things like screenshots, so I'm not sure this is exactly what we want...

> How do translations of these files work?

Various tags support a ```xml:lang="language code"``` option. Here is an example (from the Fedora packaging guidelines linked above) of this for the summary and description.

```
  <summary>A Comic Archive Reader</summary>
  <summary xml:lang="fr">Un Comic Archive Lecteur</summary>
  <description>
    <p>
      Comical is an easy to use and cross-platform CBR and CBZ reader which
      prefetches and caches pages for speed.
    </p>
    <p xml:lang="fr">
      Comical est une CBR et CBZ lecteur facile à utiliser et
      multi-plateforme qui prélectures et caches pages pour la vitesse.
    </p>
    <p>
      Resized images are crisp, and you can view pages one or two at a time.
      Comical is open-source, so feel free to contribute new features! 
    </p>
    <p xml:lang="fr">
      Images redimensionnées sont nettes, et vous pouvez voir une ou deux pages à la fois. 
      Comical est open-source, alors n'hésitez pas à apporter de nouvelles fonctionnalités!
    </p>
  </description>
```

The [AppData specification page](https://people.freedesktop.org/~hughsient/appdata/) mentions something about ```*.xml.in``` files under the question about translation, but I can't seem to find any more information about this. :(

Sput42:
I actually have no clue what we have quassel.desc for :) Deduplicating the two almost identical XML files would be cool though (could do this with CMake indeed, using a template).
I guess the description itself should be updated at some point with more recent information, too.

I'm still inclined to merge this as-is now and pimp it in later commits; any objections?

PR #217, opened 2016-06-16
Optionally verify SSL connection to IRC servers, polish

digitalcircuit:
## In brief
* Optionally verify SSL connection to IRC servers
 * Validates certificates, blocking trivial man-in-the-middle interception
 * When enabling *Use encrypted connection*, new networks also default to verifying SSL
 * No change to upgraded installs
* Sync default server port in server add/edit dialog with ```useSsl``` checkbox
 * Change port ```6667``` → ```6697``` when checking ```useSsl```, revert ```6697``` → ```6667``` when unchecking
 * Can always be changed, existing networks unmodified
 * Port ```6697``` is the new 'sort-of standard' for SSL, [according to Freenode and other networks](https://freenode.net/news/port-6697-irc-via-tlsssl )
* Update preset networks to include SSL variants
 * SSL verification is not enabled by default (no point when non-SSL fallbacks present)

Criteria | Rank | Reason
Impact | ★★★ *3/3* | User-facing security change, foundation for future improvements
Risk | ★★☆ *2/3* | No roll back, but SSL not changed on upgrade, easily opted out
Intrusiveness | ★★★ *3/3* | Database schema change, feature flag, translations, user interface

*This supersedes [pull request #184](https://github.com/quassel/quassel/pull/184 ).*

*After merging this, [pull request #236](https://github.com/quassel/quassel/pull/236 ) will need updated to resolve conflicts.*

## Rationale
Quassel should validate SSL certificates by default to avoid giving a false sense of security with the *Use encrypted connection* checkbox and to reduce concerns in the wake of mass-surveillance and network interceptions by ISPs and governments.

Nearly all browsers and many IRC clients already validate SSL certificates, giving the user choice about their security.  [Let's Encrypt](https://letsencrypt.org/ ) allows any concerned IRC network to get a certificate for free, and if needed one can always uncheck *Verify connection security* to get Quassel's prior behavior.

[ZNC](http://znc.in/ ) implemented similar [SSL certificate validation with a small bounty from concerned users](https://github.com/znc/znc/issues/156 ).

## Implementation
* Optionally verify SSL connection to IRC servers
 * In database, add flag ```sslverify``` to ```server``` table, defaulting to false so upgrades won't break existing setups
 * In protocol, add flag ```sslVerify``` to ```Network::Server```, configuring if SSL errors should be ignored or result in disconnect
 * Add ```VerifyServerSSL``` feature flag so client can tell users about upgrading their core
 * If SSL verification fails and verify enabled, disconnect and trigger a retry, logging the reason
 * If SSL verification fails and verify disabled, log the reason at ```Info``` level and continue connecting
* Add *Verify connection security* checkbox to the client IRC server add/edit dialogs
 * Change tooltip and disable checkbox if the core does not support the feature
 * Connections verified by default when *Use encrypted connection* is checked, can be disabled
 * Change *Use SSL* and *Use secure connection* to *Use encrypted connection* to better reflect what the option does
 * Fix tooltips and other strings in network editing dialogs
* Sync default port in server add/edit dialog with *Use encrypted connection*
 * Change from port ```6667``` to ```6697``` when checking *Use encrypted connection*, revert from ```6697``` to ```6667``` when unchecking
 * Port can be manually changed, stopping the link between checkbox and port
 * Existing networks are unmodified
 * Port ```6697``` is the new 'sort-of standard' for SSL, [according to Freenode and other networks](https://freenode.net/news/port-6697-irc-via-tlsssl )
* Update preset networks to include SSL variants
 * Non-SSL versions still included by default to reduce risk of breaking new setups
 * SSL verification not enabled by default (no point when including non-SSL variants in preset)

## Testing
**Database schema handling**
* SQLite new install - works
* SQLite upgrade - works
* SQLite migrate to Postgres - works
 * ```sudo -u quasselcore quasselcore --configdir=/var/lib/quassel --select-backend=PostgreSQL```
* Postgres new install - works
* Postgres upgrade - works

**Client/core connections**
* New client, new core - works
* New client, old core - works
 * *Verify connection security* checkbox grayed out with explanatory tooltip
* Old client, new core - works
 * *Verify connection security* preserved in general use
 * When editing settings, verify SSL resets to false for all networks
 * Unfortunately no way to use a boolean and tell if the client is old, or actually setting verify SSL to false.  To fix, this would need the work-in-progress ClientFeature system @esainane designed.

### Status buffer messages
**Verify connection security checked**
*Errors are treated like ping timeouts, attempting to reconnect up to the reconnect limit.  This lets Quassel recover from temporary SSL errors*
```
* Connecting to ircsource.baconsvin.org:6697...
* Encrypted connection couldn't be verified, disconnecting since verification is required (Reason: The certificate is self-signed, and untrusted)
* Disconnecting. (Encrypted connection not verified)
* Connection failure: The certificate is self-signed, and untrusted
* Connection failed. Cycling to next Server
* Connecting to ircsource.baconsvin.org:6697...
[...]
```

**Verify connection security unchecked**
```
* Connecting to ircsource.baconsvin.org:6697...
* Encrypted connection couldn't be verified, continuing since verification is not required (Reason: The certificate is self-signed, and untrusted)
* Requesting capability list...
[...]
```

### Client interface
**'Verify connection security', add server dialog**
*New client, new core, 'Use encrypted connection' manually checked*
Add network dialog - new core, use SSL manually checked, tooltip shown

**'Verify connection security', edit server dialog**
*New client, new core, 'Use encrypted connection' not checked*
Edit networks dialog - new core, use SSL unchecked

**Upgrade warning, edit server dialog**
*New client, old core*
Edit networks dialog - old core warning tooltip shown

**Toggling 'Use encrypted connection' checkbox, add server dialog**
*Port only modified when it's at the expected default of ```6667``` for non-SSL, or ```6697``` for SSL*
Add network dialog - toggling Use encrypted connection and effects on chosen port


kalaksi:
The wording is a bit confusing , e.g. "continuing since encryption isn't required". Surely it still uses encryption but ignores the certificate? So, maybe the option names could be " use encrypted connection" and "verify connection security" to make a better distinction (and similarly in some error messages). Tooltip could warn about unchecking the latter.

digitalcircuit:
I'm not certain it's still encrypted [given the ```QSslError::NoSslSupport``` error](https://doc.qt.io/qt-4.8/qsslerror.html ), however, that's probably too confusing to convey to users.

I agree with your suggestions; thank you very much!  I wasn't happy with the original names and tooltips, but I wasn't sure of what to say instead.  This is much nicer.

If anyone's been testing this branch before now, **the database schema has changed ```sslrequired``` to ```sslverify```**.  You'll need to manually update your database.

*Normal upgrades from the ```master``` branch work as before.*

kalaksi:
I'm fine with the messages/text now :+1:. For the tooltip, I'd highlight the importance of verifying the certificate more e.g. by replacing ", increasing security" with ", which is essential for connection security".

One thing I though of is that where does Qt get the CA certificates against which it verifies the server certificate. I did some digging and apparently it should use the system's certificate store. I was thinking that maybe we needed an option for manually selecting the CA certificate, e.g. in case of self-hosted internal CAs and private IRC servers. It would be useful where appropriate installation of the internal CA certificate is impossible, e.g. by having not enough privileges, which could happen if using someone else's core. 

It would probably complicate the UI and should be put somewhere on the settings sub-page and not on "add server/network" screen. Also, I'm not sure if it's too much to add for this PR.

digitalcircuit:
### Tooltip phrasing
I agree with your sentiment, but I'm concerned about the role of the tooltip.  If someone who's not knowledgeable about security wants to connect to a server and can't due to an invalid certificate, they're probably not going to stop to read the tooltip, especially if ```Verify connection security``` isn't enough of a warning (*and it's checked by default for new networks when enabling ```Use encrypted connection```*).

There have been [studies on managing web browser design to stop people clicking through invalid certificate warnings](https://nakedsecurity.sophos.com/2015/02/03/google-redesigns-security-warnings-after-70-of-chrome-users-ignore-them/ ), and all setups have been more complex than a simple checkbox.  Firefox has a big, scary warning that takes four steps to bypass.

Unfortunately, self-signed or otherwise invalid certificates are more common for IRC, and an unverified but encrypted connection still avoids passive snooping.  So long as disabling encryption outright is easily possible, **I don't think Quassel should take as strong of a stance against unverified encryption**.  Security-conscious IRC servers can [implement the 'tls' and 'Strict Transport Security' specifications from IRCv3](http://ircv3.net/irc/ ) to help ensure clients verify their connection.

(*Implementing those specifications in Quassel will take more consideration given the core can [re-]connect without a user around to answer questions*)

### Manual certificate management
Qt has [several ways of managing trusted CA certificates](https://doc.qt.io/qt-4.8/qsslsocket.html#caCertificates ), and I agree Quassel should have some way to manually trust certain certificates and/or certificate authorities.  Though there's privacy risks in using a core on a system where you don't have complete control, this system would be useful if you trust the server administrator, or if you simply want a certificate to be valid for IRC and not everything else (*also consider the monolithic client*).

However, as you suggest, **I think this is better addressed in a follow-up pull request**.  A certificate management system is significantly more complex to implement and carries more risk than simply modifying Quassel to optionally stop ignoring SSL errors.  I don't want certificate verification for 80% of cases blocked while solving the remaining 20%, or however the numbers may be :)

digitalcircuit:
Rebased to incorporate latest master changes and fully alphabetically sort ```sql.qrc```, which should ease merging of future schema changes.

PR #223, opened 2016-07-02
Properly handle 0 seconds in secondsToString()

mamarley:
Previously if 0 seconds was passed in, an empty string would be
returned.  Instead, "0 sec" should be returned.


digitalcircuit:
I encountered this bug for the first time ever today; thanks for the fix!

Reference for anyone curious - notice the timestamps:
**Before**:
```
[9:12:55 pm] * [Whois] Nickname is idling for  (since 2016-09-05 02:12:55 UTC)
```
**After** (*not personally tested*):
```
[9:12:55 pm] * [Whois] Nickname is idling for 0 sec (since 2016-09-05 02:12:55 UTC)
```

PR #224, opened 2016-07-02
Port ignore rules to QRegularExpression

sandsmark:
Also moves more of the creation of regex objects to initialization to improve performance a bit.

### DIFF
--- a/src/common/ignorelistmanager.h
+++ b/src/common/ignorelistmanager.h
@@ -22,7 +22,7 @@
 #define IGNORELISTMANAGER_H
 
 #include <QString>
-#include <QRegExp>
+#include <QRegularExpression>
 
 #include "message.h"
 #include "syncableobject.h"
@@ -53,6 +53,14 @@ public:
         ChannelScope,
     };
 
+    static QString convertFromWildcard(const QString originalRule) {
+        QString pattern(QRegularExpression::escape(originalRule));
+        pattern.replace("\\?", ".");
+        pattern.replace("\\*", ".*");
+        pattern = '^' + pattern + '$';
+        return pattern;
+    }
+
     struct IgnoreListItem {
         IgnoreType type;
         QString ignoreRule;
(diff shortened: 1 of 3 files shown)


sandsmark:
Qt5 only, for obvious reasons, so not ready to be merged until next release.

sandsmark:
Most visible change I guess is when switching buffers for the first time.

Bombe:
Buffer switching time is awesomely low!

src/common/ignorelistmanager.h
```diff
@@ -53,22 +53,66 @@ class IgnoreListManager : public SyncableObject
...
+                pattern = convertFromWildcard(pattern);
+            }
+
+            regEx.setPattern(pattern);
+            regEx.setPatternOptions(QRegularExpression::CaseInsensitiveOption);
+
+            if (!regEx.isValid()) {
+                qWarning() << "Ignore rule" << ignoreRule << "is invalid!";
+            } else {
+                regEx.optimize();
```
romibi: When i tried to build with Qt5.3 I noticed:
optimize() (here and one below) needs Qt>=5.4
http://doc.qt.io/qt-5/qregularexpression.html#optimize

digitalcircuit:
It should be possible to get this in for 0.13 if you want to ```#ifdef {...} #else {...} #endif``` all the changes to ```QRegularExpression```, or create a custom class (*e.g. usage of [```KApplication```/```QApplication``` in ```qtuiapplication.h```](https://github.com/quassel/quassel/blob/0a43227b8cd44625f4881cc1545d42c8c8a4876c/src/qtui/qtuiapplication.h#L24-L28 )*), but merely waiting for the next-next release is understandable, too.

sandsmark:
Yeah, it's not fixing any bugs or really adding any useful features, so for me the extra work and mess in the code it creates to get it working with qt4 is not worth it.

digitalcircuit:
@sandsmark Thank you for making this pull request and providing the initial momentum to tackle this!

I've [created pull request #415](https://github.com/quassel/quassel/pull/415 ) to work on Qt 5 and Qt 4, porting highlight rules and nickname matching alongside ignore rules, adding escaping of wildcard characters, and fixing up some other stuff too.  Hopefully it carries the spirit of this pull request; I've added a thanks to you in it for this reason :slightly_smiling_face:

sandsmark:
thanks!

PR #210, opened 2016-06-10
QtWebEngine

romibi:
If QtWebEngine is available use that instead of QtWebKit.
Building with QtWebkit is still possible.

Introduces the Build Option "WITH_WEBENGINE"

Appveyor Builds:
* MSVC Builds with Qt5.6 with QtWebEngine (x64/x86)
* mingw Build with Qt5.6 but no WebPreview
* mingw Build with Qt5.5 with QtWebKit

(To build with QtWebEngine on Windows Visual Studio is needed)

PR #225, opened 2016-07-02
Disable webkit by default

sandsmark:
Qt WebKit is deprecated and a bit of a walking security hole, so it
makes sense to disable by default.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -76,7 +76,7 @@ if (NOT WITH_KDE)
 endif()
 
 # For this, the feature info is added after we know if QtWebkit is installed
-option(WITH_WEBKIT "WebKit support (for link previews)" ON)
+option(WITH_WEBKIT "WebKit support (for link previews)" OFF)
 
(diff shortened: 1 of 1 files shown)


digitalcircuit:
It's worthwhile noting that ```QtWebKit``` may be staying around after all:  [Phoronix: Qt WebKit Looks To Be Making A Comeback](https://www.phoronix.com/scan.php?page=news_item&px=QtWebKit-Next-Gen ).  From what I've heard, ```QtWebEngine``` is much more difficult for distributions to package.

Granted, I don't know when this revival effort would result in a new release that Quassel can use, so disabling by default *might* still be the better option once @romibi 's [QtWebEngine pull request](https://github.com/quassel/quassel/pull/210 ) is merged.  @mamarley may need to update the Quassel Ubuntu build PPAs to re-enable ```QtWebKit``` for Ubuntu releases without ```QtWebEngine``` if this isn't already done.

*If @mention pings were undesired, pardon, I'll avoid them next time.*

PR #231, opened 2016-07-09
Use Qt5.7 for appveyor.

TheOneRing:
Also use gcc 5.3 shipped by Qt.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -138,7 +138,7 @@ build_script:
     NsisDeployImage $env:APPVEYOR_BUILD_FOLDER\scripts\build\NullsoftInstaller.nsi
 
 environment:
-    QT_VER: 5.6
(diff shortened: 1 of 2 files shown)


digitalcircuit:
Small note, the AppVeyor mingw build broke due to this change - it's not one of those random network issues.  See [the build log](https://ci.appveyor.com/project/Quassel/quassel/build/300/job/5ojk6640fb76q9a0#L144 ), clipped below:

```
Image-Name:  quassel_master_9da068_Qt5.7_mingw49_32
Calling "C:\Qt\5.7\mingw49_32\\bin\qtenv2.bat" ""
The running command stopped because the preference variable "ErrorActionPreference" or common parameter is set to Stop: The system cannot find the path specified.
```

TheOneRing:
Fixed

PR #233, opened 2016-07-14
EnableQt's high-DPI scaling introduced in Qt5.6

romibi:
Just a simple QGuiApplication Attribute …
[http://doc.qt.io/qt-5.6/qtlabscontrols-highdpi.html](http://doc.qt.io/qt-5.6/qtlabscontrols-highdpi.html)

Fixes "too small font in Chat/Nick-List" on high-dpi screens.
Fixes "too big Buttons" and "too big font Chat-View" on non-high-dpi screen if high-dpi-screen is connected.

[https://gyazo.com/ced7bf452e18eb09f8841230401c4819](https://gyazo.com/ced7bf452e18eb09f8841230401c4819)

[https://gyazo.com/92cbfa41e7d705493345bc67e60fcf4c](https://gyazo.com/92cbfa41e7d705493345bc67e60fcf4c)

(The Window-Title-Bar on the non-high-dpi screen is still a bit to big … if someone knows how to fix this …)

### DIFF
--- a/src/common/main.cpp
+++ b/src/common/main.cpp
@@ -177,13 +177,19 @@ int main(int argc, char **argv)
     }
 #endif
 
-#  if defined BUILD_CORE
+#if defined BUILD_CORE
     CoreApplication app(argc, argv);
-#  elif defined BUILD_QTUI
+#elif defined BUILD_QTUI
+# if QT_VERSION >= 0x050600
+    QtUiApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
+# endif
     QtUiApplication app(argc, argv);
-#  elif defined BUILD_MONO
+#elif defined BUILD_MONO
+# if QT_VERSION >= 0x050600
+    MonolithicApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
+# endif
     MonolithicApplication app(argc, argv);
-#  endif
+#endif
 
 #ifndef HAVE_KDE4
     // the non-KDE version parses after app has been instantiated


TheOneRing:
Hm I'd say quassel would need to ship the new breeze icons instead of the oxygen icons else it looks horrible.

romibi:
I don't have a problem with the oxygen icons … but i don't cling to them anymore …
So if someone else makes a PR to switch the icons (or make it choosable) … I don't mind …

romibi:
@TheOneRing there you go: PR #234

TheOneRing:
Qt QtCon I just learned that they recommend using the fusion style on Windows if you enabled HIDPI as the Windows style is broken and looks horrible with HIDPI......

romibi:
so … default style Fusion instead of 'System Default' via `#ifdef Q_OS_WIN`? hm …

TheOneRing:
hm probably...
@Sput42  what do you think?

Sput42:
I don't object to using a different default style on Windows if it fixes things - however I don't use Quassel on Windows, so I rely on you guys to find the best approach :)

me4502:
This PR appears to have caused issues on secondary displays that are High DPI, but aren't scaling at the same factor as the primary display.

This is what it looks like for me on my secondary 1440p display, http://i.imgur.com/RXHJu2j.png
My primary display looks fine, and is 4k.

me4502:
Okay, so it works fine on the 1440p display when the DPI scale is set to the same factor, or a greater factor. However when it's a greater factor, the 4k display is off. So it seems it uses the highest scale factor of all displays.

romibi:
So does this problem exist for you since this PR or since #248 ? (That was merged afterwards.)
I have one 1080p and one 4k screen … and i noticed that i have to restart quassel client for it to work correctly after I connect or disconnect a screen.
But this PR fixes the problem of multiple different scaling factors for me…
(Without this PR either buttons were huge or text was tiny when using different scaling factors.)

You are using Windows 10, right?
Looks like I have to test this a bit further…

me4502:
This one, and yeah - Windows 10.

I'm currently working around it by just running my 1440p screen at the same scale factor as my 4K screen

TheOneRing:
Qt is still pretty broken when the resolution changes or you change the screens.
But if started on a hdpi screen the application is at least readable with this changes.

PR #234, opened 2016-07-16
Make Packaging and switching between multiple Icon Themes possible

romibi:
Always heard of that Breeze Icon-Theme but I'm mainly on Windows and there was only Oxygen packaged with Quassel …

So I decided to make it possible to choose which icon themes get packaged while building and choose which icon theme gets used while running via an option in the settings. (If icon themes get packaged.)

[https://gyazo.com/28cd8ee6bf09e2a0de91bcd2b8d423b2](https://gyazo.com/28cd8ee6bf09e2a0de91bcd2b8d423b2)

I modified the Import Script so that it works with multiple icon themes … but it still works only on unix …

Default for cmake is still packaging "Oxygen only" if not WITH_KDE (and "nothing" with kde).
And Default for the Settings is "\<System Default\>" which has still the fallback to Oxygen.

~~Question: why is "irc-channel-joined" and "irc-channel-parted" in icons/import/blacklisted-icons ?~~
~~Why are the~~ ~~_oxygen_~~ ~~hicolor icons used there … (see bufferlist)~~
I think i found the answer … see below


icons/import/import_theme.pl
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/perl
```
digitalcircuit: When renaming the import script, you removed the execute permissions (*the ```100755 → 100644``` note in Github*).

Was this intentional?  If so, feel free to ignore this remark.  If not, you may need to fix the permissions.  From a quick look, you might be able to do it on Windows [with ```git update-index --chmod=+x <file>```](http://stackoverflow.com/questions/6476513/git-file-permissions-on-windows#13593391 ).

src/qtui/settingspages/appearancesettingspage.cpp
```diff
@@ -98,6 +99,24 @@ void AppearanceSettingsPage::initLanguageComboBox()
...
+# endif
+# if defined WITH_BREEZE
+    ui.iconthemeComboBox->addItem(tr("Breeze Light"), QVariant("breeze"));
+# endif
+# if defined WITH_BREEZE_DARK
+    ui.iconthemeComboBox->addItem(tr("Breeze Dark"), QVariant("breezedark"));
+# endif
+#else
+    ui.iconthemeComboBox->hide();
+    QtUiSettings().remove("IconTheme");
```
digitalcircuit: Rather than clearing the settings on loading the ```Settings``` page (*which might not ever happen*), is it possible to check which icon themes are compiled in during startup?

Say, in ```src/qtui/qtuiapplication.cpp```, you could switch on the stored icon theme...
```cpp
        // Set the icon theme
        if (Quassel::isOptionSet("icontheme")) {
            QIcon::setThemeName(Quassel::optionValue("icontheme"));
        } else if (QtUiSettings().value("IconTheme", QVariant("")).toString() != "") {
            QString _iconTheme = QtUiSettings().value("IconTheme", QVariant("")).toString();
            if (_iconTheme == "oxygen") {
#if defined WITH_OXYGEN
            QIcon::setThemeName(_iconTheme);
#endif
            } else if (_iconTheme == "breeze") {
#if defined WITH_BREEZE
            QIcon::setThemeName(_iconTheme);
#endif
            } else if (_iconTheme == "breezedark") {
#if defined WITH_BREEZE_DARK
            QIcon::setThemeName(_iconTheme);
#endif
            }
            // If all icon themes aren't available, leave the theme name at platform default.
        }

        // In case no icon theme is set, or the chosen theme isn't available
        if (QIcon::themeName().isEmpty()) {
            // Some platforms don't set a default icon theme; chances are we can find our bundled Oxygen theme though
            QIcon::setThemeName("oxygen");
        }
```

Reason being, it seems excessive to reset unavailable values when ignoring works and it might be useful in a different context.  What if someone tries a different build of Quassel, or synchronizes their settings across systems with different icon themes, if that even works?

**I'm not certain the above is the best approach**, either.  Feel free to keep what you have, change it, etc :)

However, if you want to make sure that unavailable icon themes won't be kept in the client settings, you should probably check on each launch, some time before setting the icon theme name in ```src/qtui/qtuiapplication.cpp```.

digitalcircuit:
Ack, sorry, I didn't realize you were still in the midst of pushing changes.  I've left two comments on "an outdated diff" according to Github, but I don't think you changed that code.

Otherwise, seems like a nice idea, and certainly looks a lot better!

romibi:
Thanks for your input…

**About the lost execution permission:**
No that was not intentional. And i have a good idea why that happend:
I edited and tested that script via ssh & sftp on my rpi, but i commited the changes on my windows machine … I'll fix it when i have time to look at the next point …

**About clearing invalid settings:**
Yes i realize the current solution is not perfect. My main goal was to not let users who switch to a version with less/no packaged icon-themes get stuck with no icons…
I'll have a look at it again… 

**About the comments on outdated diffs:**
strange … you must have had the (outdated) diffs open for some time …
But yes … i had to rebase after I realized that i pushed the wrong version of the import script while cleaning up my rpi …

**Other:**
Perhaps we could solve my question regarding "irc-channel-joined" and "irc-channel-parted" (the icons in the chatlists) before i push new changes. I'd like to make Quassel use the icon theme icons for that too … But there must be a reason why it doesn't at the moment

digitalcircuit:
### Script permissions
Import script seems fixed!  I don't know why Github kept my comment on the new diff, so I edited my comment.

### Settings clearing
The new invalid settings approach seems good, too.

Eventually it'd be neat to extend it to allow picking and using system icon themes that aren't compiled into Quassel (*e.g. on Linux*), but I couldn't find any Qt way to determine what icon themes are available, asides from replicating the icon theme searching code in Quassel.  This can be investigated later :)

### Icons
Also noted on the ```irc-channel-joined``` and ```irc-channel-parted``` icons.  I think un-blacklisting the icons makes sense.

### General

**Looks good!  I've not been able to test it, but from a quick look nothing seems amiss.**

Aside, [Github has a way to create 'Task Lists'](https://guides.github.com/features/mastering-markdown/#GitHub-flavored-markdown ) which might be useful in the future, like this:
*Markdown*
```md
- [X] Task 1
- [ ] Task 2
```
*Result*
- [X] Task 1
- [ ] Task 2

*Pardon my slow reply; currently visiting family in a different state*

romibi:
hm …
QIcon::themeSearchPaths() gives you the path were the icons are … (or a list of them)
and the shell command `ls */*.theme | rev | cut -d '/' -f 2 | rev` gives you the list of names …
(if you are in that path)
but yea … i don't really want to invest more time into this

(and i knew about the task lists … but thought it looks strange on big paragraphs)

digitalcircuit:
Agreed.  Could be a possible improvement in the future, but for now, this pull request is rather useful as it is.

(*Ack, right, I didn't try Task Lists with large paragraphs.  Noted!*)

PR #239, opened 2016-07-28
Fix some instances of nicks shown as offline when actually online

digitalcircuit:
## In short
* Disconnect signals from the old ```_ircUser``` before deleting it in ```QueryBufferItem```
 * Prevents ```_ircUser``` from triggering ```removeIrcUser()``` again when it's destroyed out of scope
 * Fixes some cases of client showing nicks as offline when actually online
* Limit ```emit dataChanged()``` to when ```_ircUser``` is removed in ```removeIrcUser()```

Criteria | Rank | Reason
Impact | ★★☆ *2/3* | User-facing fix for nick tracking for certain edge cases
Risk | ★☆☆ *1/3* | Could result in wrong nick being tracked in client query list

## Example
Starting with Quassel connected to Bitlbee, an XMPP account active, and a query open with someone in your friend list, when they go offline, their nick quits as expected.  With this pull request, Quassel now shows the nickname as online but away when Bitlbee sends the follow-up ```away-notify``` message indicating the user's able to receive offline messages.

Specifically, receiving the following messages (*no delay between ```AWAY```s*):
```
:nick!name@gmail.com JOIN :&bitlbee
:nick!name@gmail.com AWAY :Away (Testing)
:nick!name@gmail.com QUIT :Leaving...
:nick!name@gmail.com AWAY :User is offline
```

Before, ```nick``` was wrongly shown as not available.  After, ```nick``` is shown as away with the message ```User is offline```.  (*Bitlbee does this as GTalk/XMPP allow for messaging offline users*)

This may also improve handling of nicks during netsplits or rapid quit and re-join.

### DIFF
--- a/src/client/networkmodel.cpp
+++ b/src/client/networkmodel.cpp
@@ -656,8 +656,22 @@ void QueryBufferItem::setIrcUser(IrcUser *ircUser)
 
 void QueryBufferItem::removeIrcUser()
 {
-    _ircUser = 0;
-    emit dataChanged();
+    if (_ircUser) {
+        // Disconnect the active IrcUser before removing it, otherwise it will fire removeIrcUser()
+        // a second time when the object's destroyed due to QueryBufferItem::setIrcUser() connecting
+        // SIGNAL destroyed(QObject*) to SLOT removeIrcUser().
+        // This fixes removing an active IrcUser if the user had quit then rejoined in a nonstandard
+        // manner (e.g. updateNickFromHost calling newIrcUser, triggered by an away-notify message).
+        disconnect(_ircUser, 0, this, 0);
+
+        // Clear IrcUser (only set to 0 if not already 0)
+        _ircUser = 0;
+
+        // Only emit dataChanged() if data actually changed.  This might serve as a small
+        // optimization, but it can be moved outside the if statement if other behavior depends on
+        // it always being called.
+        emit dataChanged();
+    }
 }
 
 


dukzcry:
Thanks for your fix! Before I've tried `set offline_user_quits false` (an undocumented option of Bitlbee), but better to have it fixed properly in Quassel itself.
The old behavior would be fine though, wouldn't it to be broken - it never updated offline query icon to current status until I reconnect to the core.

By the way, I have a question. Currently in Quassel query list both away and offline Bitlbee correspondents have an away icon. Any way to fix this?

digitalcircuit:
Glad I helped!

Ah, nice.  I had to use ```set offline_user_quits false``` for a different reason (*account-specific for [Discord](https://github.com/sm00th/bitlbee-discord ) as that service generates tons of quits*).  Since it's undocumented, might be worthwhile making a pull request for [Bitlbee's ```commands.xml```](https://github.com/bitlbee/bitlbee/blob/master/doc/user-guide/commands.xml ) to document it :)

As to the Away icon, unfortunately, it's a limitation of IRC - there are only three states (*Online, Away, and Offline*), and offline implies you can't send messages.  Bitlbee sends an [```away-notify```](http://ircv3.net/specs/extensions/away-notify-3.1.html ) message when a user quits with the status of ["User is offline"](https://github.com/bitlbee/bitlbee/blob/2e78f75dc5e33fe6d804e71d999f9cab933afd91/irc_send.c#L273-L274 ), letting the IRC client know you can still send them messages.  You could distinguish between Away and Offline via Quassel's Away message in the tooltip.

I don't think the behavior is widespread, so I'm not sure about modifying Quassel to parse that specific message.  It might be worthwhile checking the [IRCv3 ideas and specifications repositories](https://github.com/ircv3/ ) and the ```#ircv3``` channel on Freenode to see if others have discussed additional status types.

(*I'm not sure on this; it's quite likely I'm overlooking something*)

dukzcry:
Hope that option is not just leftover https://bugs.bitlbee.org/ticket/800#comment:3

How about:
1. Add an option to Bitlbee for not sending away info in WHO and away-notify replies for people who are actually away (so online, not offline). Not sure whether Bitlbee will take this though, never seen other IRC client besides Quassel showing user icons in query list. Its meant you go into channel(s) and see people statuses marked with IRC modes from there.
The downside is that's no more possible to read away messages of people who are actually just away.
2. Parsing "User is offline" message, like you suggested, to mark user with offline icon. That must be an optional per-network setting.

Looking at tooltip is not that bad, but at least for me away message is usually "(unknown)" until I do manual WHOIS on that buddy. Maybe it's a problem of bitlbee-libpurple only.

As for IRCv3 ideas, thanks, I will look into. Additional statuses would be ok if ever added. It also means that using IRC modes in Bitlbee for showing user statuses will be no more needed.

PR #241, opened 2016-08-18
Prevent the spelling highlighter from disabling itself

Scheirle:
* Prevent the spelling highlighter from disabling itself, if there are too many spelling mistakes (KDE).
[Sonnet::Highlighter::setAutomatic(bool)](https://api.kde.org/frameworks/sonnet/html/classSonnet_1_1Highlighter.html#abfa334a6fafbefb78dab715b2ed3da94)
* Also enables spell checking by default if it is not configured.

----

Is the `#ifdef HAVE_KDE` in `qtui/inputwidget.cpp` really necessary?
If we have KDE, Sonnet normally loads the default values. 
Furthermore in my case `s.value("EnableSpellCheck", false)` always returns `false` (But [sonnet] spell checking is set to enabled by default).

----
This PR Should work for KF5 and KDE4, the API is the same, but I could only test it on KF5.

### DIFF
--- a/src/qtui/inputwidget.cpp
+++ b/src/qtui/inputwidget.cpp
@@ -504,12 +504,6 @@ void InputWidget::onTextEntered(const QString &text)
     fmt.clearForeground();
     fmt.clearBackground();
     inputLine()->setCurrentCharFormat(fmt);
-
-#ifdef HAVE_KDE
-    // Set highlighter back to active in case it was deactivated by too many errors.
-    if (ui.inputEdit->highlighter())
-        ui.inputEdit->highlighter()->setActive(true);
-#endif
 }
 
 
--- a/src/uisupport/multilineedit.cpp
+++ b/src/uisupport/multilineedit.cpp
@@ -90,6 +90,14 @@ MultiLineEdit::~MultiLineEdit()
 {
 }
 
+#if defined HAVE_KF5 || defined HAVE_KDE4
+void MultiLineEdit::createHighlighter()
+{
+    KTextEdit::createHighlighter();
+    if (highlighter())
+        highlighter()->setAutomatic(false);
+}
+#endif
 
 void MultiLineEdit::setCustomFont(const QFont &font)
 {
--- a/src/uisupport/multilineedit.h
+++ b/src/uisupport/multilineedit.h
@@ -74,6 +74,9 @@ public:
     inline bool emacsMode() const { return _emacsMode; }
 
     void addCompletionSpace();
+#if defined HAVE_KF5 || defined HAVE_KDE4
+    virtual void createHighlighter() override;
+#endif
 
 public slots:
     void setMode(Mode mode);


src/qtui/inputwidget.cpp
```diff
@@ -103,7 +103,7 @@ InputWidget::InputWidget(QWidget *parent)
 
 #ifdef HAVE_KDE
     s.notify("EnableSpellCheck", this, SLOT(setEnableSpellCheck(QVariant)));
-    setEnableSpellCheck(s.value("EnableSpellCheck", false));
+    setEnableSpellCheck(s.value("EnableSpellCheck", true));
```
digitalcircuit: If you change the default value to true here, you also need to change the default value of [```<bool>false</bool>``` in ```inputwidgetsettingspage.ui```](https://github.com/quassel/quassel/blob/9585227400dd89f8a4d6bbc723f2d36ca0dc1a70/src/qtui/settingspages/inputwidgetsettingspage.ui#L52-L57 ).  Otherwise, the checkbox won't show the default as enabled.

digitalcircuit:
### Preventing automatic disabling
Calling ```setAutomatic``` rather than calling ```setActive``` looks like a nicer approach!  Fixing it so it works with ```HAVE_KF5``` is a nice addition :)

However, did you intend to put the force-enable ```setAutomatic``` call only in ```MultiLineEdit``` and **not** in ```InputWidget```?  This seems like a regression.

### Enabling spell-check by default
*Some comments made on the commit itself:  https://github.com/quassel/quassel/pull/241/commits/6ab28b6d6b71a38b6036fd683de7d9da1f4e1ee3*

Due to how settings are stored, this will change the defaults for anyone's who's not specifically configured their spell-checking settings.  This might be fine, but if instead you'd rather preserve Quassel's old behavior for those upgrading (*avoiding unpleasant surprises*), you could build upon the settings upgrade logic [in pull request #185](https://github.com/quassel/quassel/pull/185 ), whenever that's merged.  This isn't required of course.

### Git practices
I'd suggest limiting the [first line of your commit message to 52 characters with a longer summary below wrapped at 72 characters](https://github.com/erlang/otp/wiki/writing-good-commit-messages ).  This makes the commit messages easier to follow in ```git shortlog``` and such.

Additionally, I'd suggest making a new commit for changing the defaults, so if either change needs tested or rolled back, it won't require undoing both changes.

The AppVeyor build failure seems to be unrelated to your change and should fix itself next time you push a commit.

Scheirle:
Thanks for the feedback.

> Fixing it so it works with HAVE_KF5 is a nice addition

Embarrassed I noted after creating this PR that with KF5 `automatic` is `false` per default. So this would be only a fix for KDE4 (and for this I didn't test it). But setting it anyway to be consistent is ok, I guess.

> However, did you intend to put the force-enable setAutomatic call only in MultiLineEdit and not in InputWidget?

Yes, the `InputWidget` uses internally the `MultiLineEdit` [1] so this should be sufficient.

### Enabling spell-check by default
I investigated this a bit more and changing the default value is not the correct way. See #242. 


[1] https://github.com/quassel/quassel/blob/9585227400dd89f8a4d6bbc723f2d36ca0dc1a70/src/qtui/ui/inputwidget.ui#L271

digitalcircuit:
Happy to help!

> Embarrassed I noted after creating this PR that with KF5 automatic is false per default. So this would be only a fix for KDE4 (and for this I didn't test it). But setting it anyway to be consistent is ok, I guess.

Huh, I didn't realize that either.  Agreed, setting it to be consistent seems reasonable.

> Yes, the InputWidget uses internally the MultiLineEdit [1] so this should be sufficient.

Thank you for explaining, and pardon my confusion.

> I investigated this a bit more and changing the default value is not the correct way. See #242.

You should still be able to enable spell-checking by default for KDE 4 by changing the setting here and in ```inputwidgetsettingspage.ui``` as mentioned in the old commit comment.  However, Quassel 0.13 will be the last version to support Qt 4 and KDE 4, so you might decide to not bother with the extra work.

Both this pull request and pull request #242 look good!

Scheirle:
>However, Quassel 0.13 will be the last version to support Qt 4 and KDE 4, so you might decide to not bother with the extra work.

In that case i do not bother with it.

avih:
> In that case i do not bother with it.

That sounds like a weird-ish decision. 0.13 is not out yet, and 0.12 has been the stable release for ages, therefore judging by the past, 0.13 could be the stable release for w while - and it will still support KDE4, but your patch ignores KDE4?

Unless I misunderstood something?

Scheirle:
> Unless I misunderstood something?

You misunderstood it.
* Only KDE4 users benefit from this patch.
* With KF5 the spelling highlighter does not disable it self anymore (by default).
* My "In that case i do not bother with it." refers to changing if the spell checking should be enabled by default for KDE4 users.

PR #244, opened 2016-09-01
Add backlog index over (bufferid, messageid) in sqlite

ansiwen:
Connecting to the core with a big sqlite storage (>500MB) can become
extremely slow or even time out.  Adding an index over the tuples (bufferid,
messageid) speeds up the connection significantly, because at connection
time a lot of queries over these two values are processed.


ansiwen:
This is my first attempt. I'm not sure if some upgrade code has to be added, or even a new schema version is required.

ansiwen:
continuous-integration/appveyor/pr failure seems not to be related to this PR

digitalcircuit:
Unfortunately, database changes aren't quite this easy :)

Take a look at [the guidelines I added to ```upgradeSchema.sh``` in my "Optionally verify SSL…" pull request](https://github.com/digitalcircuit/quassel/blob/ft-require-server-ssl/src/core/SQL/upgradeSchema.sh#L6-L86 ) (*thanks to @mamarley and others on IRC for their help in figuring this out*).

In this case, you **do** need to upgrade the schema for SQLite, but *not* for Postgres.

From a high-level perspective, you'll want to...

1.  Run ```./upgradeSchema.sh "SQLite"``` in the ```/src/core/SQL``` folder
2.  Add the new ```(bufferid, messageid)``` index to the ```setup_….sql``` scripts
 * You've already done this step with [```setup_110_backlog_idx3.sql```](https://github.com/ansiwen/quassel/blob/0e7526ebd15d2e7a6ac09997c20df4023fdd898c/src/core/SQL/SQLite/18/setup_110_backlog_idx3.sql ), but it may need moved or changed as part of upgrading the schema
3.  Create an upgrade script for SQLite, perhaps ```upgrade_000_alter_backlog_add_idx_msgid.sql```
 * Possibly identical to the modified ```setup_110_backlog_idx3.sql``` script from above - you can use [```upgrade_000_alter_quasseluser_add_passwordversion.sql```](https://github.com/ansiwen/quassel/blob/0e7526ebd15d2e7a6ac09997c20df4023fdd898c/src/core/SQL/SQLite/18/upgrade_000_alter_quasseluser_add_passwordversion.sql ) or the scripts in my pull request as a guideline
4.  Add the new SQL queries to ```src/core/sql.qrc```, update all existing files to account for the schema number change (*don't change the existing ```upgrade_…``` scripts; those stay in place*)
5.  Test **everything**
 * No, really - create a new SQLite database, upgrade an existing one, migrate to Postgres, and create/upgrade Postgres databases
 * Even though you didn't change all of the above, it's worthwhile testing in case there's any weird issues

Database schema changes do **not** allow downgrading.  Make a backup of any important data while testing.

You can also look at all the database changes I made when [modifying the database schema for optionally verifying SSL](https://github.com/quassel/quassel/pull/217/commits ), and for [optional custom rate-limiting](https://github.com/quassel/quassel/pull/236/commits ).

*I might be missing steps or forgetting critical information; feel free to ask others for feedback, too*

ansiwen:
Thanks @digitalcircuit for your detailed answer! Since it is actually not a schema change I hoped there is a simpler workflow, since this feels like overkill for an basically optional index. I would think it is enough to have a check if the index exists, and if not create it. Your case is different, since you actually added a new column in a table. Don't you think?

digitalcircuit:
Hm, right, I agree the case is different.  Unfortunately, I don't know of an easier way to handle this...
* Upgrade the database schema
 * Overhead due to formality, no downgrades (*though other pull requests will also do that if merged*)
 * Is the well-known way things have been done
* Modify Quassel's C++ code to automatically call the SQLite-specific upgrade query
 * Requires checking/calling another place for database migrations, complicating logic - what if it fails?
 * Allows downgrades

(*If anyone reading this knows differently, please correct me!*)

Given the number of connection timeouts reported in Freenode/```#quassel```, it seems worthwhile including for existing installs.  It'd also make sure Quassel can consistently rely on that index being available in case that's ever important in the future.

I'd vote for a schema upgrade to be consistent, but you can also ask in the IRC channel or wait a while for feedback from the developers (*I think ```al```/@AlD has reviewed database changes in the past*).

avih:
I have an sqlite db of about 1.3G, and the connection and backlog fetching are relatively fast (2-5 seconds till everything is ready).

However, when I changed the backlog fetching from "fixed number of messages per buffer" to "All unread messages" (even with relatively low limit), then connection choked the core for few minutes (i.e. it also became non-responsive for other clients etc).

While I don't doubt that adding such index could indeed make things faster, is it possibly that it only/mostly affects backlog fetching of "unread messages per chat", but doesn't matter when fetching fixed number of messages per chat?

ansiwen:
@avih I have my quassel set to "Fixed amount per chat" and it had a huge effect on my database. So actually I know it has an effect in this case, and don't know if it has an effect on "unread messages per chat". Also, if you look at the code, you can see that the query actually does exactly the select on bufferid and messageid.

ansiwen:
So, I can really re-emphasize, that this additional index improved my performance tremendously. I can do the work, but I need clear directions how to do it, because I don't wanna do it twice.

digitalcircuit:
@ansiwen I'd be happy to help with the steps needed to upgrade the database schema to add the index, but @Sput42 and co. would be the ones needing to approve the change.

If you don't get any response here after waiting a while, try poking ```Freenode/#quassel``` whenever ```Sput``` or ```al``` is around?

digitalcircuit:
I quickly checked in on IRC, seems like this would get merged if tests show no issues.  Once ```TC01``` gets their pull request reviewed/merged, we can work out the needed schema change (*it should be similar to [the steps I mentioned above](https://github.com/quassel/quassel/pull/244#issuecomment-244295248 )*).  It shouldn't be too much work :)

Meanwhile, whenever I have the time, I'll try to verify your results with my old Quassel SQLite setup and an artificially-slow virtual machine.  It's not required, but extra evidence that it works will help.

### Freenode chat
```
<digitalcircuit> Sput: Can you offer any feedback on
https://github.com/quassel/quassel/pull/244#issuecomment-251617535
("Add backlog index over (bufferid, messageid) in sqlite"), i.e. if a schema change would
be the preferred approach?
<Sput> digitalcircuit: I don't know enough about the database stuff to judge that really,
however we've had a bunch of schema changes already for the upcoming release, so one more
wouldn't hurt
<Sput> however I want to save TC01 from having to rebase his stuff *again*, so maybe this
should wait until I have reviewed and merged the LDAP stuff?
```

Sput42:
This would need a rebase and testing from someone who knows what they're doing (@digitalcircuit?), then we can get that in.

digitalcircuit:
@Sput42 I'd be up for testing in The Near Future™.  This is SQLite-specific, which cuts the needed test-cases in half.. and I need to test `justJanne`'s branches too :)

@ansiwen Now that the database schema system's been reworked, it might be easier to save your changes and do a commit from scratch - it's much simpler now.

You'd need to...
* Move `SQLite/18/setup_110_backlog_idx3.sql` from this pull request to `SQLite/setup_110_backlog_idx3.sql`
* Add a new schema version folder and upgrade script in `src/core/SQL/SQLite/version/##`
  * E.g. `SQLite/version/23/upgrade_000_create_buffer_idx.sql` with the same contents as your setup script...
```sql
CREATE INDEX backlog_buffer_msg_idx ON backlog (bufferid, messageid)
```
*See [`src/core/SQL/SQLite/version/9/upgrade_020_create_buffer_idx.sql`](https://github.com/quassel/quassel/blob/master/src/core/SQL/SQLite/version/9/upgrade_020_create_buffer_idx.sql ) for reference.*
* Run the `updateSQLResource.sh` script in `src/core/SQL`
  * If you can't run this script (*e.g. Windows without Bash*), you can alphabetically add in the files manually, or just ask someone to run the script for you with the new files :)

For reference on a more advanced schema change, see this commit: https://github.com/quassel/quassel/commit/1d1178f7d73ab19e62798f23c1d0cd5e0b4cdb59 (originally from https://github.com/quassel/quassel/pull/290 ).

digitalcircuit:
@ansiwen Also, if you don't have time in the next week or so, I'd be willing to adopt your changes while still giving you credit (*via setting `GIT_AUTHOR`*), similar to what I did with the "no sender brackets" pull request :)

digitalcircuit:
@ansiwen I've opened [pull request #313](https://github.com/quassel/quassel/pull/313 ) to handle the database schema changes for you.  If anything seems amiss, including improper credit, feel free to let me know!

@Sput42 You should be able to merge the above linked pull request instead.

ansiwen:
@digitalcircuit: that's perfect, thanks! 😊

Sput42:
PR #313 is merged, obsoleting this.

PR #247, opened 2016-09-08
Use unread backlog fetching by default if not changed

digitalcircuit:
## In short
* Use ```PerBufferUnread``` by default for backlog fetching rather than ```PerBufferFixed```
  * With slow channels or multiple PMs, reduces backlog fetched on connect by ignoring old chat
  * Modifies new installs and existing configurations if prior default wasn't changed
  * In most cases, should speed up connecting and show more missed highlights

Criteria | Rank | Reason
Impact | ★★☆ *2/3* | User-facing, faster connects and highlights slightly more reliable
Risk | ★★☆ *2/3* | Can request more messages than fixed, complex SQL, risks timeouts

## Rationale
Quassel prioritizes recent over past, like other modern chat systems.  However, when connecting to the Quassel core, by default the client currently requests 500 messages of the past for all channels and private messages, regardless of age.

Quassel should instead use the ```Unread messages per chat``` backlog requester, reducing bandwidth spent on unneeded past while requesting more chat history for unread channels and private messages that the user generally cares about more.  This also allows highlights to work slightly more reliably, and additional backlog remains available via scrolling up as before.

As this should provide improvements or at least generally cause no trouble, Quassel should also change the default for existing setups where backlog fetching hasn't been configured.

*If this causes issues, the version upgrade system could be used to preserve old defaults for existing installs*

### Before - fixed backlog requester as default
Settings -> Configure Quassel... -> Interface: Backlog Fetching -> Backlog request method, old default of Fixed
### After - unread backlog requester as default
Settings -> Configure Quassel... -> Interface: Backlog Fetching -> Backlog request method, new default of Unread

### DIFF
--- a/src/client/backlogsettings.h
+++ b/src/client/backlogsettings.h
@@ -23,11 +23,16 @@
 
 #include "clientsettings.h"
 
+// For backlog requester types
+#include "backlogrequester.h"
+
 class BacklogSettings : public ClientSettings
 {
 public:
     BacklogSettings() : ClientSettings("Backlog") {}
-    inline int requesterType() { return localValue("RequesterType", 1).toInt(); }
+    inline int requesterType() { return localValue("RequesterType", BacklogRequester::PerBufferUnread).toInt(); }
+    // Default to PerBufferUnread to help work around performance problems on connect when there's
+    // many buffers that don't have much activity.
     inline void setRequesterType(int requesterType) { setLocalValue("RequesterType", requesterType); }
 
     inline int dynamicBacklogAmount() { return localValue("DynamicBacklogAmount", 200).toInt(); }
--- a/src/qtui/settingspages/backlogsettingspage.cpp
+++ b/src/qtui/settingspages/backlogsettingspage.cpp
@@ -23,6 +23,9 @@
 #include "qtui.h"
 #include "backlogsettings.h"
 
+// For backlog requester types
+#include "backlogrequester.h"
+
 BacklogSettingsPage::BacklogSettingsPage(QWidget *parent)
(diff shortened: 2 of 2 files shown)


digitalcircuit:
Unfortunately, SQLite does not behave as consistently with ```Unread messages per chat```, as discovered by ```avih``` (*```Freenode/#quassel```*).  It's possibly an issue with the version of SQLite in Ubuntu 12.04, server configuration, inefficiency in Quassel, or something else.

If anyone wants to look for performance issues, server configuration is as follows:
* Ubuntu 12.04 32-bit
* 2 GB RAM
* Roughly 1.3 GB SQLite database
* 3 connected clients
* ```Unread messages per chat```
 * With regression, this is about 20× slower than ```Fixed amount per chat```

For the time being, I **do not advise merging this pull request** as is until potential performance regressions with SQLite have been fixed.  I'll update/delete this comment when ready.  Alternatively, the configuration upgrade system could only change the default for new installs.

digitalcircuit:
Unfortunately, I've not been able to reproduce this problem with Ubuntu 12.04 or 16.04 in a resource-limited virtual machine.  I might not have the right amount of backlog or other things.

After discussion with ```avih``` on the Freenode channel, the suggested way forward is to try ```Unread``` fetching by default during Quassel's development cycle.  If anyone else has issues, we can revert the default.

As an in-between, the configuration upgrade system could be used to only change the default for new installs.

@Sput42 Do either of these options sound okay?

digitalcircuit:
Now that https://github.com/quassel/quassel/pull/273 has been merged and once https://github.com/quassel/quassel/pull/274 is merged, this might no longer be a performance issue.  Also see the previous comment.

PR #248, opened 2016-09-08
Use Qt::AA_UseHighDpiPixmaps, this will fix some issues and introduce…

TheOneRing:
… new


Nevcairiel:
This could use a bit more of an explanation what it fixes, especially since the commit message reads like its cut-off.

TheOneRing:
Some pixmaps are now cut, the old banner pixmap still is pixelated.
But overall it looks much less pixelated.

Sput42:
Taking the liberty of rewriting the commit message...

PR #249, opened 2016-09-15
Updates old links/urls in the about dialog

Scheirle:
* http://www.oxygen-icons.org seems to have a new owner and is not related to the Oxygen Icons anymore
* http://www.trolltech.com is parked and not related to Qt anymore


Sput42:
Closed in b9e68a1

PR #254, opened 2016-09-25
Disable high DPI scaling on everything but Windows

sandsmark:
The automatic scaling only works well on Windows.

See e. g. e33532e048fa12ee32429ca83ee31aa8f065147d in qttools,
QTBUG-50698 and QTBUG-52318.

### DIFF
--- a/src/common/main.cpp
+++ b/src/common/main.cpp
@@ -186,7 +186,7 @@ int main(int argc, char **argv)
 #if defined BUILD_CORE
     CoreApplication app(argc, argv);
 #elif defined BUILD_QTUI
-# if QT_VERSION >= 0x050600
+# if QT_VERSION >= 0x050600 && defined(Q_OS_WIN)
     QtUiApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
 #endif
 # if QT_VERSION >= 0x050700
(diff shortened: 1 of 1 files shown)


sandsmark:
https://codereview.qt-project.org/#/c/155169/

TheOneRing:
Whats about mac?

sandsmark:
I don't have an OS X machine to test on, so I'm just following Qt which disables it for OS X as well.

src/common/main.cpp
```diff
@@ -186,15 +186,15 @@ int main(int argc, char **argv)
 #if defined BUILD_CORE
     CoreApplication app(argc, argv);
 #elif defined BUILD_QTUI
-# if QT_VERSION >= 0x050600
+# if QT_VERSION >= 0x050600 && defined(Q_OS_WIN)
     QtUiApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
 #endif
 # if QT_VERSION >= 0x050700
```
digitalcircuit: Here and on line 200, should ```Qt::AA_UseHighDpiPixmaps``` also be disabled on non-Windows platforms?

I'm not sure, and it's hard to test as I've not encountered this problem.  You could probably put the ```QT_VERSION``` check for 5.7 inside the 5.6 block.

sandsmark: the only problematic attribute seems to be AA_EnableHighDpiScaling.

AlD:
FWIW doesn't look different to me with or without this PR on an MBP. Should I be looking for something in particular?

sandsmark:
I don't think it affects OS X at all.

PR #255, opened 2016-09-28
Update ru.po

POPSuL:
`строкаввода` without space between `строка` (line) and `ввода` (input) is not correct for Russian language.


AlD:
Thanks for your contribution.

Our translations are managed via Transifex, though. Can you please head over to https://www.transifex.com/ald/quassel/translate/#ru to make your change there?

PR #262, opened 2016-11-15
Fix sasl authentication to fail on servers which don't support sasl

Adam-:
If you connect quassel to a server which doesn't support sasl, but with sasl enabled on the client, you get this:

```
-> :adam.home NOTICE * :*** Looking up your hostname...
-> :adam.home NOTICE * :*** Checking Ident
-> :adam.home NOTICE * :*** No Ident response
-> :adam.home NOTICE * :*** Couldn't look up your hostname
<- CAP REQ :sasl
-> :adam.home CAP * NAK :sasl
<- NICK Adam
<- USER quassel 8 * :Adam
<- AUTHENTICATE PLAIN
(~30 second wait)
-> ERROR :Closing Link: 192.168.1.16 (Registration timed out)
```

You can see this by trying to connect to irc.rizon.net.

This behavior of not requesting CAP LS first to see if sasl is even available (and it isn't) is very bad.

It looks to me like this has all be rewritten in master to be correct, so I've just hacked in a check for CAP NAK coming back into 0.12. I think it should be fixed in 0.12 despite master being fixed because it is preventing users from connecting and is a clearly broken behavior.

Thanks

### DIFF
--- a/src/core/coresessioneventprocessor.cpp
+++ b/src/core/coresessioneventprocessor.cpp
@@ -144,19 +144,25 @@ void CoreSessionEventProcessor::processIrcEventCap(IrcEvent *e)
     // additional CAP messages (ls, multi-prefix, et cetera).
 
     if (e->params().count() == 3) {
-        if (e->params().at(2).startsWith("sasl")) { // Freenode (at least) sends "sasl " with a trailing space for some reason!
-            // FIXME use event
-            // if the current identity has a cert set, use SASL EXTERNAL
+        if (e->params().at(1) == "NAK") {
+            // CAP REQ sasl was denied
+            coreNetwork(e)->putRawLine("CAP END");
+        }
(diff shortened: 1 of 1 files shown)


Lillecarl:
I can confirm this bug, Adam helped me troubleshoot and found why i couldn't connect (thought the server had SASL support). :)

Sput42:
Do we need to forward-port this to master?

Sput42:
Merged, thanks!

digitalcircuit:
This shouldn't need forward-ported; `master` [already handles SASL authentication being missing](https://github.com/quassel/quassel/blob/91db6f78781558105f27a2300309359827e901d9/src/core/coresessioneventprocessor.cpp#L250-L295) (alongside other capabilities).

*Aside, [pull request #267](https://github.com/quassel/quassel/pull/267 ) should make it more obvious to users.*

PR #263, opened 2016-11-23
Split IRC messages on newlines at protocol level

digitalcircuit:
## In brief

* When handling input via protocol, split messages on ```QChar::LineFeed```
  * IRC protocol [does not allow newlines in message](https://tools.ietf.org/html/rfc1459#page-8 )
  * Process and display as individual messages sent one after the other
  * Applies to ```/query```, ```/say```, ```/notice```, and ```/me```
  * Fixes sending multi-line content via copy-paste on Quasseldroid
  * Quassel desktop not affected, sends messages for every line break

Criteria | Rank | Reason
Impact | ★☆☆ *1/3* | User-facing fix for multi-line messages not sending on mobile
Risk | ★★☆ *2/3* | Modifies message sending path, might impact performance or break

*Note: Quasseldroid will still need to change message parsing in order to handle multiple different commands, e.g. ```/me does something \n /notice #quassel A notice```.  This pull request fixes the case for a single command with newlines.*

## Test case

1. Install Quasseldroid (*stable or alpha*), connect to a core
2. Copy some text to the Android clipboard that has a line break in it (not just word-wrap).
3. Paste into Quasseldroid
4. Send message

**Before**
Message shows up with line break (*LF character on desktop*), IRC server treats all lines after first as raw commands.

**After**
Message gets split into multiple smaller messages at each line break.

*This should also work for ```/query```, ```/say```, ```/notice```, and ```/me```.*

### Source text
```
Test 1
Test 2
```

### Before

**Quasseldroid**
Several messages of 'Test 1' followed by awkward new line, then 'Test 2', treated as if it's a single message

**Quassel desktop**
![Several messages of 'Test 1 [LF] Test 2', where [LF] is a single character in a Unicode box](https://dl.dropbox.com/u/839888/Hosting/Utilities/Quassel/Development/pr/fix-split-newline/Quassel%20desktop%20-%20before.png#v1 )

### After

**Quasseldroid**
One message of 'Test 1', followed by a separate message of 'Test 2', no awkward new line

**Quassel desktop**
![One message of 'Test 1', followed by a separate message of 'Test 2', no [LF] character](https://dl.dropbox.com/u/839888/Hosting/Utilities/Quassel/Development/pr/fix-split-newline/Quassel%20desktop%20-%20after.png#v1 )

### DIFF
--- a/src/core/coreuserinputhandler.cpp
+++ b/src/core/coreuserinputhandler.cpp
@@ -30,6 +30,13 @@
 #  include "cipher.h"
 #endif
 
+#if QT_VERSION < 0x050000
+// QChar::LineFeed is Qt 5
+static const QChar QCharLF = QChar('\n');
+#else
+static const QChar QCharLF = QChar::LineFeed;
+#endif
+
 CoreUserInputHandler::CoreUserInputHandler(CoreNetwork *parent)
     : CoreBasicHandler(parent)
 {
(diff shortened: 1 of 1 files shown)


AlD:
We should probably also strip \r in front of \n.

Could you ifdef include the QChar constants at file scope so we don't miss this when ditching Qt4?

digitalcircuit:
I've done both, removing carriage returns via [```QString.trimmed()```](https://doc.qt.io/qt-4.8/qstring.html#trimmed ) and replaced the ```\n``` constants with a static ```QChar``` constant at the top of the file.

```c++
#if QT_VERSION < 0x050000
// QChar::LineFeed is Qt 5
static const QChar QCharLF = QChar('\n');
#else
static const QChar QCharLF = QChar::LineFeed;
#endif
```

AlD:
Thanks!

PR #208, opened 2016-06-09
Reload SSL certificates on signal SIGHUP

digitalcircuit:
## In brief
* Catch SIGHUP, use it to reload configuration (SSL certs)
 * Similar to how [```nginx```](https://www.nginx.com/resources/wiki/start/topics/tutorials/commandline/ ) and other server programs handle it
 * Allows easy automation of reloading certificates, important with services such as
Let's Encrypt
* If reloading certificates fails, keep old certificates to not disrupt new connections

Criteria | Rank | Reason
---------|---------|-------------
Impact | ★★☆ *2/3* | Less prominent feature, helps uptime for high-availability cores
Risk | ★☆☆ *1/3* | Edge-cases might interfere with loading certificates
Intrusiveness | ★☆☆ *1/3* | Small set of changes, shouldn't interfere with other pull requests

## Rationale
[Let's Encrypt](https://letsencrypt.org/), a free, automated certificate authority, issues certificates that last for up to 90 days.  To make use of their certificates, Quassel must either restart, or have a manner to reload.

Ability to swap out certificates on the fly may also help in environments where uptime more of a concern, such as hosting for tens or hundreds of users.

## Examples
*Command or input by me* ```> this```
*Otherwise, Quassel log output*
```
> ./quasselcore
2016-06-08 16:00:01 Info: Restoring previous core state... 
> killall -SIGHUP quasselcore
2016-06-08 16:00:12 Info: Caught signal 1 - reloading configuration 
2016-06-08 16:00:12 Info: Successfully reloaded configuration 
> mv quasselCert.pem quasselCert-notvisible.pem
> killall -SIGHUP quasselcore
2016-06-08 16:00:22 Info: Caught signal 1 - reloading configuration 
2016-06-08 16:00:22 Warning: SslServer: Certificate file /PATH/quasselCert.pem does not exist 
2016-06-08 16:00:22 Warning: SslServer: Unable to reload certificate file, reverting
           Quassel Core will use the previous key to provide SSL for client connections.
           Please see http://quassel-irc.org/faq/cert to learn how to enable SSL support. 
> mv quasselCert-notvisible.pem quasselCert.pem
> killall -SIGHUP quasselcore
2016-06-08 16:00:31 Info: Caught signal 1 - reloading configuration 
2016-06-08 16:00:31 Info: Successfully reloaded configuration 
> [Ctrl]-[C]
2016-06-08 16:00:33 Warning: Caught signal 2 - exiting. 
[...]

PR #264, opened 2016-11-27
Make reloadCerts also handle the IPv6 server

wjp:
`Core::reloadCerts` (from #208) was only calling `reloadCerts` on `_server`, instead of on both `_server` and `_v6server`.


digitalcircuit:
Thanks for catching this!  I completely overlooked IPv6 when writing the certificate reloading code as I don't (yet) have an IPv6-capable dynamic DNS setup.

PR #266, opened 2016-12-03
Optionally color nicks, new colors, auto-locale timestamp, fixes

digitalcircuit:
## In short
* Optionally use system locale for chat timestamp, using [```QLocale::system()```](https://doc.qt.io/qt-5/qlocale.html#system )[```::timeFormat()```](https://doc.qt.io/qt-5/qlocale.html#timeFormat )
  * 12-hour: ``` h:mm:ss ap```, 24-hour: ``` hh:mm:ss```
  * Old style kept on upgrades, can still do custom
* New default sender colors!
  * Now 16 instead of 8, all loosely readable and distinguishable (subjective)
  * Based on algorithm suggested for [Pidgin](https://www.pidgin.im ) by [lobais](https://developer.pidgin.im/ticket/4534#comment:4 )
  * Old colors kept for those on ```master```, otherwise no effect
  * Add ```sender-color-[hash]``` to valid palette color roles, simplifying themes
* Optionally color nicknames in ```join```, ```part```, ```quit```, nick change, etc
  * Matches Quasseldroid, Quassel Web, and other chat programs
  * Enabled by default, old behavior kept on upgrades, can be disabled
* New default command colors
  * Notices, modes, and join/part/quit now gray
  * Errors now red, unique from server messages as suggested in [pull request #267](https://github.com/quassel/quassel/pull/267 )
  * Keeping old color on upgrade not easily possible as it'd change other colors
* Fix ```Self``` flag not being set on various messages

Criteria | Rank | Reason
---------|---------|-------------
Impact | ★★☆ *2/3* | Changes to default Quassel look, locale-aware behavior, fixes storage
Risk | ★★☆ *2/3* | Protest of defaults, possible theme conflicts, core processing changes
Intrusiveness | ★★☆ *2/3* | Theme engine changes, core change, may interfere with other PRs

## Rationale
Modern chat clients should respect the user's language preferences (*locale*), and also emphasize important information while downplaying the less relevant details.  In casual conversation, it matters more who joins or leaves, not the details of their connection.  Quasseldroid and Quassel Web already take this approach.

All changes are optional for those who prefer the older style.  Upgrades will keep the older look by default.

*Note: Rationale follows [pull request #185](https://github.com/quassel/quassel/pull/185 ).  Changes to defaults can be reverted if needed.  I am not a designer; some of these modifications are subjective.*  **I'm open to discussion!**

### Modern (default for new installs)
*Nicknames colored, new message colors, sender brackets hidden, sender colors enabled*
[...]

PR #267, opened 2016-12-04
Encourage SASL over NickServ when available, tidy up

digitalcircuit:
## In short
* Suggest SASL in place of NickServ
  * Place ```Use SASL Authentication``` above ```Auto Identify```
  * Mark SASL as ```Supported by network``` when connected and server supports it
  * Add ```Details...``` button to explain why SASL is or is not supported
  * Add explanations in tooltips
  * All combinations still allowed, just a visual design change
* Tidy up capability handling code
* Fix SASL appearing active when ```Use SASL Authentication``` was unchecked

Criteria | Rank | Reason
Impact | ★★☆ *2/3* | User-facing, clarifies, answers "How do I identify before joining?"
Risk | ★★☆ *2/3* | Misleading wording if server capabilities aren't detected correctly

## Rationale
In most cases, SASL authentication offers a better experience than NickServ identification.  As authentication happens before joining channels, channels requiring registration won't block attempts and hostmask cloaking will apply to ```JOIN``` messages.  Unfortunately, not all servers support SASL.

Quassel should make use of the IRCv3 capability list sent from the core to client in [pull request #198](https://github.com/quassel/quassel/pull/198 ) to check if SASL is available or not, then update the settings interface accordingly.

As networks may be disconnected or not accurately specify support for SASL (*e.g. temporary unavailability*), Quassel should not force configuring SASL one way or the other.

## Implementation
* Suggest SASL in place of NickServ
  * Rearrange the Network ```Auto Identify``` tab to place ```Use SASL Authentication``` at the top.  People tend to prioritize items shown first.
  * Add new row with icon and label saying ```Supported by network```, ```Not currently supported by network``` (*uses warning icon*), ```Cannot check if supported when disconnected```, or ```Could not check if supported by network```
  * Add new button ```Details...``` to explain each of the above states, including suggestions
  * Add tooltips to authentication options, including recommending SASL if you need to identify before joining channels (*this gets asked a lot in ```#quassel```*).
  * All past combinations of SASL and NickServ-based authentication are still allowed
  * *Also handles capability updates while the settings dialog is open!  That was.. interesting to figure out.*
* Tidy up capability handling code
  * Move ```IrcCap::SaslMechs::maybeSupported``` to ```Network::saslMaybeSupports```, simplifying usage
  * Add ```capAvailable``` to ```Network```, similar to ```capEnabled```, useful for showing elsewhere whether or not a certain capability could've been requested
  * Hide SASL EXTERNAL notice while the client's still loading
* Fix SASL appearing active when ```Use SASL Authentication``` was unchecked
  * Avoid conflict between ```sslUpdated()``` and ```sasl``` QGroupBox state
  * Simplify widget enabling/disabling by using an invisible frame

Explanations shown via ```Details...``` button:
* Messages
  1.  **SASL supported by network**:  The network "{network name}" supports SASL.  In most cases, you should use SASL instead of NickServ identification.
  2.  **SASL not currently supported by network**:  The network "{network name}" does not currently support SASL.  However, support might be added later on.
  3.  **Cannot check if SASL supported when disconnected**:  Quassel cannot check if "{network name}" supports SASL when disconnected.  Connect to the network, or try using SASL anyways.
  4.  **Could not check if SASL supported by network**:  Quassel could not check if "{network name}" supports SASL.  This may be due to unsaved changes or an older Quassel core.  You can still try using SASL.
* Bottom text, same for all
  * *SASL is a standardized way to log in and identify yourself to IRC servers.*

### After - SASL placed above, encouraged
Network settings dialog on client showing the Auto Identify tab.  'Use SASL Authentication' is placed above 'Auto Identify', icon and label indicate it's preferred

### Before - SASL placed below as if secondary
Network settings dialog on client showing the Auto Identify tab.  'Auto Identify' is placed above 'Use SASL Authentication', no icon and label to indicate it's preferred

### SASL messages when supported
*Server announces support for SASL via ```CAP LS```, and if specific support is advertised, ```SASL PLAIN``` is supported*
Network settings dialog on client showing the Auto Identify tab.  At the bottom of the 'Use SASL Authentication' box, an information icon procedes the message 'Supported by network'.  The 'Details...' button shows message #1 from Implemention.

### SASL messages when unsupported
*Server did not announce support for SASL via ```CAP LS```, or if specific support is advertised, ```SASL PLAIN``` was not supported*
Network settings dialog on client showing the Auto Identify tab.  At the bottom of the 'Use SASL Authentication' box, a warning icon procedes the message 'Not currently supported by network'.  The 'Details...' button shows message #2 from Implemention.

### SASL messages when disconnected
*Not connected to server*
Network settings dialog on client showing the Auto Identify tab.  At the bottom of the 'Use SASL Authentication' box, an information icon procedes the message 'Cannot check if supported when disconnected'.  The 'Details...' button shows message #3 from Implemention.

### SASL messages when unknown
*Network was added without being saved, core does not support capability negotiation, etc*
Network settings dialog on client showing the Auto Identify tab.  At the bottom of the 'Use SASL Authentication' box, an information icon procedes the message 'Could not check if supported by network'.  The 'Details...' button shows message #4 from Implemention.

### All Auto Identify tooltips
* **Use SASL Authentication:** Authenticate using your nickname and password before joining any channels
  * **Account:** Account name, often the same as your nickname
  * **Password:** Account password
* **Auto Identify:** Authenticate to services using your password.  Use SASL instead to identify before joining channels.
  * **Service:** Service user to send your password to, usually NickServ
  * **Password:** Account password

Network settings dialog on client showing all tooltips in the Auto Identify tab.

### DIFF
--- a/src/common/irccap.h
+++ b/src/common/irccap.h
@@ -141,22 +141,6 @@ namespace IrcCap {
      * http://ircv3.net/specs/extensions/sasl-3.1.html
      */
     namespace SaslMech {
-
-        /**
-         * Check if the given authentication mechanism is likely to be supported.
-         *
-         * @param[in] saslCapValue   QString of SASL capability value, e.g. capValue(IrcCap::SASL)
-         * @param[in] saslMechanism  Desired SASL mechanism
-         * @return True if mechanism supported or unknown, otherwise false
-         */
-        inline bool maybeSupported(const QString &saslCapValue, const QString &saslMechanism) { return
-                    ((saslCapValue.length() == 0) || (saslCapValue.contains(saslMechanism, Qt::CaseInsensitive))); }
-        // SASL mechanisms are only specified in capability values as part of SASL 3.2.  In
-        // SASL 3.1, it's handled differently.  If we don't know via capability value, assume it's
-        // supported to reduce the risk of breaking existing setups.
-        // See: http://ircv3.net/specs/extensions/sasl-3.1.html
-        // And: http://ircv3.net/specs/extensions/sasl-3.2.html
-
         /**
          * PLAIN authentication, e.g. hashed password
          */
--- a/src/common/network.cpp
+++ b/src/common/network.cpp
@@ -243,6 +243,14 @@ QString Network::support(const QString &param) const
 
 bool Network::saslMaybeSupports(const QString &saslMechanism) const
 {
+    if (!capAvailable(IrcCap::SASL)) {
+        // If SASL's not advertised at all, it's likely the mechanism isn't supported, as per specs.
+        // Unfortunately, we don't know for sure, but Quassel won't request SASL without it being
+        // advertised, anyways.
+        // This may also occur if the network's disconnected or negotiation hasn't yet happened.
+        return false;
+    }
+
     // Get the SASL capability value
     QString saslCapValue = capValue(IrcCap::SASL);
     // SASL mechanisms are only specified in capability values as part of SASL 3.2.  In SASL 3.1,
--- a/src/common/network.cpp
+++ b/src/common/network.cpp
@@ -241,6 +241,20 @@ QString Network::support(const QString &param) const
 }
 
 
+bool Network::saslMaybeSupports(const QString &saslMechanism) const
+{
+    // Get the SASL capability value
+    QString saslCapValue = capValue(IrcCap::SASL);
+    // SASL mechanisms are only specified in capability values as part of SASL 3.2.  In SASL 3.1,
+    // it's handled differently.  If we don't know via capability value, assume it's supported to
+    // reduce the risk of breaking existing setups.
+    // See: http://ircv3.net/specs/extensions/sasl-3.1.html
+    // And: http://ircv3.net/specs/extensions/sasl-3.2.html
+    return (saslCapValue.length() == 0)
+            || (saslCapValue.contains(saslMechanism, Qt::CaseInsensitive));
+}
+
+
 IrcUser *Network::newIrcUser(const QString &hostmask, const QVariantMap &initData)
 {
     QString nick(nickFromMask(hostmask).toLower());
--- a/src/common/network.h
+++ b/src/common/network.h
@@ -39,6 +39,9 @@
 #include "ircuser.h"
 #include "ircchannel.h"
 
+// IRCv3 capabilities
+#include "irccap.h"
+
 // defined below!
 struct NetworkInfo;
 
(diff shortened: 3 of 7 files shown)


CarrotEngineer:
The documentation for this PR is great as per usual!

I'm a bit unsure about the "might not work" language. In the context of the PR / people knowledgeable about the protocol you probably know why, but I can imagine being frustrated as a general user, about not knowing why it doesn't work.

It's effectively showing whether the server advertises SASL or not, so I suspect having "preferred" show up when the server has actively advertised SASL would be enough - no "might not work" case at all.

digitalcircuit:
After discussion in ```#quassel```, rather than try to explain the nuances of SASL availability in the tiny space of the ```QGroupBox```, it might be better to just mark it as ```... (supported by server)``` when available and keep the default text otherwise.  Placing SASL above Auto Identify might do enough to imply it's preferred.

Instead, the [default colors pull request](https://github.com/quassel/quassel/pull/266 ) should change the ```Error message``` color to something different than the ```Server message``` color, so the [```SASL authentication currently not supported by server``` message](https://github.com/quassel/quassel/blob/b5b598dd7fbd64a72ffdeaa6df39a97100dd2810/src/core/corenetwork.cpp#L1221-L1223 ) stands out more.

AlD:
Great PR as always.

Please don't modify the title in the settings widget, though. I don't think people usually expect this to be dynamic content and may find it confusing.
I'd suggest to add ℹ (INFORMATION SOURCE, U+2139)/⚠ (WARNING SIGN, U+26A0) or something more Qt-appropriate with a dynamic tooltip. Even a plain old button labelled "check current server" creating a dialog would work imho.

The tooltip/dialog should include texts like

> ℹ SASL authentication is supported by the IRC server you are currently connected to (irc.server.name).

> ℹ Quassel has not been able to detect if this network supports SASL authentication. This may be because Quassel is not currently connected to it.

> ⚠ The IRC server you are currently connected to (irc.server.name) does not seem to support SASL authentication.

digitalcircuit:
Agreed!  I'll try to look for any Qt UI patterns to mimic and experiment with this in the next few days.

I'd lean towards a solution that's visible and simple by default to encourage use, but allows you to get more information if you want it.  Perhaps a one-liner similar to...

* :information_source: Supported by ```${server name}```
* :information_source: Unknown or disconnected from ```${server name}```
* :warning: Possibly not supported by ```${server name}```

But with better phrasing, icons from system theme, and a ```More details...``` button that pops up a dialog with longer text akin to what you've written, maybe even a link to the IRCv3 page.

digitalcircuit:
I've implemented a simple label and icon with a ```Details...``` button, updated the pull request text and screenshots, and fixed a preexisting related bug, too (*SASL appeared enabled on first load when ```Use SASL Authentication``` was disabled*).

Let me know if you want any changes to the text.  I tried to come up with simple, good explanations, but I'm almost certain they can be improved.

digitalcircuit:
Quick update - fixed a regression with handling ```Use SASL Authentication``` enabling and disabling, and removed the link to the IRCv3 specifications as that's sort-of useless to those who aren't likely already familiar with IRC internals.  It'd be better to link to something like a wiki page on Quassel.

digitalcircuit:
Removed some mentions of authentication since SASL itself refers to authentication (*Simple Authentication and Security Layer*), and also makes the dialogs simpler to read.  To avoid resetting translations and maintain some hint of purpose, the ```Use SASL Authentication``` checkbox was *not* renamed.

Thanks to ```SaberUK``` and others on ```Freenode```/```#ircv3``` for the suggestion and discussion.

PR #268, opened 2016-12-06
Sort sql.qrc when generating, emphasize automated script

digitalcircuit:
## In short
* Use version sorting when generating ```sql.qrc```
  * Makes ```git diff``` output easier to understand
  * More readable, maybe slightly faster, too
* Move ```UPDATE_SQL_RESOURCE.sh``` to ```SQL/updateSQLResource.sh```
  * Easier to find right next to the other SQL management script
* Add documentation to ```upgradeSchema.sh``` and ```updateSQLResource.sh```
  * Mention the update resource script, but keep the manual instructions, too

Criteria | Rank | Reason
Impact | ★☆☆ *1/3* | Dev. quality-of-life improvement, better repository history
Risk | ★☆☆ *1/3* | Easily reverted, might not work on some dev. systems

*I really wish I would've known about this script earlier...*

### Before
*Order of ```sql.qrc``` depends entirely on filesystem traversal*
```
<!DOCTYPE RCC><RCC version="1.0">
<qresource>
    <file>./SQL/PostgreSQL/15/upgrade_000_alter_buffer_add_markerlinemsgid.sql</file>
    <file>./SQL/PostgreSQL/16/upgrade_000_alter_network_add_sasl.sql</file>
    <file>./SQL/PostgreSQL/17/upgrade_000_alter_quasseluser_add_passwordversion.sql</file>
    <file>./SQL/PostgreSQL/18/upgrade_000_alter_ircserver_add_sslverify.sql</file>
    <file>./SQL/PostgreSQL/19/delete_backlog_by_uid.sql</file>
[...]
    <file>./SQL/PostgreSQL/19/upgrade_003_alter_network_add_ratelimit_unlimited.sql</file>
    <file>./SQL/SQLite/1/upgrade_020_update_schemaversion.sql</file>
    <file>./SQL/SQLite/1/upgrade_000_drop_coreinfo.sql</file>
    <file>./SQL/SQLite/1/upgrade_010_create_coreinfo.sql</file>
    <file>./SQL/SQLite/15/upgrade_000_fix_network.sql</file>
    <file>./SQL/SQLite/15/upgrade_000_fix_ircservers.sql</file>
    <file>./SQL/SQLite/8/upgrade_010_alter_buffer_add_key.sql</file>
    <file>./SQL/SQLite/8/upgrade_020_alter_buffer_add_joined.sql</file>
    <file>./SQL/SQLite/8/upgrade_030_update_buffer_set_joined_for_channels.sql</file>
    <file>./SQL/SQLite/8/upgrade_000_alter_network_add_connected.sql</file>
    <file>./SQL/SQLite/6/upgrade_010_set_statusbuffertype.sql</file>
[...]
</qresource>
</RCC>
```

### After
*Order of ```sql.qrc``` is consistent*
```
<!DOCTYPE RCC><RCC version="1.0">
<qresource>
    <file>./SQL/PostgreSQL/15/upgrade_000_alter_buffer_add_markerlinemsgid.sql</file>
    <file>./SQL/PostgreSQL/16/upgrade_000_alter_network_add_sasl.sql</file>
    <file>./SQL/PostgreSQL/17/upgrade_000_alter_quasseluser_add_passwordversion.sql</file>
    <file>./SQL/PostgreSQL/18/upgrade_000_alter_ircserver_add_sslverify.sql</file>
    <file>./SQL/PostgreSQL/19/delete_backlog_by_uid.sql</file>
[...]
    <file>./SQL/PostgreSQL/19/upgrade_003_alter_network_add_ratelimit_unlimited.sql</file>
    <file>./SQL/SQLite/1/upgrade_000_drop_coreinfo.sql</file>
    <file>./SQL/SQLite/1/upgrade_010_create_coreinfo.sql</file>
    <file>./SQL/SQLite/1/upgrade_020_update_schemaversion.sql</file>
    <file>./SQL/SQLite/2/upgrade_000_drop_buffergroup.sql</file>
    <file>./SQL/SQLite/2/upgrade_010_update_schemaversion.sql</file>
    <file>./SQL/SQLite/3/upgrade_000_update_backlog_flags.sql</file>
    <file>./SQL/SQLite/3/upgrade_010_update_schemaversion.sql</file>
    <file>./SQL/SQLite/4/upgrade_000_rename_buffertable.sql</file>
    <file>./SQL/SQLite/4/upgrade_010_create_buffertable.sql</file>
    <file>./SQL/SQLite/4/upgrade_020_copy_buffertable.sql</file>
    <file>./SQL/SQLite/4/upgrade_030_drop_oldbuffertable.sql</file>
    <file>./SQL/SQLite/4/upgrade_040_create_buffer_idx.sql</file>
    <file>./SQL/SQLite/4/upgrade_050_create_buffer_cname_idx.sql</file>
[...]
</qresource>
</RCC>
```


src/core/SQL/updateSQLResource.sh
```diff
@@ -0,0 +1,38 @@
...
+#
+# NOTE: In most cases you must upgrade the database schema version to make a
+# change, which includes adding upgrade scripts and modifying existing setup
+# queries.  See 'upgradeSchema.sh' for details.
+
+# SQL resource file
+SQL_RES_FILE="../sql.qrc"
+
+# Name of the current directory
+SQL_DIR_NAME=$(printf '%s\n' "${PWD##*/}")
```
AlD: I think this should be used to make $SQL_RES_FILE absolute.

digitalcircuit: Good point.  Furthermore, I think it shouldn't depend on the working directory, and instead use the location of the script itself.  I'll try that now.

digitalcircuit:
I've updated the script according to feedback from @AlD, and it now works when run from any working directory.

For example, 
```sh
cd quassel/src
./core/SQL/updateSQLResource.sh
```
...works the same as...
```sh
cd quassel/src/core/SQL
./updateSQLResource.sh
```

src/core/SQL/updateSQLResource.sh
```diff
@@ -0,0 +1,45 @@
...
+# change, which includes adding upgrade scripts and modifying existing setup
+# queries.  See 'upgradeSchema.sh' for details.
+
+set -u # Bash - fail on undefined variable
+
+# Path to the directory containing this script, resolving any symlinks, etc
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+# Name of the current directory
+SQL_DIR_NAME=$(printf '%s\n' "${SCRIPT_DIR##*/}")
```
AlD: No need for the subshell and printf.

I'd remove this and use `${SCRIPT_DIR##*/}` directly below.

src/core/SQL/updateSQLResource.sh
```diff
@@ -0,0 +1,45 @@
...
+EOF
+
+# In a subshell, change to the SQL directory so find output remains consistent,
+# then...
+# Find all files with a .sql ending...
+# | ...sort the results by version
+# | ...modify the beginning to change '.[...]' into '    <file>./SQL[...]' (add the directory path back)
+#   ...and modify the end to add '[...]</file>'
+# | ...append to the resource file.
+(cd "$SCRIPT_DIR" ; find . -name "*.sql" | sort --version-sort | sed -e "s/^./    <file>.\/$SQL_DIR_NAME/g" -e "s/$/<\/file>/g" >> "$SQL_RES_FILE" )
```
AlD: `--version-sort` is a fairly new feature (e. g. missing in macOS Sierra's default sort 5.93). 

`sort -t/ -k2,2 -k3n` is a bit convoluted, but does the trick here.

AlD: Just use `${SCRIPT_DIR##*/}` instead of `$SQL_DIR_NAME` here.

src/core/SQL/updateSQLResource.sh
digitalcircuit: Agreed; I misunderstood the [source answer on StackOverflow](https://stackoverflow.com/questions/1371261/get-current-directory-name-without-full-path-in-bash-script/1371283#1371283 ).

However, given this is not performance-critical and ```${SCRIPT_DIR##*/}``` isn't immediately obvious as to its function (*well, to me at least*), would it be fine to remove the subshell, but keep the variable?

src/core/SQL/updateSQLResource.sh
digitalcircuit: Makes sense!  I'll add some comments to explain this.

digitalcircuit:
I think I've addressed your requested changes asides from concern over removing `SQL_DIR_NAME` as [per my comment above](https://github.com/quassel/quassel/pull/268#discussion_r95009890 ).

*No hurry, I was merely a bit slow to realize Github doesn't have a clear way of saying "I addressed this review", hence a comment.*

justjanne:
Could this script be integrated into the cmake workflow, and the file that’s output added to .gitignore?

CMake seems to have a [function](https://cmake.org/cmake/help/v2.8.10/cmake.html#command%3aadd_custom_target) for that, according to this [stackoverflow post](stackoverflow.com/questions/15972898/cmake-how-to-run-a-add-custom-command-before-everything-else#answer-15973676)

PR #63, opened 2014-03-16
Added possibility of displaying timestamp of starting being away state.

bzyx:
It is a nice feature for my friends to see in a simple way at what time I've last time been seen. This feature provides possibility of adding customized time-stamp of last seen time.

Example message: "All Quassel clients vanished from the face of the earth... %%hh:mm:ss%%" which formats to: "All Quassel clients vanished from the face of the earth... 23:20:34"

PR #270, opened 2017-01-08
Allow custom timestamp in away messages

digitalcircuit:
## In short
* Treat `%%` in away messages as timestamp marker
  * Applies to away-on-detach, default away message, and manual `/away Message`
  * Handles multiple entries, e.g. `Away since %%hh:mm%% on %%dd.MM%%`
  * `%%%%` escapes to a literal `%%`
  * Invalid entries are mostly ignored according to [`QDateTime`](https://doc.qt.io/qt-4.8/qdatetime.html#toString )
* Don't re-process existing away message on startup
  * Preserves any consecutive escapes, e.g. a stored `%%%%` remains as `%%%%`
* Explain formatting in tooltips for away settings
  * Use a table for pretty listing, only mention `t` (*timezone*) [for Qt 5](https://doc.qt.io/qt-5/qdatetime.html#toString )
  * Add feature flag `AwayFormatTimestamp = 0x0200`, only show tooltip when present

Criteria | Rank | Reason
Impact | ★★☆ *2/3* | User-facing feature, easier to see if someone's gone for long
Risk | ★★☆ *2/3* | Changes away handling, `%%` needs escaped as `%%%%`
Intrusiveness | ★★☆ *2/3* | Minor changes, new feature flag might conflict with other PRs

This expands upon @bzyx's [pull request #63](https://github.com/quassel/quassel/pull/63 ) (thank you!), addressing some of the concerns from review.

## Rationale
Quassel allows setting an away message whenever all clients detach, a handy way to see if someone's able to get messages.  However, this doesn't indicate how recently they disconnected.  If on Quasseldroid with a poor signal, the connection might drop every few minutes, whereas if one does not use IRC for a while, it could be months.

By allowing custom timestamp formatting in away messages, Quassel simplifies indicating how long ago one actually used their client, or for any other reason in a custom away message.  This remains optional, as always.

This changes handling of `/away` in a non-compatibile manner as `%%` is now treated specially.  `%%%%` translates to a literal `%%`.  However, this shouldn't affect most situations, and `%%` can be changed to any other identifier if needed.

### Custom away message
*Complex example with multiple time/date format entries and repeated escape codes*
Hovering over IRC nickname in Nick list showing the away message tooltip

**Specified away message**
```Away since %%h:mm ap%% (%%t%%) on %%yyyy-MM-dd%% - %%%% %%%%%%%%be back later%%%%%%%% %%%%```

**Converted away message**
```Away since 4:36 pm (CST) on 2017-01-08 - %% %%%%be back later%%%% %%```

### Settings → Identity tooltips
*New core, new client, Qt 5 (so timezone code is supported)*
Tooltip when hovering over "Away on Detach" in Settings - IRC - Identities - Away tab.

**Tooltip text**
> \[Original tooltip message\].
>
> You can add date/time to this message using the syntax: 
> %%*<format>*%%, where *<format>* is:
> 
> n/a| \[headers invisible, just for Markdown\]
> ---|-----------------
> hh | the hour
> mm | the minutes
> ss | seconds
> AP | AM/PM
> dd | day
> MM | month
> t  | current timezone
> 
> Example: Away since %%hh:mm%% on %%dd.MM%%.
> %%%% without anything inside represents %%.  Other format codes are available.

### DIFF
--- a/src/common/quassel.h
+++ b/src/common/quassel.h
@@ -75,8 +75,9 @@ public:
         VerifyServerSSL = 0x0040,          /// IRC server SSL validation
         CustomRateLimits = 0x0080,         /// IRC server custom message rate limits
         DccFileTransfer = 0x0100,
+        AwayFormatTimestamp = 0x0200,      /// Timestamp formatting in away (e.g. %%hh:mm%%)
 
-        NumFeatures = 0x0100
+        NumFeatures = 0x0200
     };
     Q_DECLARE_FLAGS(Features, Feature)
 
--- a/src/common/util.cpp
+++ b/src/common/util.cpp
@@ -201,17 +201,77 @@ QByteArray prettyDigest(const QByteArray &digest)
     return prettyDigest;
 }
 
-QString formatCurrentDateTimeInString(QString str)
-{
-    /*
-     * Find %%<text>%% in string. Repleace inside text which is format to QDateTime
-     * with current timestamp.
-     */
-    QRegExp rx("\\%%(.*)\\%%");
-    rx.setMinimal(true);
-    int s = rx.indexIn(str);
-    if (s >= 0)
-        str.replace(s, rx.cap(0).length(), QDateTime::currentDateTime().toString(rx.cap(1)));
 
-    return str;
+QString formatCurrentDateTimeInString(const QString &formatStr)
+{
+    // Work on a copy of the string to avoid modifying the input string
+    QString formattedStr = QString(formatStr);
+
+    // Exit early if there's nothing to format
+    if (formattedStr.isEmpty())
+        return formattedStr;
+
+    // Find %%<text>%% in string. Replace inside text formatted to QDateTime with the current
+    // timestamp, using %%%% as an escape for multiple %% signs.
+    // For example:
+    // Simple:   "All Quassel clients vanished from the face of the earth... %%hh:mm:ss%%"
+    // > Result:  "All Quassel clients vanished from the face of the earth... 23:20:34"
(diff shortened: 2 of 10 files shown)


digitalcircuit:
Someone else pointed out that `strftime` (or in this case, [`QDateTime::toString`](https://doc.qt.io/qt-4.8/qdatetime.html#toString ) ) already has a way to escape literal text - use single quotes.  I hadn't realized that when embarking on this.  So...

### Alternative Option 1
Add option to enable/disable timestamp parsing, require away messages to match `QDateTime`'s expectations.

* Benefits
  * Simpler escaping - less need for `%%` everywhere
  * Less risk of performance and implementation issues; Qt is well-tested, don't re-implement the wheel
* Drawbacks
  * Yet another option with database schema change, client UI/protocol modification, etc
  * Complications over scope - is timestamp format active for this identity/network/account, or not?  How to explain to the user?

### Alternative Option 2
Convert away messages into a format that matches `QDateTime`'s expectations, using custom logic to invert the default behavior.

* Benefits
  * Simpler escaping depending on how it's implemented
  * Possibly better performance by relying on `QDateTime`'s internals
* Drawbacks
  * Convoluted conversion from one type of escape code to another
  * No option to add custom format codes in the future, e.g. `%%uptime%%`

---
All core-side approaches suffer from a common issue of using the Quassel client in one timezone while the core resides in another timezone.  To avoid this problem, either the client needs to transmit timezone information to the core (*which client wins?*), the core needs to allow configuring the timezone, the core needs to use UTC, or something else.

Discussion is welcome!

PR #273, opened 2017-03-03
Improve performance of PostgreSQL database in instances where buffers exist with no recent activity

ReimuHakurei:
This patch improves the performance of PostgreSQL in cases where there exists no recent activity.

For example, if there are 80 million rows in the backlog table, before, if your initial backlog was 50, Quassel would seek backwards from row 80 million until it found 50 messages, even if the most recent message was at say... row 60 million.

This patch makes it so Quassel jumps straight to row 60 million in that case, and then scans backwards for 50 messages.

In my case, this improved the time it takes Quassel to synchronize from over 5 minutes down to 25 seconds.


AlD:
Great initiative, thanks!

Can you please clarify which backlog fetch method (fixed or unread) this targets?

From skimming through the diff I assume it's the fixed fetch method. It would be useful to improve the unread fetcher using the same field, by only selecting buffers where lastseenmsgid < lastmsgid. Would it be feasible for you to add that as well?

AlD:
Please also make the commit message a bit more verbose, e.g. mention the database schema change, briefly outline the approach.

ReimuHakurei:
Is this commit message a bit closer to what you wanted?

Yes, it is changing the fixed backlog method. The unread backlog method has its' own similar thing right now using lastseenmsgid, so this change would not help that.

ReimuHakurei:
On further testing, looks like this WOULD actually help the unread backlog method, so I have duplicated this improvement on that as well.

ReimuHakurei:
Hold off on merging this for the moment -- some quick testing shows that unread backlog method is still taking much longer. Fixed took me 15 seconds to connect, and unread took 3 minutes. I imagine adding some indexes could speed up unread, so it'd be good to add those with this schema update to avoid another schema update right afterwards. 

I'll look into this more tonight.

AlD:
Thanks again!

The commit message looks good now.

It's been a long time since I looked at this, but I think the unread fetcher would benefit from looking just at backlog for buffers with lastseenmsgid < lastmsgid.

Right now I think it tries to fetch messages for all buffers.

ReimuHakurei:
Alright, this should fix that issue with unread mode - on my core it is now nice and speedy, just like fixed mode. It should also slightly improve SQLite3 performance.

egs-me:
Nice work! Thanks a lot!

Short question about the sqlite implementation: I was confused about the D-Tour via the buffer_new. I guess that's to add the CHECK constraint?

Wouldn't it suffice to add the check along with the newly added column? Or does that have a different effect that I'm unaware of? I.e., in upgrade_000_alter_buffer_add_lastmsgid.sql:

`ALTER TABLE buffer
ADD COLUMN lastmsgid INTEGER NOT NULL DEFAULT 0 CHECK (lastseenmsgid <= lastmsgid);`

Cheers,
Marcus

ReimuHakurei:
It shouldn't, in theory, cause problems, but from personal experience I can SQLite can be... weird about workarounds like that, so I did it the long way just to be safe. On my rather bloated massive test database with a few hundred buffers, it only takes a few seconds to run the long way anyway.

egs-me:
Great to hear that you've also tested it on a populated database. I was worried for a second that there might be a foreign key from backlog to buffer with an ON DELETE CASCADE option set. Luckily this is not the case.

Again: thank you for your work! I'm sure you made quite a couple of users happy.

Cheers,
Marcus

PR #274, opened 2017-03-04
Fix SQLite setup, PostgreSQL migrate, cleanup SQL

digitalcircuit:
## In short
* Fix SQLite buffer setup, fix PostgreSQL migration
  * Replace `ALTER TABLE` command in setup with `CHECK` constraint as seen in the upgrade script
  * Add handling for `lastmsgid` when migrating from SQLite to PostgreSQL
  * Fix PostgreSQL migration `SELECT populate_lastmsgid()` call by first resetting query
  * Fixes regressions from [pull request 273](https://github.com/quassel/quassel/pull/273 )
* Add `lastmsgid` limit on fetching backlog to SQLite
  * Update `select_messagesNewerThan` and `select_messagesNewestK` to use `lastmsgid`
  * Should provide additional performance improvements for SQLite
  * May help with [pull request 247](https://github.com/quassel/quassel/pull/247 )
* Clean up SQL files
  * `select_messages` → `select_messagesNewestK` (PostgreSQL) or `select_messagesRange` (SQLite)
  * Maintain tabs in `.sql` files, remove extra newline
  * Sort `sql.qrc` alphabetically
  * Add documentation to the `setLastMsg` function

Criteria | Rank | Reason
Impact | ★★★ *3/3* | Fixes broken default setup, Postgres migrate, performance
Risk | ★★☆ *2/3* | Might break odd or older setups, possible performance regressions
Intrusiveness | ★★☆ *2/3* | Schema changes, might interfere with other pull requests

*Many thanks to @ReimuHakurei for finding these performance improvements in the first place!*

### Before
*Submitted by @justJanne in Freenode/`#quassel`*
```
2017-03-04 20:57:11 Warning: SslServer: Certificate file /home/heinz/.config/quassel-irc.org/quasselCert.pem does not exist
2017-03-04 20:57:11 Warning: SslServer: Unable to set certificate file
           Quassel Core will still work, but cannot provide SSL for client connections.
           Please see http://quassel-irc.org/faq/cert to learn how to enable SSL support.
2017-03-04 20:57:11 Warning: SslServer: Certificate file /home/heinz/.config/quassel-irc.org/quasselCert.pem does not exist
("QIBASE", "QSQLITE", "QSQLITE3", "QMYSQL", "QMYSQL3", "QODBC", "QODBC3", "QPSQL", "QPSQL7", "QTDS", "QTDS7")
Core is currently not configured! Please connect with a Quassel Client for basic setup.
Storage Schema is missing!
unhandled Error in QSqlQuery!
                  last Query:
 CREATE TABLE buffer (
        bufferid INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
        userid INTEGER NOT NULL,
        groupid INTEGER,
        networkid INTEGER NOT NULL,
        buffername TEXT NOT NULL,
        buffercname TEXT NOT NULL, -- CANONICAL BUFFER NAME (lowercase version)
        buffertype INTEGER NOT NULL DEFAULT 0,
        lastmsgid INTEGER NOT NULL DEFAULT 0,
        lastseenmsgid INTEGER NOT NULL DEFAULT 0,
        markerlinemsgid INTEGER NOT NULL DEFAULT 0,
        key TEXT,
        joined INTEGER NOT NULL DEFAULT 0, -- BOOL
        ALTER TABLE buffer_new RENAME TO buffer;
)
              executed Query:
 
                bound Values:
                Error Number: 1
               Error Message: near "ALTER": syntax error Der Befehl konnte nicht ausgeführt werden
              Driver Message: Der Befehl konnte nicht ausgeführt werden
                  DB Message: near "ALTER": syntax error
Unable to setup Logging Backend!
FATAL: Selected storage backend is not available: "SQLite"
```

### After
Setup and migration works.

# Test results - ✅ pass
### Fresh configuration - ✅ pass
*Quassel core from this PR set up without any initial data*

Test conditions  | Result  | Remarks
-----------------|---------|-------------
SQLite, new core | ✅ *pass* |
SQLite, new monolithic | ✅ *pass* |
Postgres, new core | ✅ *pass* |

### Migration - ✅ pass
*Quassel core from this PR set up without any initial data, then migrated*

Test conditions  | Result  | Remarks
-----------------|---------|-------------
SQLite → Postgres, new core | ✅ *pass* |

### Upgrading existing - ✅ pass
*Quassel core 0.12.2 set up (0.13-git-master for monolithic), then upgraded to this PR*

Test conditions  | Result  | Remarks
-----------------|---------|-------------
SQLite, old → new core | ✅ *pass* |
SQLite, old monolithic → new monolithic | ✅ *pass* |
Postgres, old → new core | ✅ *pass* |

### Fresh configuration - ✅ pass
*Quassel core from this PR set up without any initial data*

Test conditions  | Result  | Remarks
-----------------|---------|-------------
SQLite, new core | ✅ *pass* |
SQLite, new monolithic | ✅ *pass* |

### Migration - ❓ N/A
*Without Postgres on Windows, migration could not be tested*

### Upgrading existing - ✅ pass
*Quassel core 0.12.4 set up, then upgraded to this PR*

Test conditions  | Result  | Remarks
-----------------|---------|-------------
SQLite, old → new core | ✅ *pass* | Didn't auto-reconnect networks on first launch
SQLite, old monolithic → new monolithic | ✅ *pass* |


avih:
If I understand correctly, this whole change is to address a single issue:
- The last seen message id (per buffer) can be bigger than the last message id (for the same buffer) at the db.

This, if I understand correctly, is a suspect cause for slowness while fetching all unread messages, possibly mostly with sqlite backend.

The fix here, again, if I understand correctly,  is to update the core/schema/migration such that the core can reject incorrect values of last seen id sent by the client.

However, what's missing, as far as I can tell, is a client-side fix which make it not send incorrect values. Such fix would probably have been enough on its own, assuming a user updates all her clients, as it would overwrite bad values at the db, and will make the db correct itself quickly.

Of course, also fixing the core to be more robust against incorrect values sent by clients is desirable, however, it should, IMHO, be secondary to the actual fix at the client side to not send bad values in the first place.

digitalcircuit:
This pull request focuses on fixing the currently-broken `master` branch.

I agree, this should be fixed client-side as well *in a follow-up pull request*.  Unfortunately, I encountered some trouble figuring out Qt and don't have the time right now to fix it.

Following the discussion on the `#quassel` channel, it shouldn't be too involved for someone who knows how the Qt UI is handled:
```cpp
MsgId ChatView::lastMsgId() const
    if (!scene())
        return MsgId();

    QAbstractItemModel *model = scene()->model();
    if (!model || model->rowCount() == 0)
        return MsgId();
    // Do something here to roll backwards from rowCount() until
    // a non-DayChanged message is found!

    return model->index(model->rowCount() - 1, 0).data(MessageModel::MsgIdRole).value<MsgId>();

MsgId ChatView::lastVisibleMsgId() const
{
    // Call lastVisibleChatLine(true); to ignore DayChanged messages!
    ChatLine *line = lastVisibleChatLine();

    if (line)
        return line->msgId();

    return MsgId();
}
```
If someone else wants to tackle this, feel free to!  Otherwise, it's on my `TODO` list.

avih:
> Of course, also improving the core to be more robust against incorrect values sent by clients is desirable, however, it should, IMHO, be secondary to the actual fix at the client side to not send bad values in the first place.

Note also that users can typically more easily update clients than cores. core could be on some LTS distro release which won't update to a new version of core, thus leaving all those users in the cold. However, if a client update is enough to fix it, all those users would benefit too.

justjanne:
At the same time, while an admin can update a core, a single device on an old version could kill the performance again.

Both sides should be fixed ideally, although I agree that this belongs in a follow-up.

On 5 March 2017 06:36:59 GMT+01:00, avih <notifications@github.com> wrote:
>> Of course, also improving the core to be more robust against
>incorrect values sent by clients is desirable, however, it should,
>IMHO, be secondary to the actual fix at the client side to not send bad
>values in the first place.
>
>Note also that users can typically more easily update clients than
>cores. core could be on some LTS distro release which won't update to a
>new version of core, thus leaving all those users in the cold. However,
>if a client update is enough to fix it, all those users would benefit
>too.
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/quassel/quassel/pull/274#issuecomment-284207047

-- 
Diese Nachricht wurde von meinem Android-Mobiltelefon mit K-9 Mail gesendet.

avih:
> At the same time, while an admin can update a core

That depends. If an admin wants to stick to the distro's packages, which I think many do, especially in an LTS setup which prefers stability, then they might not want to update core from some external source.

E.g. I'm guessing Ubuntu LTS 16.04 will not see this core update. And some would use it for several years from now still.

> a single device on an old version could kill the performance again.

True. But there's still great value in the fact that some setups would benefit for it, and arguably these setups (use only new clients) are easier to set than updating the core.

Again, making the core more robust is a great thing which I have nothing against. It just should not be a replacement for fixing the root cause.

avih:
```cpp
MsgId ChatView::lastMsgId() const
{
    ...
    return model->index(model->rowCount() - 1, 0).data(MessageModel::MsgIdRole).value<MsgId>();
}
```

As far as I can tell, only `ChatView::lastMsgId()` is used to set a last-seen value on the buffer at the core/db. The `ChatView::lastVisibleMsgId()` function is only used to display the marker line - client side only, and as far as I can tell doesn't propagate back to the core. Though I could be reading the code incorrectly.

Note that the code itself states that the id of a day-change message should be that of the preceding message, I'm guessing recursively until a "real" message is found: https://github.com/quassel/quassel/blob/065ad418911de85ee1a94a178cd41ae32d0ac91e/src/client/messagemodel.cpp#L155

Therefore the `ChatView::lastMsgId()` should not be able to be bigger than that of the actual last message-id at the db for that buffer - on cases where there are several day-change messages, even if they're at the end of the buffer.

So I think something else might be at play here.

egs-me:
Ouch. I trusted that it was tested and didn't spot the error in the original pull request. My fault.
From what I can see, the changes look good and the testing is solid, so I'm going to merge this too.

Thanks and cheers,
Marcus

PR #275, opened 2017-03-05
Added ctrl+u to clear buffer.

weabot:
The client was unusable for me without the ctrl+u hotkey which clears the buffer from GNU Readline (I believe). So I changed that.

It's a very simple change that should make the line editing more intuitive to users used to GNU Readline.

### DIFF
--- a/src/qtui/bufferwidget.cpp
+++ b/src/qtui/bufferwidget.cpp
@@ -228,6 +228,13 @@ bool BufferWidget::eventFilter(QObject *watched, QEvent *event)
         return false;
 
     switch (keyEvent->key()) {
+	case Qt::Key_U:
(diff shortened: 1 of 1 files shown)


rikai:
Firstly, this is a wonderful update.

However, I think may cause some confusion for many users as CTRL+U is also often used for applying text formatting (usually underline, which Quassel has support for, though it is also currently missing the ability to be binded to a key) in input fields in many text input fields in GUIs. Running it by several other users, also almost unanimously revealed the same expectation from them.

So, to rectify this, my proposal for a change you could make would be to instead add a keybind for this functionality, allowing it to be user-assignable to whatever a user prefers with a default of no binding assigned.

Do you foresee any issues with such a solution or is that good for you as well? :)

PR #277, opened 2017-03-09
Fix lastseenmsgid updates in PostgreSQL

mamarley:
GH-273 introduced a workaround for the issue where sometimes the
lastseenmsgid for a buffer was set to a msgid not in that buffer.
However, the workaround was incorrect as the "min()" function in
PostgreSQL is an aggregate function for use on a column, not a
function to return the least of a list of values passed in.  The
correct function for use here is "least()".


digitalcircuit:
Looks good here!  Yet another thing I overlooked with two fix-up PRs by now, eesh :)

For easier reference during review, here's the [PostgreSQL docs for `least()`]( https://www.postgresql.org/docs/current/static/functions-conditional.html#FUNCTIONS-GREATEST-LEAST ).

digitalcircuit:
While this doesn't affect fetching unread messages until the core restarts, it seems to affect loading most recent messages regardless of restarting, e.g with Quasseldroid.

@egs-me , would you be willing to take a look?

egs-me:
Thanks for catching that one!

~ Marcus

PR #278, opened 2017-03-16
Store sender modes in the database

justjanne:
In short
--------

* When processing a new message, find the mode the user that sent this message has in a channel (for example `o` or `hv`)
* Convert it to a prefix (for example `@` or `+%`)
* Store this in a new `sendermode` column in the backlog table

Rationale
---------

There have been multiple feature requests made asking to support sender modes, and displaying them in the client. This will require some larger changes, specifically to the way objects are serialized and how client features are acknowledged to the core.

To prepare for this change, this PR stores the relevant data already, so once users upgrade to the future version which will display the data, a large part of their backlog will have the user modes already included.

Impact
-------

This affects the schema of the databases, for both PostgreSQL and SQLite. It affects migration, creation, and upgrades. It also increments the schema number.

A minor performance impact is introduced on the core, looking up the modes of the sender of every message, and transforming them into prefixes.

No impact on the client exists.


src/core/coresession.h
```diff
@@ -222,6 +222,8 @@ private slots:
     QList<RawMessage> _messageQueue;
     bool _processMessages;
     CoreIgnoreListManager _ignoreListManager;
+
+    QString senderModeForMessage(const RawMessage &rawMsg, const BufferInfo &bufferInfo) const;
```
digitalcircuit: Minor: Add documentation for this new function?

*Sput's pointed to [`src/uisupport/aboutdata.h`](https://github.com/quassel/quassel/blob/master/src/uisupport/aboutdata.h) in the past for the example of what style to use now.*

src/common/message.cpp
```diff
@@ -67,6 +69,7 @@ QDataStream &operator>>(QDataStream &in, Message &msg)
     msg._bufferInfo = buf;
     msg._timestamp = QDateTime::fromTime_t(ts);
     msg._sender = QString::fromUtf8(s);
+    msg._senderMode = QString("");
```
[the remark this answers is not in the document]
justjanne: Correct, that’s the idea.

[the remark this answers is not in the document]
digitalcircuit: Alright!  Minor thing, I'd add an in-code comment to here and to `operator<<`, serving to explain to anyone else that stumbles through the code.

src/common/network.h
```diff
@@ -173,6 +173,7 @@ public :
     inline QString prefixToMode(const QCharRef &prefix) const { return prefixToMode(QString(prefix)); }
     QString modeToPrefix(const QString &mode) const;
     inline QString modeToPrefix(const QCharRef &mode) const { return modeToPrefix(QString(mode)); }
+    QString modesToPrefixes(const QString &mode) const;
```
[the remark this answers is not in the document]
justjanne: I’m not sure this is really that useful, as this method just maps 1:1 to modeToPrefix (it literally just maps over the characters, one by one, and then returns the result). I’ll do it anyway.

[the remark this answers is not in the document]
digitalcircuit: That's a fair point; it could either be postponed, or perhaps using [Doxygen method grouping](https://www.stack.nl/~dimitri/doxygen/manual/grouping.html ) to describe all of them at once.

Sput42:
> Sput's pointed to src/uisupport/aboutdata.h in the past for the example of what style to use now.

Indeed, although I think we can get rid of the [in] for input parameters again, it just adds noise and is rarely useful.

src/core/SQL/SQLite/version/24/upgrade_000_create_mode.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE backlog
```
Sput42: The file should probably be called upgrade_000_create_senderprefixes.sql now...

src/core/SQL/PostgreSQL/version/23/upgrade_000_create_mode.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE backlog
```
Sput42: The file should probably be called upgrade_000_create_senderprefixes.sql now...

src/core/coresession.cpp
```diff
@@ -421,6 +424,25 @@ void CoreSession::processMessages()
...
+
+    if (bufferInfo.type() != BufferInfo::ChannelBuffer) {
+        return "";
+    }
+
+    IrcChannel *currentChannel = currentNetwork->ircChannel(bufferInfo.bufferName());
+    if (!currentChannel) {
+        return "";
+    }
+
```
Sput42: It's theoretically more efficient to return a default-constructed value (i.e. `return {};`) instead of constructing an empty string explicitly. Although I would assume compiler optimization would get rid of the unnecessary type conversion, I find it also stylistically nicer.

src/common/message.cpp
```diff
@@ -67,6 +69,7 @@ QDataStream &operator>>(QDataStream &in, Message &msg)
     msg._bufferInfo = buf;
     msg._timestamp = QDateTime::fromTime_t(ts);
     msg._sender = QString::fromUtf8(s);
+    msg._senderPrefixes = QString("");
```
Sput42: Default-constructed string should be slightly more efficient and is more readable:
` msg._senderPrefixes = {};`

src/common/message.h
```diff
@@ -66,9 +66,9 @@ class Message
     };
     Q_DECLARE_FLAGS(Flags, Flag)
 
-    Message(const BufferInfo &bufferInfo = BufferInfo(), Type type = Plain, const QString &contents = "", const QString &sender = "", Flags flags = None);
+    Message(const BufferInfo &bufferInfo = BufferInfo(), Type type = Plain, const QString &contents = "", const QString &sender = "", const QString &senderPrefixes = "", Flags flags = None);
     Message(const QDateTime &ts, const BufferInfo &buffer = BufferInfo(), Type type = Plain,
-        const QString &contents = "", const QString &sender = "", Flags flags = None);
+        const QString &contents = "", const QString &sender = "", const QString &senderPrefixes = "", Flags flags = None);
```
Sput42: While your additions are consistent with the existing code, you could replace all occurrences of `= ""` by `= {}` if you like to modernize the signature.

src/common/network.h
```diff
@@ -172,10 +172,37 @@ public :
...
 
+    /**@{*/
+    /**
+     * Translates a user’s prefix to the channelmode associated with it.
+     * @param prefix Prefix to be translated.
+     */
     QString prefixToMode(const QString &prefix) const;
     inline QString prefixToMode(const QCharRef &prefix) const { return prefixToMode(QString(prefix)); }
+    inline QString prefixesToModes(const QString &prefix) const {
+        QString mode = "";
```
Sput42: No need to initialize (QString is empty by default)

src/common/network.h
```diff
@@ -172,10 +172,37 @@ public :
...
+        for (QChar c : prefix) {
+            mode += prefixToMode(c);
+        }
```
Sput42: If you'd like to modernize, you could use an algorithm instead of a manual for loop, something like
```
QString modes;
std::transform(prefix.cbegin(), prefix.cend(), std::back_inserter(modes), [](auto c) { return prefixToMode(c); });
```
You'd need to `#include <algorithm>` and `#include <iterator>` for that.

Also it would be worth considering if we can assume that both modes and prefixes are always single characters, and then optimize the related single-mode methods (`prefixToMode` and friends) to use QChar instead of QString. Since these are frequently called methods, avoiding QString heap allocations could have quite some positive impact.
Obviously this is not directly related to this PR, but useful to keep in mind for potential follow-up commits... unless you decide to make it part of this PR since you're touching the mode stuff anyway :)
If we can assume that both modes and prefixes are single characters, you could also pre-size the `modes` string since then it must be of the same length as `prefix`.

src/common/network.h
```diff
@@ -172,10 +172,37 @@ public :
...
+     * @param prefix Prefix to be translated.
+     */
     QString modeToPrefix(const QString &mode) const;
     inline QString modeToPrefix(const QCharRef &mode) const { return modeToPrefix(QString(mode)); }
+    inline QString modesToPrefixes(const QString &mode) const {
+        QString prefix = "";
+        for (QChar c : mode) {
+            prefix += modeToPrefix(c);
+        }
+        return prefix;
```
Sput42: Same comments that I gave for `prefixToModes()` apply here.

src/common/message.h
Sput42: Something that I would tackle once we start optimizing the codebase for Modern C++, but that could also be already started in existing code we touch now: call-by-value with proper use of std::move() is likely more efficient than using const refs in cases where the receiving object stores the input parameter (in case an rvalue is given, which is e.g. the case for `senderPrefixes` which is called with the result of a method, one would need only one copy to store the value in the msg instance, whereas the const ref requires two copies).

Such things seem to be minor (and should be backed by actual benchmarks rather than theoretical considerations), but since we're handling millions of messages this seems like a likely hotspot for optimization.

src/common/network.h
justjanne: The problem we have with that is that when looking up a prefix, we might get either a single character back, or no character.

So a lookup never returns > 1 character, but may return 0 characters. I’m not sure we can accurately express and use that with QChar.

PR #279, opened 2017-03-16
Fix .gitignore accidentally ignoring tracked (and important) files

justjanne:
In short
--------

* Before, all folders with the name `build`, no matter in what subfolder, were ignored by git.
* After, only the `build` folder in the root of the repository will be ignored.

Rationale
---------

The project has several files under `scripts/build/`, which are important to create and generate the Windows installer, Mac OS X package, and configure Snore.

Those files are matched by the ignore rule `build*`, and therefore any change to them is untracked.

This pull requests fixes aforementioned issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-build*
+/build*
 *.pyc
 tags
 


digitalcircuit:
From looking through https://git-scm.com/docs/gitignore , this seems correct!  I hadn't noticed this before, nice catch.

Nevcairiel:
For the record, git doesn't ignore changes to existing tracked files, even if they match an ignore rule- only new files would be affected by this and can still be manually added, overriding the ignore rule.

But harmless change either way.

justjanne:
@Nevcairiel It actually does – git will refuse to add changes to existing files if they match the .gitignore. I’ve had this issue myself before, (or it might just be because of my git config).

PR #284, opened 2017-03-25
Add 'Configure' to the right-click Menu of a Network in the buffer list

genius3000:
This adds a menu item of 'Configure' to the right-click menu for a Network in the buffer list.
Clicking on 'Configure' will open the Configure Networks dialog to the respective Network's settings.

**Use-Case:** 
* Large network lists, when you need to make a change to a network...it's now just a click away instead of going through the Main Menu or the Main Settings dialog.
* Continuous modification of Networks, ie: IRC Development, TestNets, etc.

**Impact:**
* 'Configure' is added above 'Disconnect/Connect', this could cause some users to re-learn habits if they are very used to using the 'Disconnect/Connect' items in this menu.

**Testing:**
Tested this with:
* A single, connected Network (as seen below)
* A single, disconnected Network
* Multiple Networks, both connected and disconnected

Every method provides the expected results and no crashes.

quassel-configure

_This is my first time working with Qt code, my apologies if this is done horribly wrong._ Any comments, suggestions, etc are welcome!

### DIFF
--- a/src/qtui/mainwin.cpp
+++ b/src/qtui/mainwin.cpp
@@ -204,6 +204,7 @@ void MainWin::init()
         SLOT(messagesInserted(const QModelIndex &, int, int)));
     connect(GraphicalUi::contextMenuActionProvider(), SIGNAL(showChannelList(NetworkId)), SLOT(showChannelList(NetworkId)));
     connect(Client::instance(), SIGNAL(showChannelList(NetworkId)), SLOT(showChannelList(NetworkId)));
+    connect(GraphicalUi::contextMenuActionProvider(), SIGNAL(showNetworkConfig(NetworkId)), SLOT(showNetworkConfig(NetworkId)));
     connect(GraphicalUi::contextMenuActionProvider(), SIGNAL(showIgnoreList(QString)), SLOT(showIgnoreList(QString)));
     connect(Client::instance(), SIGNAL(showIgnoreList(QString)), SLOT(showIgnoreList(QString)));
 
@@ -1390,6 +1391,15 @@ void MainWin::showChannelList(NetworkId netId)
     channelListDlg->setAttribute(Qt::WA_DeleteOnClose);
     channelListDlg->setNetwork(netId);
     channelListDlg->show();
+}
+
+
+void MainWin::showNetworkConfig(NetworkId netId)
+{
+    SettingsPageDlg dlg(new NetworksSettingsPage(this), this);
+    if (netId.isValid())
+        qobject_cast<NetworksSettingsPage *>(dlg.currentPage())->bufferList_Open(netId);
+    dlg.exec();
 }
 
 
--- a/src/qtui/mainwin.h
+++ b/src/qtui/mainwin.h
@@ -119,6 +119,7 @@ private slots:
     void messagesInserted(const QModelIndex &parent, int start, int end);
     void showAboutDlg();
     void showChannelList(NetworkId netId = NetworkId());
+    void showNetworkConfig(NetworkId netId = NetworkId());
     void showCoreConnectionDlg();
     void showCoreConfigWizard(const QVariantList &, const QVariantList &);
     void showCoreInfoDlg();
--- a/src/qtui/settingspages/networkssettingspage.cpp
+++ b/src/qtui/settingspages/networkssettingspage.cpp
@@ -567,6 +567,14 @@ QListWidgetItem *NetworksSettingsPage::insertNetwork(const NetworkInfo &info)
 }
 
 
+// Called when selecting 'Configure' from the buffer list
+void NetworksSettingsPage::bufferList_Open(NetworkId netId)
+{
+    QListWidgetItem *item = networkItem(netId);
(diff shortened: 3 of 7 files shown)


src/qtui/mainwin.cpp
```diff
@@ -1390,6 +1391,15 @@ void MainWin::showChannelList(NetworkId netId)
 }
 
 
+void MainWin::showNetworkConfig(NetworkId netId)
+{
+    SettingsPageDlg dlg(new NetworksSettingsPage(this), this);
+    if (netId.isValid())
+        qobject_cast<NetworksSettingsPage *>(dlg.currentPage())->bufferList_Open(netId);
+    dlg.exec();
+}
```
digitalcircuit: You might be missing some close brackets for your `if` statement.  Alternatively, you can just return from the function early, e.g.
```cpp
void MainWin::showNetworkConfig(NetworkId netId)
{
    if (!netId.isValid())
        return;
    SettingsPageDlg dlg(new NetworksSettingsPage(this), this);
    qobject_cast<NetworksSettingsPage *>(dlg.currentPage())->bufferList_Open(netId);
    dlg.exec();
}
```
I don't know if it's a good idea to have some sort of fallback message (*e.g. a warning message box*) if this check fails.

*Note: not tested locally.*

genius3000: I'll run a few test cases when I get time again (possibly not till next week) but I want to say that right now, with a bad netId, it will just open the Network Settings page as per usual (top network selected). Your thought could be a better approach, maybe even with a message of the (unusual) error, but still opening the Networks Settings page.

digitalcircuit: Noted!

I'm not sure if it's common enough to warrant more complication with messages/strings, but that would be nice for anyone that might get stuck in this situation.  Maybe one of those more knowledgeable with Quassel can remark on the likelihood.

If nothing else I'd add a comment explaining the intent so people like me don't get confused :)

(*Quassel doesn't handle invalid/missing other objects such as Identities very well at the moment, sometimes crashing, so props for considering invalid IDs here!*)

genius3000: I did some testing and this seems safe as is. I was mistaking (at least recently) the `.isValid()` check as a check for a valid Network ID as seen in the DB. That's incorrect, it only checks for being greater than 0.
The setting of the active Network in NetworksSettingsPage::bufferList_Open will fail gracefully with an invalid or missing netId resulting in no Network being selected and no crash scenario (afaik). 

#### Testing (as is)
##### netId = -1
Dialog loads, selected Network is first one in the list.

##### netId = 0
Dialog loads, selected Network is first one in the list.

##### netId = 754 (does not exist)
Dialog loads, no Network is selected but first one is shown but untouchable until a Network is selected in the list. This should be the most probable case (still rare though) and I like the outcome. If you selected a Network in the client that somehow doesn't exist in the DB, you can't change anything before selecting a Network in the list, which also shouldn't show a "missing" Network, but I'm unsure on that possibility.

##### netId unmodified (using the requested Id)
Dialog loads, selected Network is the requested one and is touchable.

genius3000:
Rebased this to latest master, signed the commit and updated the message. Also re-tested on an up-to-date master core without issue.

genius3000:
@digitalcircuit thanks for the testing/review. In a couple days I'll have some decent free time and will take a look into your findings; would be nice to keep it clean :)  
Just to make sure I understand correctly,  
The first scenario is within the network list,  with chosen network being down out of view without scrolling?  
The second is network list again, just with scroll bars unnecessarily shown?

digitalcircuit:
@genius3000 I've updated my original review with a screenshot.  You're correct on the first one, sort-of for the second.  I don't know if trying to fit the network list is a good idea, just the dialog itself unnecessarily having a scrollbar is weird.

genius3000:
@digitalcircuit 
1) I couldn't get `scrollToItem()` to work within `bufferList_Open()` either (tried many variations). It does work within, (for ex.) `widgetHasChanged()` but that's not ideal for this.  
As is, it does do a form of (auto)Scroll to all but the very last item in the list. While not ideal, I can agree is just a minor issue.
2) I could only replicate the scrollbars on the entire window by changing `Client style` to a few that had larger font/more spacing (Windows and GTK+ for example). The scrollbars appeared when opening via this new method or `File->Networks->Configure Networks...`, so it appears to be a separate issue to look into with the way it's opened separately of the larger Settings Page.  

For now, I've resolved the merge conflict/rebased on current master and updated the `setCurrentItem()` call with an explicit [`QItemSelectionModel::SelectionFlag`](https://doc.qt.io/qt-5/qitemselectionmodel.html#SelectionFlag-enum).

digitalcircuit:
@genius3000 That's unfortunate… :/  I'm not sure on these either, maybe this can be revisited/someone else will find out?

@Sput42 , may we include this before the 0.13 beta as it introduces new strings?  There's no functionality issues, just UI wonkiness that could be worked out during beta and/or this change backed out if needed.

PR #286, opened 2017-04-06
Changed Core to allow redirecting [Whois] messages

justjanne:
In short
--------
* When processing new `[Whois]` messages, mark them as `Message::Notice` instead of `Message::Server`
* Set the sender of these messages to be `Whois`, and remove the `[Whois]` prefix (as otherwise they’d be displayed as `[Whois] [Whois] is nick!ident@host`)

Rationale
---------

There were several requests in the past to allow redirecting `[Whois]` responses to the currently open chat. This method allows redirecting them as if they were server notices.

Impact
------

This implementation affects the way messages are stored in the database, and therefore only works with updated core.

The schema is not changed.

The client does not have to be updated.

### DIFF
--- a/src/core/eventstringifier.cpp
+++ b/src/core/eventstringifier.cpp
@@ -235,7 +235,7 @@ void EventStringifier::processIrcEventNumeric(IrcEventNumeric *e)
     default:
         if (_whois) {
             // many nets define their own WHOIS fields. we fetch those not in need of special attention here:
-            displayMsg(e, Message::Server, tr("[Whois] ") + e->params().join(" "), e->prefix());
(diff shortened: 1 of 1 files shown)


CounterPillow:
That seems like a really hacky solution. Whois replies aren't NOTICE messages, and they don't come from a sender called "Whois". If the goal is to change which buffer the Whois replies are displayed in, changes this confusing should in my opinion not be made.

justjanne:
@CounterPillow: Then find a better solution that doesn't break the protocol and works without major changes.

The only real alternative is checking for messages without sender and type Message::Server starting with [Whois], or setting Message:: Redirected on the core.

Both are equally ugly.

PR #287, opened 2017-04-14
Pick first IRC server specified when not failed

genius3000:
With this change, any connect to an IRC Server on initial core start, first connect to new network, or a manual connect after a manual disconnect will use the first server in the Network's server list. This seems to be the intended behavior as there is an ordered list of servers.
Currently Quassel will keep connecting to the last used and working server regardless of position in the list. It would only change servers if the connection fails. There is also the 'useRandomServer' flag, which I believe might be disabled in the SettingsPages right now (or I'm totally missing it). Either way, that behavior is untouched and will still use a random server if enabled.

I also moved the clearing of the '_previousConnectionAttemptFailed' flag to only when it was 'true' and we're trying the next server in the list.

**Use-Case:**
- Please refer to the two issues referenced below as they explain the use cases.

**Impact:**
- Minor to no negative impact on users as this changes nothing user-facing.
- A minor positive impact should be noted as the ordered server list will be used as intended.

**Testing:**
- Please see below comment from @digitalcircuit as he performed a set of testing. I failed to include my testing originally in this PR.

Reference: http://bugs.quassel-irc.org/issues/693
And: http://bugs.quassel-irc.org/issues/1006
Fixes #693 and #1006

### DIFF
--- a/src/core/corenetwork.cpp
+++ b/src/core/corenetwork.cpp
@@ -214,12 +214,16 @@ void CoreNetwork::connectToIrc(bool reconnecting)
     }
     else if (_previousConnectionAttemptFailed) {
         // cycle to next server if previous connection attempt failed
+        _previousConnectionAttemptFailed = false;
         displayMsg(Message::Server, BufferInfo::StatusBuffer, "", tr("Connection failed. Cycling to next Server"));
         if (++_lastUsedServerIndex >= serverList().size()) {
             _lastUsedServerIndex = 0;
         }
     }
-    _previousConnectionAttemptFailed = false;
+    else {
+        // Start out with the top server in the list
+        _lastUsedServerIndex = 0;
+    }
 
     Server server = usedServer();
     displayStatusMsg(tr("Connecting to %1:%2...").arg(server.host).arg(server.port));


digitalcircuit:
## Functionality - 🗹 *pass*

Functionality looks good!  I've tested by setting up `127.0.0.1` and `localhost` as connections to [Bitlbee](https://www.bitlbee.org/ ), and whenever I immediately restart Bitlbee, Quassel moves to the next server, but manually disconnecting/reconnecting jumps to the first server.

I also tested with `master` and as expected, Quassel did *not* behave correctly there - disconnecting/reconnecting kept reusing the existing server.

## Style (subjective)

I'd suggest making your commit message itself more informative.  You'll also want to wrap lines at roughly 50 characters for the first line, the "subject", then wrap following lines at 70 characters, the "body".  This way it'll look nicer on Github and in `git shortlog`.

For example...
```
Pick first IRC server specified when not failed

When connecting to the IRC network, only clear the
'_previousConnectionAttemptFailed' flag when trying the next server
in the list.  Otherwise, reset server index to start from the top.

Currently, Quassel will continue to use the last used (working)
server (or a random server if set to). This doesn't match the
intention of having an ordered list of servers to use.  The change
here just makes Quassel start with the first server in the list
unless it's a failed attempt, in which it continues down the list
(same as currently).

This fixes the issue of getting "stuck" on a specific server even if
you manually disconnect and connect.

Reference: http://bugs.quassel-irc.org/issues/693
And: http://bugs.quassel-irc.org/issues/1006
Fixes #693 and #1006
```

The *`Fixes #....`* is [special syntax to tell Redmine](https://www.redmine.org/projects/redmine/wiki/RedmineSettings#Referencing-issues-in-commit-messages ) on http://bugs.quassel-irc.org/ that your commit fixes these issues.  If you just want to refer to an issue, use "refs" or (*in my opinion, better yet*) provide the full URL.

In the future, you might also want to demonstrate that your pull request works with a few examples.  I've done that this time :)

## Example
### Configuration
* `127.0.0.1` as first server
* `localhost` as second server

### Before, connect keeps reusing existing server
*Text within [brackets] are comments*
```
* Connecting to 127.0.0.1:6667...
* Requesting capability list...
[...]
[Restart Bitlbee]
* Error from server: Closing link: BitlBee server shutting down
* Connection failure: The remote host closed the connection
* Connection failed. Cycling to next Server
* Connecting to localhost:6667...
* Requesting capability list...
[...]
[Manually disconnect]
* Disconnecting. (Quiting room...)
[Manually connect]
* Connecting to localhost:6667...
[WRONG: it should've reset to 127.0.0.1 at the top of list]
* Requesting capability list...
[...]
```

### After, connect resets on manual dis/connect
*Text within [brackets] are comments*
```
* Connecting to 127.0.0.1:6667...
* Requesting capability list...
[localhost] BitlBee-IRCd initialized, please go on
* Ready to negotiate (found: userhost-in-names, multi-prefix, extended-join, away-notify, sasl)
[...]
[Restart Bitlbee]
* Error from server: Closing link: BitlBee server shutting down
* Connection failure: The remote host closed the connection
* Connection failed. Cycling to next Server
* Connecting to localhost:6667...
* Requesting capability list...
[...]
[Manually disconnect]
* Disconnecting. (Quiting room...)
[Manually connect]
* Connecting to 127.0.0.1:6667...
[CORRECT: it resets to 127.0.0.1 at the top of list]
* Requesting capability list...
[...]
```

genius3000:
@digitalcircuit Thank you for testing and showing your testing here, I apologize for not having that shown in my original message. I have updated the commit message with your example as it is precisely what is done here.
I also updated the original PR message with a better description and matched the styling to my previous PR for consistency.

digitalcircuit:
@genius3000 You didn't have to exactly copy it :)  Glad my comment helped, though, and no worries.

PR #293, opened 2017-05-01
Update main.cpp : High-DPI options fix

beanpole135:
Cleanup the High-DPI scaling options, and ensure they are set on the current QCoreApplication (needs to be after init - not a global setting before).

NOTE: Not compile-tested locally - but a simple enough change that it should work just fine.

### DIFF
--- a/src/common/main.cpp
+++ b/src/common/main.cpp
@@ -84,7 +84,14 @@ int main(int argc, char **argv)
 #if QT_VERSION < 0x050000 && defined Q_OS_MAC && !defined BUILD_CORE
     QApplication::setGraphicsSystem("raster");
 #endif
-
+//Setup the High-DPI settings
+# if QT_VERSION >= 0x050600 && defined(Q_OS_WIN)
+    QCoreApplication::setAttribute(Qt::AA_EnableHighDpiScaling); //Added in Qt 5.6
+#endif
+# if QT_VERSION >= 0x050400
+   //Added in the early Qt5 versions (5.0?)- use 5.4 as the cutoff since lots of high-DPI work was added then
+    QCoreApplication::setAttribute(Qt::AA_UseHighDpiPixmaps); 
+# endif
     // We need to explicitly initialize the required resources when linking statically
 #ifndef BUILD_QTUI
     Q_INIT_RESOURCE(sql);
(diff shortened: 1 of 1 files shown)


Sput42:
According to the documentation, at least AA_EnableHighDpiScaling must be set before constructing the application object. I'd assume it won't hurt to set the other one before, too... so I'm not sure if this does the right thing here...

beanpole135:
Sorry about that - I forgot about the restriction on the DPI scaling attribute. I have moved the DPI scaling rules into the other block of Qt version-specific ifdefs near the top of the file, and verified/commented with the Qt version restrictions for the individual attributes. I have not verified yet that your Qt version check method is accurate - but I think that is safe to assume based on all the other version checks in the file. :-)

FYI: The impetus for this is that I am running Quassel 0.12.4 right now on FreeBSD (Qt 5.7.1) and the high-DPI pixmaps option is definitely not enabled (boss just got me a 4K monitor). I have been going through and fixing all the Qt5 utilities for TrueOS to work with high-DPI and since I use quassel on a daily basis I figured I would send in a patch here too.

digitalcircuit:
@romibi and @TheOneRing , since you first added those Qt flags to Quassel for use on Windows, are you able to check if the Windows builds still function correctly with high-DPI monitors and this pull request?

(*Unfortunately I don't have a system to test this on, Windows or Linux, unless VirtualBox can emulate a high-DPI monitor*)

Sput42:
So the HIDPI options were explicitly disabled on Linux/OSX by @sandsmark in #254, since they apparently don't work correctly on anything but Windows. Did this change in later versions of Qt? Is it safe to enable this for all platforms again?

justjanne:
The options used here seem to work correctly on Linux in Qt 5.7 and later for me.

TheOneRing:
QtCreator still enables Qt::AA_EnableHighDpiScaling only on Windows http://code.qt.io/cgit/qt-creator/qt-creator.git/tree/src/app/main.cpp#n176

But Qt::AA_UseHighDpiPixmaps is always set http://code.qt.io/cgit/qt-creator/qt-creator.git/tree/src/app/main.cpp#n360

romibi:
Under newest Windows 10 (Which has HiDpi improvements) i cannot detect any scaling behaviour change with or without this PR …  

It is not perfect but no change.  
(SVG in buttons get scaled too much on non-scaled screens in multi screen setup. With and without this PR)

I don't know how it behaves on older Windows … it still might make sense to merge …

PR #296, opened 2017-06-22
Add 'twitch.tv/membership' capability for Twitch.tv

Orphis:
Twitch.tv implementation doesn't send NAMES list automatically
unless supported by the client as they can be quite big.

See: https://dev.twitch.tv/docs/v5/guides/irc/#twitch-irc-capability-membership

### DIFF
--- a/src/common/irccap.h
+++ b/src/common/irccap.h
@@ -107,6 +107,18 @@ namespace IrcCap {
     namespace Vendor {
 
         /**
+        * Twitch.tv membership message support
+        *
+        * User list in a channel can be quite large and often non required for bot users and is then optional.
+        *
+        * From Twitch.tv documentation:
+        * "Adds membership state event data. By default, we do not send this data to clients without this capability."
+        *
+        * https://dev.twitch.tv/docs/v5/guides/irc/#twitch-irc-capability-membership
+        */
+        const QString TWITCH_MEMBERSHIP = "twitch.tv/membership";
+
+        /**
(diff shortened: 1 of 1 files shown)


digitalcircuit:
Furthering the pull request review (*I'm not sure if I'm supposed to use that?  I don't have merge/manage powers*), this looks good.  I appreciate including the documentation so others in the future can see why Twitch is included in Quassel :)

Not tested, but if it works for you it should be fine!  If you can, try this with very large chats to make sure Quassel won't break.  Large user counts are a bottleneck for logging in and might be a cause for concern.  E.g. https://github.com/magne4000/quassel-webserver/issues/238 and https://github.com/sandsmark/QuasselDroid/issues/60#issuecomment-6904519 .

Eventually this'll need fixed properly via a protocol change, of course.  Large IRC channels should be handled by an IRC client, and it's on the Big Three list of things to make Quasseldroid (and Quassel) quick and lightweight.

In short, this seems fine, just might want to see who all will be affected.

Orphis:
So, I tested it in large-ish (600 viewers) channel and it didn't cause any issue. In huge channels (larger than 1000), as the API says, you only get the channel administrators listed, so they are even easier to handle :)

And well, when you have like 50k viewers, you don't really care about any specific individual anyway.

digitalcircuit:
In that case, I'd agree with merging this change.  No problems for you, and as you've stated the API won't let it exceed 1,000 users anyways - large IRC channels already go above that.

PR #297, opened 2017-06-30
Quassel GUI in snap

kalikiana:
Add the Quassel Qt5 GUI to the snap.
Move files into a **snap** folder.


snap/snapcraft.yaml
```diff
@@ -0,0 +1,45 @@
...
+grade: stable
+confinement: strict
+
+icon: icons/hicolor/128x128/apps/quassel.png
+apps:
+    core:
+        command: wrap quasselcore --syslog
+        plugs: [network-bind]
+        daemon: simple
+    quassel-kalikiana:
```
Sput42: Is this correct? I.e. should your nickname be referenced here, or did you mean to have just "quassel" here? I know nothing about snaps :)

snap/snapcraft.yaml
```diff
@@ -0,0 +1,45 @@
+name: quassel
+version: "0.13-pre"
```
Sput42: Do you know of any way to extract this value from CMAKE or another file? I don't like having the version specified in multiple places...

PR #298, opened 2017-07-24
Prevent DNS leaks on connection when using proxy

crtxcr:
After using tcpdump on my server I realized that I was seeing DNS queries that only quassel should be making. However, for those a proxy (tor) has been configured, so I shouldn't have seen them. This implied a DNS leak in quassel. 

96fe7e713 introduced a call to QHostInfo::fromName() in order to avoid problems related to the caching of DNS queries. It however is also called when the user has specified that a proxy should be used. This behavior therefore introduces a DNS leak.

This commit ensures that QHostInfo::fromName() is not called when a proxy has been specified.

### DIFF
--- a/src/core/corenetwork.cpp
+++ b/src/core/corenetwork.cpp
@@ -242,8 +242,10 @@ void CoreNetwork::connectToIrc(bool reconnecting)
     // Qt caches DNS entries for a minute, resulting in round-robin (e.g. for chat.freenode.net) not working if several users
     // connect at a similar time. QHostInfo::fromName(), however, always performs a fresh lookup, overwriting the cache entry.
     if (! server.useProxy) {
-		QHostInfo::fromName(server.host);
-	}
+        //Avoid hostname lookups when a proxy is specified. The lookups won't use the proxy and may therefore leak the DNS
+        //hostname of the server. Qt's DNS cache also isn't used by the proxy so we don't need to refresh the entry.
+        QHostInfo::fromName(server.host);
+    }
 #ifdef HAVE_SSL
     if (server.useSsl) {
         CoreIdentity *identity = identityPtr();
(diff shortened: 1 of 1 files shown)


src/core/corenetwork.cpp
```diff
@@ -241,8 +241,9 @@ void CoreNetwork::connectToIrc(bool reconnecting)
 
     // Qt caches DNS entries for a minute, resulting in round-robin (e.g. for chat.freenode.net) not working if several users
     // connect at a similar time. QHostInfo::fromName(), however, always performs a fresh lookup, overwriting the cache entry.
-    QHostInfo::fromName(server.host);
-
+    if (! server.useProxy) {
+		QHostInfo::fromName(server.host);
```
digitalcircuit: Quassel uses spaces rather than tabs, so you might want to change that :)

I'd also suggest adding a small comment to explain why in the code itself, e.g.
```cpp
    if (!server.useProxy) {
        // Don't allow hostname lookups when using a proxy, otherwise Qt leaks the DNS lookup
        // outside of the proxy, e.g. when using TOR.
        QHostInfo::fromName(server.host);
    }
```

crtxcr:
> (Someone may need to confirm if QHostInfo::fromName() still forces a fresh DNS lookup - https://doc.qt.io/qt-5/qhostinfo.html seems to claim there's a DNS cache regardless of method.)

After a quick check of Qt's source and a test with a test program, this still appears to be the case.

digitalcircuit:
The AppVeyor failure is unrelated, 7zip failing to download three times in a row.  Seems Chocolatey did not have a good time...

https://ci.appveyor.com/project/Quassel/quassel/build/512/job/lpr7tvudgm2wcyoq#L1049

A force-push or quick close/reopen would force a rebuild, or one of the project maintainers could do that, too.

Sput42:
Just for the record, I'm the one updating the ChangeLog, and I routinely (before every release) go through all the commits and add that sort of stuff...

PR #302, opened 2017-08-27
Let users see a list of their own connected clients, and disconnect them

justjanne:
## In Short
* Via the coreinfo sync, clients can get a list of all peers of the same user that are connected.
* This includes client version, build date, IP, current lag, and a unique ID
* Introduces an RPC call to kick a peer by unique ID
* The coreinfo dialog shows this list of clients (maybe even with GeoIP?) and allows users to kick them

## Status
- [x] Transmit the relevant information to the client
- [x] Implement RPC call functionality
- [x] Implement the required UI
- [x] Implement graceful disconnects (currently the client just reconnects)

## Rationale
Often the request to kick clients of yourself, or to see currently connected versions (so users can see if an old device is still eating new messages) is requested.

## Problems
The client will always reconnect, no matter the error message, so we might want to add a graceful disconnection feature that requests from the client that it disconnects (without then reconnecting)

## Future Improvement Opportunities
- Once Client Features are implemented, they could be used to detect if a client supports this feature
- Once Session Tokens are implemented, they should probably be integrated with this menu
- Once Qt4 support is dropped, the fallback for it can be removed
- Potentially implement GeoIP

### DIFF
--- a/src/client/client.cpp
+++ b/src/client/client.cpp
@@ -160,6 +160,8 @@ void Client::init()
     p->attachSignal(this, SIGNAL(requestPasswordChange(PeerPtr,QString,QString,QString)), SIGNAL(changePassword(PeerPtr,QString,QString,QString)));
     p->attachSlot(SIGNAL(passwordChanged(PeerPtr,bool)), this, SLOT(corePasswordChanged(PeerPtr,bool)));
 
+    p->attachSignal(this, SIGNAL(requestKickClient(int)), SIGNAL(kickClient(int)));
+
     //connect(mainUi(), SIGNAL(connectToCore(const QVariantMap &)), this, SLOT(connectToCore(const QVariantMap &)));
     connect(mainUi(), SIGNAL(disconnectFromCore()), this, SLOT(disconnectFromCore()));
     connect(this, SIGNAL(connected()), mainUi(), SLOT(connectedToCore()));
@@ -674,6 +676,11 @@ void Client::changePassword(const QString &oldPassword, const QString &newPasswo
     account.setPassword(newPassword);
     coreAccountModel()->createOrUpdateAccount(account);
     emit instance()->requestPasswordChange(nullptr, account.user(), oldPassword, newPassword);
+}
+
+
+void Client::kickClient(int peerId) {
+    emit instance()->requestKickClient(peerId);
 }
 
 
--- a/src/client/client.cpp
+++ b/src/client/client.cpp
@@ -161,6 +161,7 @@ void Client::init()
     p->attachSlot(SIGNAL(passwordChanged(PeerPtr,bool)), this, SLOT(corePasswordChanged(PeerPtr,bool)));
 
     p->attachSignal(this, SIGNAL(requestKickClient(int)), SIGNAL(kickClient(int)));
+    p->attachSlot(SIGNAL(disconnectFromCore()), this, SLOT(disconnectFromCore()));
 
     //connect(mainUi(), SIGNAL(connectToCore(const QVariantMap &)), this, SLOT(connectToCore(const QVariantMap &)));
     connect(mainUi(), SIGNAL(disconnectFromCore()), this, SLOT(disconnectFromCore()));
@@ -679,7 +680,8 @@ void Client::changePassword(const QString &oldPassword, const QString &newPasswo
 }
 
 
-void Client::kickClient(int peerId) {
+void Client::kickClient(int peerId)
+{
     emit instance()->requestKickClient(peerId);
 }
 
--- a/src/common/peer.cpp
+++ b/src/common/peer.cpp
@@ -33,6 +33,37 @@ AuthHandler *Peer::authHandler() const
     return _authHandler;
 }
 
+const QDateTime &Peer::connectedSince() const {
+    return _connectedSince;
+}
+
+void Peer::setConnectedSince(const QDateTime &connectedSince) {
+    _connectedSince = connectedSince;
+}
+
+const QString &Peer::buildDate() const {
+    return _buildDate;
+}
+
+void Peer::setBuildDate(const QString &buildDate) {
+    _buildDate = buildDate;
+}
+
+const QString &Peer::clientVersion() const {
+    return _clientVersion;
+}
+
+void Peer::setClientVersion(const QString &clientVersion) {
+    _clientVersion = clientVersion;
+}
+
+int Peer::id() const {
+    return _id;
+}
+
+void Peer::setId(int id) {
+    _id = id;
+}
 
 // PeerPtr is used in RPC signatures for enabling receivers to send replies
 // to a particular peer rather than broadcast to all connected ones.
--- a/src/common/peer.h
+++ b/src/common/peer.h
@@ -50,8 +50,12 @@ public:
 
     virtual int lag() const = 0;
 
+    virtual QString address() const = 0;
+    virtual quint16 port() const = 0;
+
     int _id = -1;
 
+    QDateTime _connectedSince;
     QString _buildDate;
     QString _clientVersion;
 
--- a/src/common/peer.h
+++ b/src/common/peer.h
@@ -34,13 +34,25 @@ class Peer : public QObject
     Q_OBJECT
 
 public:
-    Peer(AuthHandler *authHandler, QObject *parent = 0);
+    explicit Peer(AuthHandler *authHandler, QObject *parent = 0);
 
     virtual Protocol::Type protocol() const = 0;
     virtual QString description() const = 0;
 
     virtual SignalProxy *signalProxy() const = 0;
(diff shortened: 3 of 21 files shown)


src/common/signalproxy.h
```diff
@@ -77,11 +77,38 @@ class SignalProxy : public QObject
...
     inline int peerCount() const { return _peers.size(); }
     QVariantList peerData();
 
     Peer *peerById(int peerId);
 
+    /**
+     * @return If handling a signal, the Peer from which the current signal originates
+     */
+    Peer *sourcePeer() { return _sourcePeer; }
+
```
digitalcircuit: *Related to these changes, but not a remark on/something to implement in this pull request.*

Theoretically, with this and having the client send the current timezone [during initialization](https://github.com/quassel/quassel/blob/d29a6e9521e27e5d4d86fec82b5daa71085f87a5/src/common/protocols/datastream/datastreampeer.cpp#L166-L173 ), the core would know what timezone a client is in when it makes a request, correct?

One of the concerns someone mentioned to me in the ["`/away` timestamp" pull request](https://github.com/quassel/quassel/pull/270#issuecomment-271189486 ) was that the core timezone might differ from the client, so it may set an unexpected time.  Admittedly this might be more complex than is worthwhile.

src/client/client.cpp
```diff
@@ -161,6 +161,7 @@ void Client::init()
     p->attachSlot(SIGNAL(passwordChanged(PeerPtr,bool)), this, SLOT(corePasswordChanged(PeerPtr,bool)));
 
     p->attachSignal(this, SIGNAL(requestKickClient(int)), SIGNAL(kickClient(int)));
+    p->attachSlot(SIGNAL(disconnectFromCore()), this, SLOT(disconnectFromCore()));
```
digitalcircuit: The Client Features part might be important before 0.13 release to avoid potential "I clicked End Session but it doesn't go away" complaints if someone doesn't upgrade all of their clients at once.

Though at least this will print a warning in the debug logs on older clients about an unknown `SIGNAL`/`SLOT`.

The client could later be changed to show a message in the status bar, too, explaining that another client disconnected it.

src/common/signalproxy.h
justjanne: Correct, and the timezone should be exposed nicely for the core to use, too.

But the issue is more with awayOnDeatch timestamps (where this isn't known), I'd think. We'd need a separate config for timezone.

src/client/client.cpp
justjanne: We're implementing client features to support message modes anyway, so tgis shouldn't be much of an issue.

src/common/signalproxy.h
```diff
@@ -77,11 +77,38 @@ class SignalProxy : public QObject
...
     void dumpProxyStats();
     void dumpSyncMap(SyncableObject *object);
+
+    /**
+     * This method allows to send a signal only to a limited set of peers
+     * @param peerIds A list of peers that should receive it
+     * @param closure Code you want to execute within of that restricted environment
+     */
+    void restrictTargetPeers(QSet<Peer*> peers, std::function<void()> closure);
+    void restrictTargetPeers(Peer *peer, std::function<void()> closure) {
```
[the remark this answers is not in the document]
justjanne: I manually grouped them, this should just work now

src/common/peer.h
```diff
@@ -50,6 +50,15 @@ class Peer : public QObject
...
     virtual int lag() const = 0;
 
+    virtual QString address() const = 0;
+    virtual quint16 port() const = 0;
+
+    int _id = -1;
+
+    QDateTime _connectedSince;
+    QString _buildDate;
+    QString _clientVersion;
```
Sput42: I don't like public attributes. Please make them private and provide getters and setters (could maybe unify the setters of some of those).

src/common/signalproxy.h
```diff
@@ -140,6 +178,7 @@ private slots:
     static void disconnectDevice(QIODevice *dev, const QString &reason = QString());
 
     QSet<Peer *> _peers;
+    QHash<int, Peer*> _peerMap;
```
Sput42: Do we need both of these? `_peers` should be equivalent to `_peerMap.values()`...

src/common/signalproxy.h
```diff
@@ -162,6 +201,13 @@ private slots:
 
     bool _secure; // determines if all connections are in a secured state (using ssl or internal connections)
 
+    int _lastPeerId = 0;
+
+    QSet<Peer *> _restrictedTargets;
+    bool _restrictMessageTarget = false;
+
+    Peer *_sourcePeer;
```
Sput42: You probably should default-initialize this to nullptr, so sourcePeer() always returns something useful

src/core/coresession.cpp
```diff
@@ -693,5 +696,18 @@ void CoreSession::changePassword(PeerPtr peer, const QString &userName, const QS
     if (uid.isValid() && uid == user())
         success = Core::changeUserPassword(uid, newPassword);
 
-    emit passwordChanged(peer, success);
+    signalProxy()->restrictTargetPeers(signalProxy()->sourcePeer(), [&]{
```
Sput42: I think you should add `Q_UNUSED(peer)` above, or remove the parameter name (not sure if we currently trigger warnings for unused parameters, but we very well might in the future)

src/qtui/CMakeLists.txt
```diff
@@ -65,6 +66,7 @@ set(FORMS
     coreconfigwizardsyncpage.ui
     coreconnectauthdlg.ui
     coreconnectionstatuswidget.ui
+        coresessionwidget.ui
```
Sput42: Identation is off, also please alphabetize

src/qtui/coresessionwidget.cpp
```diff
@@ -0,0 +1,58 @@
...
+ *                                                                         *
+ *   You should have received a copy of the GNU General Public License     *
+ *   along with this program; if not, write to the                         *
+ *   Free Software Foundation, Inc.,                                       *
+ *   51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.         *
+ ***************************************************************************/
+
+#include "coresessionwidget.h"
+#include <QtCore/QDateTime>
+#include <client.h>
```
Sput42: `#include "client.h"` (it's an internal header, not a system one)

src/qtui/coresessionwidget.h
```diff
@@ -0,0 +1,48 @@
...
+ *   GNU General Public License for more details.                          *
+ *                                                                         *
+ *   You should have received a copy of the GNU General Public License     *
+ *   along with this program; if not, write to the                         *
+ *   Free Software Foundation, Inc.,                                       *
+ *   51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.         *
+ ***************************************************************************/
+
+#ifndef CORESESSIONWIDGET_H
+#define CORESESSIONWIDGET_H
```
Sput42: New header files should use `#pragma once` instead of a guard

src/qtui/coresessionwidget.h
```diff
@@ -0,0 +1,48 @@
...
+
+#include <QWidget>
+#include <ui_coresessionwidget.h>
+#include <QMap>
```
Sput42: This is a pretty random order of `#include`s :) Please group Qt/system ones and internal ones, and use quotes for the latter. Also we like alphabetized lists (yes, this is not adhered to in most of the legacy codebase, so I'm not gonna be anal about that one until we've done a global cleanup)

src/common/peer.cpp
```diff
@@ -33,6 +33,37 @@ AuthHandler *Peer::authHandler() const
...
 
+const QDateTime &Peer::connectedSince() const {
+    return _connectedSince;
+}
+
+void Peer::setConnectedSince(const QDateTime &connectedSince) {
+    _connectedSince = connectedSince;
+}
+
+const QString &Peer::buildDate() const {
```
Sput42: It's a really bad idea to return const refs (they may dangle if peer dies)

src/common/peer.cpp
```diff
@@ -33,6 +33,37 @@ AuthHandler *Peer::authHandler() const
...
+    return _buildDate;
+}
+
+void Peer::setBuildDate(const QString &buildDate) {
+    _buildDate = buildDate;
+}
+
+const QString &Peer::clientVersion() const {
```
Sput42: Same here

src/qtui/coresessionwidget.h
```diff
@@ -43,6 +42,4 @@ private slots:
 private:
     Ui::CoreSessionWidget ui;
     int _peerId;
-};
-
-#endif //CORESESSIONWIDGET_H
+};
```
Sput42: Newline on EOF...

Sput42:
Fixed my own remaining review comments when pushing the updated branch.

PR #303, opened 2017-09-01
Implement Core-Side Highlights

justjanne:
## In Short
* Store highlight rules in a highlight manager on the core
* Search for highlights in messages on the core
* Transmit this status to the client by setting the highlight flag already on the core
* Implement sender-based highlight rules
* Implement inverse highlight rules (which negate a match)

## Status
- [x] Implement a highlight manager
- [x] Implement core-side highlight parsing
- [x] Implement a UI for configuring core-side highlights
- [x] Assign a CoreFeatures flag for core-side highlights
- [x] Implement inverse highlight rules
- [x] Implement better UI for inverse highlight rules
- [x] Implement sender-based highlight (or ignore highlight) rules?
- [ ] Implement a button to migrate client-side highlight rules to core-side

## Rationale
Currently a client needs to request backlog for all channels, to know if highlights have occured on that channel (and color it differently in the UI, etc). Long term we want to solve that with a notification system, for which this provides the first step, by moving highlight parsing to the core.

## Drawbacks
The core can not re-parse the entire backlog whenever the highlight rules are changed (the client does this), so it can always only take into account the highlight rules at the time that a message was sent.

## Screenshots
[Example showing inverse and sender-based highlight rules in action](https://i.imgur.com/pkPI31N.png)

### DIFF
--- a/src/client/client.cpp
+++ b/src/client/client.cpp
@@ -420,6 +420,11 @@ void Client::setSyncedToCore()
     _ignoreListManager = new ClientIgnoreListManager(this);
     p->synchronize(ignoreListManager());
 
+    // create Core-Side HighlightRuleManager
+    Q_ASSERT(!_highlightRuleManager);
+    _highlightRuleManager = new HighlightRuleManager(this);
+    p->synchronize(highlightRuleManager());
+
     // create TransferManager and DccConfig if core supports them
     Q_ASSERT(!_dccConfig);
     Q_ASSERT(!_transferManager);
@@ -506,6 +511,11 @@ void Client::setDisconnectedFromCore()
     if (_ignoreListManager) {
         _ignoreListManager->deleteLater();
         _ignoreListManager = 0;
+    }
+
+    if (_highlightRuleManager) {
+        _highlightRuleManager->deleteLater();
+        _highlightRuleManager = nullptr;
     }
 
     if (_transferManager) {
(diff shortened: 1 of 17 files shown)


digitalcircuit:
Here's a very bad mockup of what I had in mind...

The current Quassel Highlights tab, with the "Custom Highlights" section duplicated below the "Highlight Nicks" settings.  The topmost "Custom Highlights" section is renamed "Highlight when…" while the bottom "Custom Highlights" section is renamed "But not if…"

Two sections within the `Highlight` pane, one for `Highlight when…` and all the normal highlight rules, then a duplicate bottom pane with `But not if…` containing all the inverse rules.

* Upsides
  * Clear distinction of purpose, room for explanatory text
  * At-a-glance distinguishing of what rules are inverse
* Downsides
  * More information in one page, may not be friendly to small screens
  * Would need dedicated button to convert between regular and inverse rule

Alternatively, maybe a tab-bar could be used, as in the Networks window?  Or other options?

I'm hesistant with keeping this as a single checkbox for `Inverse` given it could be confused for "inverse RegEx", and if someone accidentally adds a match-everything rule with inverse, it might be lost among tens of other highlight rules.

justjanne:
With regards to the split UI, I would propose this:

Screenshot of the Highlight Tab

Screenshot of the Highlight Ignore Tab

digitalcircuit:
New Highlight rules UI looks good!  I like that you've put `Highlight nicks` on the top since it's kind of easy to ignore at the bottom, and we shouldn't really need to ever support migrating a highlight rule into a highlight ignore rule, or vice-versa.

The only other thing I'd suggest is hiding the client-only highlight rules behind a button on the core-side highlight rules page, along with some sort of migration logic, but this can be handled in a separate pull request.  The core-side logic is **much** more important to get in :)

PR #304, opened 2017-09-01
Fix OSX compile and deploy

romibi:
- Set additional stdlib compile flag
   I get compile errors otherwise, at least locally.
- Fix FindLdap.cmake for OSX
  I hope the way I did it is ok. But I know it can be improved!
- ~~Use older OSX image on travis.
  I don't want to upgrade my old MacBook to an newer OSX version … and with the few pushs to update the scripts for newer xcode to build on travis I didn't get deploy to work, yet.
  I might give this a try later again … maybe even with a new macbook …~~
- Fix DMG creation
  hdiutil sometimes fails if no volume size is provided.
  Fixed by taking the size of the contents (and slightly increasing it to be safe)
- cleaned up travis.yml a bit

**New since 2017-12-23:**
- Stop manually overriding CMAKE_OSX_SYSROOT
  This was the main reason why it broke if compiled on a different OSX version.
  CMake should have this variable correctly set already!
- (modified Fix for FindLdap.cmake)

**Things which still can be improved:**
To support KF5 stuff being deployed the deploy script needs to be modified to also rewrite absolute paths to relative paths in the linking inside ".dylib" packages in addition to ".framework" packages.
This can be done in a later PR but until that is done we can ship KF5 stuff with the OSX app.
(At least not automatically.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 os:
   - linux
-  - osx
 
 sudo: required
 dist: trusty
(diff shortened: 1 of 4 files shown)


romibi:
doesn't work anymore …
mainly because of the older OSX image?

PR #305, opened 2017-09-03
Initialize member variables in constructor

Kissaki:
These were missing.


Sput42:
This commit introduces lots of -Wreorder warnings (constructor shall initialize member variables in declaration order). So they need to be reordered.
Since we're using C++11 now, one can also consider initializing the members directly in the declaration in the header.

Sput42:
... of course, thanks for finding all these :)

PR #306, opened 2017-09-03
Use Travis’ container-based environment for better build performance

justjanne:
## In Short
- Disables sudo requirement for builds
- Accesses dependencies via Travis’ config syntax
- Improves build performance and waiting time

## Status
- [x] Port builds over
- [ ] Get the QCA dependency back, which is currently not supported by Travis

## Rationale
Travis provides a container-based, and a virtual machine based environment for builds. The container-based environment has far better performance, but requires minor adaptions. This PR aims to switch to that environment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ os:
   - linux
   - osx
 
-sudo: required
+sudo: false
 dist: trusty
 
 language: cpp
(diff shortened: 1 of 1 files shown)


digitalcircuit:
For reference, there's an [open issue](https://github.com/travis-ci/apt-package-whitelist/issues/1988 ) and [pull request to add `libqca2` to Travis CI's whitelist](https://github.com/travis-ci/apt-package-whitelist/pull/1997#issuecomment-224225139 ).  `setuid` bits were found, but at the time, they were merely to drop root privileges, nothing easily exploitable.  And there's been no response for over a year.

Perhaps a polite email from Quassel to the Travis CI support team (contact@travis-ci.com) would help?

digitalcircuit:
In [my past attempt]( https://github.com/digitalcircuit/quassel/tree/ft-travis-container ), I had modified the `make` call to [adapt to the number of processors](https://github.com/digitalcircuit/quassel/blob/ft-travis-container/.travis.yml#L80-L82 ).  E.g.
```diff
-    make
+    # nproc returns the number of processors.  Let's use all of them!
+    # Travis' container infrastructure has 2 cores available.
+    make -j $(nproc)
```

Though I was debugging something with `nproc` returning the wrong number.  I'm not sure if that's still an issue.  The `make` call itself might not need changed anymore, either; I'm not sure.

justjanne:
@digitalcircuit We actually want to avoid using make at all, and use ninja instead. It provides far better performance, and cmake can use it automatically, too.

Sput42:
For the time being, this doesn't seem to give us what we need. On the contrary, we'd like to move towards supporting multiple build environments through docker-inside-travis-host.

May revisit if Travis decides to update their host systems, and maybe offers a more efficient way to run docker images.

PR #309, opened 2017-09-06
Implement Client Feature Flags and Sender Prefixmodes in Protocol and UI

justjanne:
## In short
- Transmit in the RegisterClient message the client features as int32
- Make the current SignalProxy available within of the thread for the
  current user
- Store the features of a peer in a flag in it
- Use Peer features to determine how to serialize or deserialize a
  message
- Add the prefixes to the UI when formatting the sender

## Dependencies
- Due to the dependency on being able to query the source peer of a message, and the target peer, this PR depends on #302

## Status
- [x] Implement Protocol
- [x] Implement UI
- [x] Implement backwards-compatible (de-)serialization

## Rationale
There have been multiple feature requests made asking to support sender modes, and displaying them in the client.

### DIFF
--- a/src/client/clientauthhandler.cpp
+++ b/src/client/clientauthhandler.cpp
@@ -288,7 +288,7 @@ void ClientAuthHandler::startRegistration()
     useSsl = _account.useSsl();
 #endif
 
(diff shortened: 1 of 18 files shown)


Nevcairiel:
Can we keep the sender prefixes optional? I never missed them, and don't really want them.

justjanne:
@Nevcairiel Sure, this can easily be made a UI setting

PR #310, opened 2017-09-06
Only show the disconnect button in the client list if the client supports it

justjanne:
## In Short
* Via the coreinfo sync, transmit the peer features of each client.
* Hide the disconnect button if the "ClientFeatures" flag isn’t set

## Dependencies
- This depends on the client features and peer-dependent sync introduced in #309 
- Transitively, this depends on #302

## Rationale
This reduces potential confusion


justjanne:
Replaced by #316

PR #312, opened 2017-09-22
Use craft on appveyor

TheOneRing:
This will use the well maintained packages from Craft.
It will reduce the maintenance cost for providing Windows installer.
Additionally we now use up to date binaries build exactly for this compiler.
So we no longer depend on 3dparty  openssl binaries etc.


appveyor.ini
```diff
@@ -0,0 +1,50 @@
+## The cash used was created by the command
```
digitalcircuit: *Minor:* `cash` → `cache`

appveyor.ini
```diff
@@ -0,0 +1,50 @@
...
+## Craftmaster.py --config appveyor.ini --variables Ignores= CreateCache=True UseCache=False -c --install-deps quassel
+
+# Avalible variables are
```
digitalcircuit: *Minor:* `Avalible` → `Available`

appveyor.ini
```diff
@@ -0,0 +1,50 @@
...
+Command = craft
+Branch = master
+ShallowClone = True
+
+# Variables defined here override the default value
+# The variable names are casesensitive
+[Variables]
+#Root = D:\qt-sdk
+
+#Values need to be overwritten to create a chache
```
digitalcircuit: *Minor:* `chache` → `cache`

appveyor.ini
```diff
@@ -0,0 +1,50 @@
...
+# Category is case sensitive
+[GeneralSettings]
+Paths/Python = C:\Python36
+Paths/Python27 = C:\Python27
+Paths/DownloadDir = ${Variables:Root}\downloads
+ShortPath/Enabled = False
+Packager/CacheDir = ${Variables:Root}\cache
+Packager/UseCache = ${Variables:UseCache}
+Packager/CreateCache = ${Variables:CreateCache}
+Packager/RepositoryUrl = https://s3.kuschku.de/quassel
```
digitalcircuit: Does `https://s3.kuschku.de/quassel` need changed to a Quassel-project hosted server once merged?

It'd be nice to use an HTTPS endpoint.  Not sure how difficult adding Let's Encrypt to the Quassel site would be, but that needs done given browsers are marking the homepage insecure...

TheOneRing: Any server would work, things would be easier when I can upload new files on demand.

TheOneRing:
Qt does not offer 32bit build of msvc2017 not 2015.
The Craft setup will also enable us to provide KDE Frameworks 5 libraries on Windows or other 3party libraries.

digitalcircuit:
Awesome work!  Alongside supporting later compilers, having an easier way to integrate libraries like `Sonnet` and `KIdleTime` for Windows greatly helps.

Thank you for your prompt replies :)

TheOneRing:
The script now uses the binaries provided by KDE, that means we don't need to maintain a second binary cache

TheOneRing:
Patches to the blueprint can be submitted via the KDE infrastructure or by a simple GitHub pr

digitalcircuit:
@TheOneRing Should changes to Quassel's dependencies be handled in a follow-up pull request?  Or is that entirely dependent on [the Craft repository's `quassel.py`](https://github.com/KDE/craft-blueprints-kde/blob/master/qt-apps/quassel/quassel.py )?

From looking around, to add Sonnet for spell-checking on Windows appears as though it'd just be:
```diff
   self.runtimeDependencies["libs/qt5/qtscript"] = "default"
   self.runtimeDependencies["libs/qt5/qttools"] = "default"
+  self.runtimeDependencies["frameworks/tier1/sonnet"] = "default"
```

But that seems too easy to be correct...

TheOneRing:
Yes the change has to be done in Craft, also we would need to adjust the whitelist.
For sonnet, right now we still have some problems with aspell but I guess we are getting there :)

TheOneRing:
Now with sonnet :)

romibi:
Assuming only qgenericbearer.dll is needed and not qnativewifibearer.dll this also makes PR #319 unnecessary. Closed that now.

PR #316, opened 2017-12-21
Polish CoreSession UI

justjanne:
## In short

- Introduce feature flag for remote disconnecting
- Transmit feature flag for remote disconnecting with peer info
- Only show disconnect button if available
- Show if client is connected securely
- Make the UI look more native

## UI

Screenshot of revamped Core Session UI

## Protocol Changes

- Introduces new Feature Flag (we’re running out of those, we’ll need to address that soon)


digitalcircuit:
**Minor:**  If it's possible to detect the current session, we probably shouldn't show an End Session button on it.  Alternatively, the CoreSession UI should detect when connection to the core is lost so it doesn't seem to still be connected (*especially with the uptime still ticking away via timer*).

PR #301, opened 2017-08-25
Move buffer activity tracking to the core

justjanne:
## In Short
* Store the bitset of all message types that have occured in a buffer since the last seen message on the core
* Sync this with clients

## Status
- [x] Implement database backend
- [x] Integrate the activity tracking into the buffersyncer
- [x] Implement the actual sync of the activity between core and client
- [x] Assign a CoreFeatures flag
- [x] Have the client make use of this feature

## Rationale
Currently a client needs to request backlog for all channels, to know if events have occured on that channel (and potentially unhide it, show it differently in the UI, etc). This patch aims at removing that requirement by allowing the client to know the set of message types that have occured in a channel.

## Disadvantages
The core can not take ignore list items into account, so under some circumstances a channel might be marked as having new messages, but the client might not show them once the channel is activated. Additionally, this does not currently process highlights yet.

PR #317, opened 2017-12-21
Polish Core Highlights Page

justjanne:
## Status

- [x] Disable the highlights page when not connected
- [x] Offer import
- [x] Hide client-side highlight rules when connected to a modern core

### DIFF
--- a/src/core/SQL/PostgreSQL/select_buffer_bufferactivity.sql
+++ b/src/core/SQL/PostgreSQL/select_buffer_bufferactivity.sql
@@ -3,4 +3,5 @@ FROM
   (SELECT DISTINCT TYPE
    FROM backlog
    WHERE bufferid = :bufferid
+     AND flags & 1 = 0
      AND messageid > :lastseenmsgid) t;
--- a/src/core/SQL/SQLite/select_buffer_bufferactivity.sql
+++ b/src/core/SQL/SQLite/select_buffer_bufferactivity.sql
@@ -3,4 +3,5 @@ FROM
   (SELECT DISTINCT TYPE
    FROM backlog
    WHERE bufferid = :bufferid
+     AND flags & 1 = 0
      AND messageid > :lastseenmsgid) t;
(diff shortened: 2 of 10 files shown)


digitalcircuit:
Other options we discussed on IRC for handling local vs. core-side highlights...

* Hide local highlights if core-side highlights work, put `Modify` button on core-side page next to `Import` to access local highlights
  * *Details of UI to be determined*
  * Avoids confusing new users (*what does local/remote mean, why do local highlights not work with no backlog fetch, etc*)
  * Allows power-users to access the old functionality
* Keep local highlights always available as dedicated tab (*this PR as of 2018-1-5*)
   * Easy access to local highlights, minimal changes
   * Potentially confusing to new users
* *Other options..?*

Regardless, if local rules are allowed, we should make it clear on upgrading that core-side highlights and local may conflict.  If possible, perhaps disable local highlights by default/on clicking `Import`, and add a checkbox to the local highlights settings page to always enable them?

If adding a button to `Modify`/access local highlight rules on the core-side highlights page, not keeping at top-level, the checkbox could be there.

digitalcircuit:
A few other things:

1.  Add tool-tips to the list box and other GUI elements, explaining what the options do
2.  Highlighting the current nick by default is how most IRC clients work - is it possible to enable that on the core-side by default?
3.  When selecting `None` for highlight nicks, `Case Sensitive` should be marked as disabled
4.  *Minor:* When switching tabs between `Highlight Rules` and `Highlight Ignore Rules`, the list box isn't aligned (the QGroupBox headers in the `Highlight Rules` page shrinks the list box down).  *Possibly fixable by adding a header to the Ignore page, or something else.*

romibi:
About "5.", note that I mainly meant the marking of the channel/query in the channel list in orange. Which (afaik) last got modified in #301 …
Although, true … highlights/notifications shouldn't get triggered either.

digitalcircuit:
I've migrated my suggestions over to [pull request #336](https://github.com/quassel/quassel/pull/336 ), so this one can be merged without losing comments.

Ignoring self-messages appears to already be addressed, not tested.

PR #319, opened 2017-12-22
Package bearer plugin with Appveyor builds

romibi:
Adds 2 dll's … about 100KB …
Fixes Automatic Network Status Detection on Windows

Probably needs changed after #312


romibi:
Also fixed in #312

PR #320, opened 2017-12-22
Make /list open the Channel List dialog

justjanne:
## In Short

When a user runs /list, the channel list dialog is opened for the current network.

## Rationale

This was previously requested in several issues

## Impact

This shadows the actual /list command, but that command can still be accessed with /RAW LIST or /QUOTE LIST.

### DIFF
--- a/src/client/client.h
+++ b/src/client/client.h
@@ -167,6 +167,7 @@ signals:
     void requestNetworkStates();
 
     void showConfigWizard(const QVariantMap &coredata);
+    void showChannelList(NetworkId networkId);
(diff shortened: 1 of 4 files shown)


digitalcircuit:
AppVeyor failure appears to be unrelated, caused by something having trouble when Craft is installing dependencies: [`postInstall() takes 0 positional arguments but 1 was given`](https://ci.appveyor.com/project/Quassel/quassel/build/722/job/vvwfnbb5cq8rmroa#L1343 )

Just re-run the build when that's worked out, which I'll try to do.

PR #321, opened 2018-01-08
$ident for ircUser->user

SvetlanaTkachenko:
https://bugs.quassel-irc.org/issues/1454

### DIFF
--- a/src/common/aliasmanager.cpp
+++ b/src/common/aliasmanager.cpp
@@ -190,6 +190,7 @@ void AliasManager::expand(const QString &alias, const BufferInfo &bufferInfo, co
             IrcUser *ircUser = net->ircUser(params[j - 1]);
             command = command.replace(QString("$%1:hostname").arg(j), ircUser ? ircUser->host() : QString("*"));
             command = command.replace(QString("$%1:ident").arg(j), ircUser ? ircUser->user() : QString("*"));
+            command = command.replace(QString("$%1:account").arg(j), ircUser ? ircUser->account() : QString("*"));
             command = command.replace(QString("$%1").arg(j), params[j - 1]);
         }
         command = command.replace("$0", msg);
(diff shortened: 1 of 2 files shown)


src/common/aliasmanager.cpp
```diff
@@ -189,6 +189,7 @@ void AliasManager::expand(const QString &alias, const BufferInfo &bufferInfo, co
         for (int j = params.count(); j > 0; j--) {
             IrcUser *ircUser = net->ircUser(params[j - 1]);
             command = command.replace(QString("$%1:hostname").arg(j), ircUser ? ircUser->host() : QString("*"));
+            command = command.replace(QString("$%1:ident").arg(j), ircUser ? ircUser->user() : QString("*"));
             command = command.replace(QString("$%1").arg(j), params[j - 1]);
```
digitalcircuit: Since you're adding an alias for `ident`, perhaps it'd be a good time to add one for `account`, too?  Several IRC servers allow you to ban by account, including Freenode (I think).  E.g...

```cpp
command = command.replace(QString("$%1:account").arg(j), ircUser ? ircUser->account() : QString("*"));
```

**Untested**, and I'm not sure if `*` is a good default value.  `*` also represents a known user that's logged out.

*In hindsight, I should've done this as part of the [IRCv3 account tracking pull request](https://github.com/quassel/quassel/pull/198 ), which added [account information to the `IrcUser` objects](https://github.com/quassel/quassel/blob/738f5436bbc9a8c761a487ae24da65fd74a62ef9/src/common/ircuser.h#L69-L74 ).*

genius3000: It would almost be nicer to fail than set an account ban on ` ` or `*`. I realize that both hostname and ident will fall back to `*` as well, but they are less likely to be unknown than the account. This would be outside the scope of this PR and would be a new feature/bug fix in itself though.  
As is the code tests fine, just with an unintended ban if the account is unknown or logged out.

Sput42:
I took the liberty to follow @digitalcircuit's suggestion and squash the commits, also clarified the commit msgs.

PR #322, opened 2018-01-11
Fix the Persistent Modes reset

genius3000:
It is intended that `/mode -reset` will reset the Persistent Modes but this always fails to function due to having a buffer name always prepended to the string. Fix this by moving the '-reset' check to be first.   Also add a reply to the client, not sure if there's more I need to do for "adding a string" in terms of translations and so forth.

#### Use-Case:
- Sometimes you end up with persistent modes that you can't seem to get rid of (likely a bug to dig more into, but this at least gives you existing functionality to use)
- Simply to clear your persistent modes for other various reasons

#### Impact:
- Minor: using existing functionality, new success message shown to client when used

#### Testing:
- Tested with latest master core and client, sending `/mode -reset` works as expected with the new success message shown. Other `/mode` functionality remained proper.

### DIFF
--- a/src/core/coreuserinputhandler.cpp
+++ b/src/core/coreuserinputhandler.cpp
@@ -477,15 +477,16 @@ void CoreUserInputHandler::handleMode(const BufferInfo &bufferInfo, const QStrin
     QStringList params = msg.split(' ', QString::SkipEmptyParts);
     // if the first argument is neither a channel nor us (user modes are only to oneself) the current buffer is assumed to be the target
     if (!params.isEmpty()) {
+        if (params[0] == "-reset" && params.count() == 1) {
+            network()->resetPersistentModes();
+            emit displayMsg(Message::Info, BufferInfo::StatusBuffer, "",
+                            tr("Your persistent modes have been reset."));
+            return;
+        }
         if (!network()->isChannelName(params[0]) && !network()->isMyNick(params[0]))
             params.prepend(bufferInfo.bufferName());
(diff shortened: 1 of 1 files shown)


src/core/coreuserinputhandler.cpp
```diff
@@ -477,15 +477,16 @@ void CoreUserInputHandler::handleMode(const BufferInfo &bufferInfo, const QStrin
...
-            network()->updateIssuedModes(params[1]);
         if (params[0] == "-reset" && params.count() == 1) {
-            // FIXME: give feedback to the user (I don't want to add new strings right now)
             network()->resetPersistentModes();
+            emit displayMsg(Message::Info, BufferInfo::StatusBuffer, "",
+                            tr("Your persistent modes have been reset."));
             return;
         }
+        if (!network()->isChannelName(params[0]) && !network()->isMyNick(params[0]))
+            params.prepend(bufferInfo.bufferName());
```
digitalcircuit: **Non-issue remark:**  I'm assuming this is for doing e.g. `/mode +b` in a channel without having to specify the channel name?  Never realized this was a feature, and it appears to still work!

genius3000: Correct and it being first was why `-reset` didn't work. A separate improvement could be to prepend current nick when running from the StatusBuffer. I find myself running "/m +<modes>" often (/m aliased to /mode) and forgetting to put my nick :)

PR #326, opened 2018-01-31
exclude ignored messages (maybe Closes #1472)

SvetlanaTkachenko:
untested

### DIFF
--- a/src/qtui/awaylogfilter.cpp
+++ b/src/qtui/awaylogfilter.cpp
@@ -19,6 +19,7 @@
  ***************************************************************************/
 
 #include "awaylogfilter.h"
+#include "clientignorelistmanager.h"
 
 AwayLogFilter::AwayLogFilter(MessageModel *model, QObject *parent)
     : ChatMonitorFilter(model, parent)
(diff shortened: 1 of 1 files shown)


digitalcircuit:
I'd try to test this, but I'm not sure how the `Away Log` feature works in the first place.  Do you have any suggestions?

SvetlanaTkachenko:
When detached, highlights are added to away log.

To repro this try to

1. ignore someone (your alt nick if you want)
2. detach quassel client
3. have the someone highlight you
4. attach back

expected result - their message is not in the away log because they are ignored
actual result - their message is in the away log still

src/qtui/awaylogfilter.cpp
```diff
@@ -18,6 +18,7 @@
  *   51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.         *
  ***************************************************************************/
 
+#include "clientignorelistmanager.h"
```
digitalcircuit: **Minor:** This should go below `#include "awaylogfilter.h"`, to keep includes in alphabetical order.

src/qtui/awaylogfilter.cpp
```diff
@@ -33,16 +34,25 @@ bool AwayLogFilter::filterAcceptsRow(int sourceRow, const QModelIndex &sourcePar
...
 
+    // only show messages that were sent after the user detached
     if (Client::networkModel()->lastSeenMsgId(bufferId) >= sourceModel()->data(source_index, MessageModel::MsgIdRole).value<MsgId>())
         return false;
+    
+    // ignorelist handling
+    // only match if message is not flagged as server msg
+    if (!(flags & Message::ServerMsg) && Client::ignoreListManager()
+        && Client::ignoreListManager()->match(source_index.data(MessageModel::MessageRole).value<Message>(), Client::networkModel()->networkName(bufferId)))
+return false;
```
digitalcircuit: **Minor:** `return false` should be indented to be within the `if` statement scope.

```cpp
    [...]
    if (!(flags & Message::ServerMsg) && Client::ignoreListManager()
            && Client::ignoreListManager()->match(source_index.data(MessageModel::MessageRole).value<Message>(), Client::networkModel()->networkName(bufferId)))
        return false;
    [...]
```

PR #327, opened 2018-02-01
Improve Breeze Tray Icon and Icon Loading

romibi:
This is how it looks now in Breeze Light, Breeze Dark and Oxygen:

[https://gyazo.com/23a8c86b78f73b5c095e5a136b95a541](https://gyazo.com/23a8c86b78f73b5c095e5a136b95a541)
Update: Screenshot outdated: Breeze Taskbar Icons are now replaced by @justjanne's redesigned icon. 
(Oxygen Icons are unchanged.)

Previously for the tray icon the same icon was used as for the taskbar/app-icon.

This one (which isn't used in the taskbar) is still in the repo but now unused: (It was added by PR #283)

Additionally I enabled Breeze-Dark in addition to Breeze-Light by default so that users can choose which one they want in the settings…

~In future it might make sense to make the dark/light theme selection independent to the theme used in the quassel gui.~
Now you can choose if you want a light or dark tray icon independent from if you use light or dark breeze icon theme.

~Note: Tray-Blinking is disabled by Snore. This is something i want to change.~
See also: PR #330


Sput42:
We need to be sure that all this works e.g. on Linux, where we don't usually ship our own icon sets. Also we should display sensible icons if a system icon theme different from the bundled one is selected. I.e. we should not override the icons from a non-Breeze/Oxygen theme; but we should see if we find a way to override the Breeze icons with our own Breeze version...

romibi:
I started trying to fix the icons issue here: https://github.com/romibi/quassel/tree/ft-improve-breeze-tray-icon-linux-fix-v1-WIP
Some attempts while compiling on ubuntu showed that it is probably better to completely change how extra icons are handled.
The linked Changes are untested on linux, yet …  Mainly I don't know how it behaves with
-DEMBED_DATA=OFF
And i'm pretty sure that the tray icons for dbus won't work yet … the statusnotifieritem provides only icon-names and i doubt that is enough for dbus to find the extra icons … i saw there's a way to provide pixmaps which may be easier to use …
but currently I can't boot my ubuntu anymore … and i'm not in a mood to fix that at this time

romibi:
Ok this should work now …

Extra-Icons are now always either Embedded or put into ${datadir}/icons/extra/ on install.

Icon priority for ExtraIcon::load() when not using breeze, breezedark or oxygen:
quassels hicolor icon > the theme's icon > the provided extra icon for breeze

When using breeze, breezedark or oxygen it searches first for the ExtraIcon of that name, if none are found it strips the last part separated by "-" and tries again (similar to how QIcon::fromTheme works) …

If that doesn't help it defaults to ":/icons/iconname.png" which hopefully is in hicolor (but as the extra-icons are under our control this shouldn't happen).

Note: The Tray icon when using StatusNotifierItem seems to not properly update on connect/disconnect … but as far as I can tell this has nothing to do with my changes … once the tray icon gets "forcefully" updated by a notification it picks the correct icon afterwards.

romibi:
Updated to use justjannes generated icons and squashed some commits

src/qtui/settingspages/appearancesettingspage.ui
```diff
@@ -242,6 +242,22 @@
        </property>
       </widget>
      </item>
```
romibi: hm … here a spacer item is missing … and conflicts with #330 (my own PR)

src/qtui/settingspages/appearancesettingspage.ui
```diff
@@ -242,6 +242,22 @@
...
+     <item row="2" column="1">
+       <widget class="QCheckBox" name="iconthemeTrayInvertBox">
```
digitalcircuit: Small, friendly heads-up: you should add a new signal/slot combination to disable this checkbox whenever `Show system tray icon` is disabled.

For example,

```xml
<connection>
    <sender>useSystemTrayIcon</sender>
    <signal>toggled(bool)</signal>
    <receiver>iconthemeTrayInvertBox</receiver>
    <slot>setEnabled(bool)</slot>
    <hints>
     <hint type="sourcelabel">
<!-- I don't know what the hints should be; if you use QtCreator to add
  the signal/slot on the dialog pane, it should do this for you -->
      <x>91</x>
      <y>143</y>
     </hint>
     <hint type="destinationlabel">
      <x>92</x>
      <y>174</y>
     </hint>
    </hints>
</connection>
```

Qt Creator should let you add this using a section at the bottom of the dialog designer.

romibi:
Thanks @digitalcircuit … done

Not sure if i should push the following commit into this: https://github.com/romibi/quassel/commit/369c2a9151c805a253efa8adfa0a09273c46c54d

And here my analysis about missing icons under Ubuntu Unity:
https://gist.github.com/romibi/ab1a40ad70bdb0c266122938bf4ce334

Not sure what to do about those:
- move them all to the new "extra-icons" and always install/embed them?  
  Doesn't make sense because "extra-icons" are custom icons
- copy them over to the (always embedded) hicolor?  
  hicolor are (currently) all Oxygen styled but that might change … but i'm not sure
- try to find them in the system in the theme-search-paths using some custom logic?  
  might be the most reasonable option? it would also allow the detection of dark vs non-dark fallback …
- some other option I missed?

romibi:
to fix missing icons i now have: https://github.com/romibi/quassel/tree/ft-missing-icons-variant-oxygen
and https://github.com/romibi/quassel/tree/ft-missing-icons-variant-breeze
the second one looks better but i'm unhappy with the last commit … and we probably would have to ship a dark stylesheet in addition to the default one

YamashitaRen:
trayicon's icon is "empty" in LXQt if using theses branches. No issue on Mate.
This is probably related to https://bugs.quassel-irc.org/issues/1383.

Maybe you should check it yourself if you have time...

romibi:
@YamashitaRen tested here in different desktop environment with https://github.com/romibi/quassel/tree/ft-missing-icons-variant-oxygen and #330 combined … either no tray icon is shown at all (not even an empty one) or the tray has an icon

YamashitaRen:
Oh I see...
So there are three cases : 
- tray icon works as expected
- no tray icon
- "empty icon" tray icon

romibi:
the thing is … i tested LXQt and it worked … maybe I need to test a cleaner install … perhaps in a VM …

YamashitaRen:
Ubuntu 17.04, mini (netinstall) iso, checking LXQt as DE when installing the OS.
`apt install --no-install-recommends build-essentials qtbase5-dev qtscript5-dev`
`cmake .. -DEMBED_DATA=1`
Using https://github.com/romibi/quassel/tree/ft-missing-icons-variant-breeze, quassel tray icon is a "gear" (an empty icon).

romibi:
true … there's a gear icon using ubuntu 17.10 mini iso (can't find 17.04 anymore) and selecting lubuntu Qt (which is LXQt right?) …
i'll see if i can figure out what's going on …
and … @YamashitaRen are you in #quassel on freenode under some nick? (i'm romibi there too)

YamashitaRen:
Oups, my bad, I mistyped. I used 17.10 iso and selected Lubuntu Qt like you. This is indeed LXQt.
I just joined #quassel. Feel free to HL (same pseudo) if you need my help for anything.

Sput42:
OK, after reviewing this I have reached the conclusion that we're making things too complex to work reliably. We should not bypass Qt's own icon loading mechanism, which not only properly supports themes and fallbacks, but also has some useful things like caching. Instead of reimplementing this, we should make proper use of theme search paths to inject our own icons as appropriate.

As for bundled icon themes, the main use case for this is our own Windows and OSX builds. Linux distros should just hard-depend on breeze-icons (or enable the option to install/bundle the full theme in Quassel, if they so choose). According to @romibi 's analysis, most other icon themes seem to be missing vital icons. I don't think we should work around this by always bundling even more non-quassel-specific icons, but just rely on the distros here.

We probably still need to find a way to ensure that an installed Breeze is used as a fallback in case the user selected another desktop theme. Maybe we can just ship and use a dummy theme that inherits from the desktop theme as well as from Breeze...

I'll use the commits from this PR that still make sense, and go from there.

romibi:
Well … except for the few custom icons i did not completely bypass the qt icon loading …
The custom "fromTheme" in "[...]-variant-breeze" only is one further **indirection** (not replacement) to have control over the icon fallback variant "light" or "dark"
But yes. If anyone finds a cleaner solution that works everywhere I'm up to it.

"[...]but just rely on the distros here.[...]"
Well … I think we relied on the distros too much in the past … which is why I've never seen a Quassel under linux with all the icons because no fallback got found … (Granted, I don't use linux too often)
And a lot of distros don't ship Breeze by default! Should we rely on them to choose to embed breeze in the build settings?
In my opinion for any icon we use we should have at least a fallback icon we ship!
(Which looks like it is done via hicolor by default? Or is there a better way?)

But yea … this whole Icon issue is a mess … and slowly i don't care about it anymore on any other system than windows …
Having an own PR open which started as a small change and evolved into trying to fix a mess with a mess … makes me get in a bad mood everytime i look at quassel source …

"[...] I'll use the commits from this PR that still make sense, and go from there."
Thank you for this sentence … makes me feel a bit … relieved

PR #331, opened 2018-02-07
Correct issue where messages were over-trimmed

kinnison:
In an attempt to correct a problem with quasseldroid multiline
messages (97a9b1646bb0d6362cef14bac3a577481cc01e49) the core started
trimming leading whitespace from lines.  This broke pasting of things
like formatted code.

This commit causes messages to only have CRs trimmed, as the comment
originally provided suggests it ought to.

### DIFF
--- a/src/core/coreuserinputhandler.cpp
+++ b/src/core/coreuserinputhandler.cpp
@@ -460,8 +460,7 @@ void CoreUserInputHandler::handleMe(const BufferInfo &bufferInfo, const QString 
     QStringList messages = msg.split(QCharLF);
 
     foreach (auto message, messages) {
-        // Handle each separated message independently, ignoring any carriage returns
-        message = message.trimmed();
+        // Handle each separated message independently
         coreNetwork()->coreSession()->ctcpParser()->query(coreNetwork(), bufferInfo.bufferName(),
                                                           "ACTION", message);
         emit displayMsg(Message::Action, bufferInfo.type(), bufferInfo.bufferName(), message,
@@ -532,8 +531,7 @@ void CoreUserInputHandler::handleNotice(const BufferInfo &bufferInfo, const QStr
     QStringList messages = msg.section(' ', 1).split(QCharLF);
 
     foreach (auto message, messages) {
-        // Handle each separated message independently, ignoring any carriage returns
-        message = message.trimmed();
+        // Handle each separated message independently
         params.clear();
         params << serverEncode(bufferName) << channelEncode(bufferInfo.bufferName(), message);
         emit putCmd("NOTICE", params);
@@ -607,8 +605,7 @@ void CoreUserInputHandler::handleQuery(const BufferInfo &bufferInfo, const QStri
     QStringList messages = msg.section(' ', 1).split(QCharLF);
 
     foreach (auto message, messages) {
-        // Handle each separated message independently, ignoring any carriage returns
-        message = message.trimmed();
+        // Handle each separated message independently
         if (message.isEmpty()) {
             emit displayMsg(Message::Server, BufferInfo::QueryBuffer, target,
                             tr("Starting query with %1").arg(target), network()->myNick(),
@@ -656,11 +653,10 @@ void CoreUserInputHandler::handleSay(const BufferInfo &bufferInfo, const QString
 
     // Split apart messages at line feeds.  The IRC protocol uses those to separate commands, so
     // they need to be split into multiple messages.
-    QStringList messages = msg.split(QCharLF);
+    QStringList messages = msg.split(QCharLF, QString::SkipEmptyParts);
 
     foreach (auto message, messages) {
-        // Handle each separated message independently, ignoring any carriage returns
-        message = message.trimmed();
+        // Handle each separated message independently
 #ifdef HAVE_QCA2
         putPrivmsg(bufferInfo.bufferName(), message, encodeFunc,
                    network()->cipher(bufferInfo.bufferName()));


src/core/coreuserinputhandler.cpp
```diff
@@ -461,7 +461,8 @@ void CoreUserInputHandler::handleMe(const BufferInfo &bufferInfo, const QString
 
     foreach (auto message, messages) {
         // Handle each separated message independently, ignoring any carriage returns
-        message = message.trimmed();
+        while (message.at(message.size() - 1) == '\r')
```
digitalcircuit: *Same feedback as for `CoreUserInputHandler::handleSay()` applies here.*

src/core/coreuserinputhandler.cpp
```diff
@@ -533,7 +534,8 @@ void CoreUserInputHandler::handleNotice(const BufferInfo &bufferInfo, const QStr
...
```
digitalcircuit: *Same feedback as for `CoreUserInputHandler::handleSay()` applies here.*

src/core/coreuserinputhandler.cpp
```diff
@@ -608,7 +610,8 @@ void CoreUserInputHandler::handleQuery(const BufferInfo &bufferInfo, const QStri
...
```
digitalcircuit: *Same feedback as for `CoreUserInputHandler::handleSay()` applies here.*

src/core/coreuserinputhandler.cpp
```diff
@@ -660,7 +663,8 @@ void CoreUserInputHandler::handleSay(const BufferInfo &bufferInfo, const QString
...
```
digitalcircuit: This causes the Quassel core to crash when `message.size()` is 0, out of range.  [`message.at()` requires the index to be within the string](https://doc.qt.io/qt-5/qstring.html#at ).

Test case:
1. Write the following in a text editor on Android
```
Test
Newlines
```
2.  Copy it to the clipboard
3.  Paste it into Quasseldroid
4.  Try sending it

`messages` arrives in two batches, namely `["Testing"]` and `["", "Newlines"]`.

### Options to fix
1. **Ignore carriage returns entirely**
    * Guarding against carriage returns was just in case; I'm not sure they happen.  Switching the `msg.split()` to `SkipEmptyParts` might be enough, assuming nobody sends `\r\n`-based messages.
    * **Seems like the simplest option**
```diff
     // Split apart messages at line feeds.  The IRC protocol uses those to separate commands, so
     // they need to be split into multiple messages.
-    QStringList messages = msg.split(QCharLF);
+    QStringList messages = msg.split(QCharLF, QString::SkipEmptyParts);
 
     foreach (auto message, messages) {
-        // Handle each separated message independently, ignoring any carriage returns
+        // Handle each separated message independently
-        while (message.at(message.size() - 1) == '\r')
-            message.chop(1);
 #ifdef HAVE_QCA2
 […]
     }
```

2. **Handle carriage returns, modify message split to fix crash**
   * Works with the `while` loop option
```diff
     // Split apart messages at line feeds.  The IRC protocol uses those to separate commands, so
     // they need to be split into multiple messages.
-    QStringList messages = msg.split(QCharLF);
+    QStringList messages = msg.split(QCharLF, QString::SkipEmptyParts);
 
     foreach (auto message, messages) {
         // Handle each separated message independently, ignoring any carriage returns
         while (message.at(message.size() - 1) == '\r')
             message.chop(1);
 #ifdef HAVE_QCA2
 […]
     }
```

3. **Handle carriage returns, modify message trimming to remove loop**
   * Possibly simpler, though it removes carriage returns at beginning, too.  The `msg.split` change isn't needed, but reduces wasted function calls.
```diff
     // Split apart messages at line feeds.  The IRC protocol uses those to separate commands, so
     // they need to be split into multiple messages.
-    QStringList messages = msg.split(QCharLF);
+    QStringList messages = msg.split(QCharLF, QString::SkipEmptyParts);
 
     foreach (auto message, messages) {
         // Handle each separated message independently, ignoring any carriage returns
-        while (message.at(message.size() - 1) == '\r')
-            message.chop(1);
+        message = message.section('\r', 0, -1, QString::SectionSkipEmpty);
 #ifdef HAVE_QCA2
 […]
     }
```

Aside, it might be worthwhile following the use of [`QChar`](https://doc.qt.io/qt-5/qchar.html#SpecialCharacter-enum ), e.g.
```cpp
#if QT_VERSION < 0x050000
// QChar::LineFeed and ::CarriageReturn is Qt 5
static const QChar QCharLF = QChar('\n');
static const QChar QCharCR = QChar('\r');
#else
static const QChar QCharLF = QChar::LineFeed;
static const QChar QCharCR = QChar::CarriageReturn;
#endif
```

With the goal of switching to `QChar` directly after Qt 4 support is dropped.

kinnison: Thanks for the review, I'm shocked I didn't manage to try with a blank line.

My vote is for switching the loop to section, but to not skipemptyparts because that'd also lose the blanks which might be important in a paste.

Should I do that and update the PR, or will someone else do it, and close this instead?  If I update the PR, I can also add the QCharCR bit too.

digitalcircuit: Ah, good point!  (*This is why I shouldn't try to review code before going to class...*)

However, Freenode complains with `* No text to send` when I try to send a blank message, e.g. by sending `DEBUGTEST` with the following code in the Quassel core handler.

```diff
     QStringList messages = msg.split(QCharLF);
 
+    if(msg == "DEBUGTEST") {
+         messages = QString("DEBUG\nTEST\n\nNew").split(QCharLF);
+    }

     foreach (auto message, messages) {
```

This is impossible to do with normal Quassel.  The desktop client **already removes empty lines** in both [multi-line mode](https://github.com/quassel/quassel/blob/b8e812857ca64834078e1f620bf98fffa48b7768/src/uisupport/multilineedit.cpp#L703-L712 ) ([now default for 0.13](https://github.com/quassel/quassel/pull/253 )) and [single-line mode](https://github.com/quassel/quassel/blob/b8e812857ca64834078e1f620bf98fffa48b7768/src/uisupport/multilineedit.cpp#L729-L730 ).

Single-line mode actually already uses `.split('\n', QString::SkipEmptyParts)` if paste protection is enabled, or [removes all new lines if paste protection is disabled](https://github.com/quassel/quassel/blob/b8e812857ca64834078e1f620bf98fffa48b7768/src/uisupport/multilineedit.cpp#L727-L728 ).

Given servers shouldn't trust validation logic on the client, and on at least some networks it can cause unexpected errors, I think we should still switch to `SkipEmptyParts`.

Now, do any clients actually generate `\r` in messages?  It seems I can force one in by manually typing it, but everywhere else in Quassel, it's automatically converted `\r\n` to `\n`.

This seems to be [Qt behavior as per `qtbase.git/src/gui/text/qtextcursor.cpp`](https://code.qt.io/cgit/qt/qtbase.git/tree/src/gui/text/qtextcursor.cpp?id=c48f4bde0044bd5b23af231f3639d0779ecbdb03#n1448 ):
```cpp
            if (ch == QLatin1Char('\r')
                && (i + 1) < text.length()
                && text.at(i + 1) == QLatin1Char('\n')) {
                ++i;
                ch = text.at(i);
            }
```

I vote we should just entirely remove carriage return handling.  The only source would be third-party clients, Quasseldroid doesn't do this, and ideally no other clients would do it either.  Misinformed idea on my part to introduce it in the first place :)

In short, I vote we **switch to `SkipEmptyParts` and remove all carriage return handling.**

```diff
     // Split apart messages at line feeds.  The IRC protocol uses those to separate commands, so
     // they need to be split into multiple messages.
-    QStringList messages = msg.split(QCharLF);
+    QStringList messages = msg.split(QCharLF, QString::SkipEmptyParts);
 
     foreach (auto message, messages) {
         // Handle each separated message independently, ignoring any carriage returns
-        while (message.at(message.size() - 1) == '\r')
-            message.chop(1);
 #ifdef HAVE_QCA2
 […]
     }

```

digitalcircuit: And.. yes, you can go ahead and do this.  Just force-push an edited commit atop your current one :)

*Apologies for the double-message, I accidentally gave the "Comment" button focus before hitting `Enter` for a new line...*

kinnison:
@digitalcircuit was that update (faf7b7e) more suitable?

src/core/coreuserinputhandler.cpp
```diff
@@ -457,11 +457,10 @@ void CoreUserInputHandler::handleMe(const BufferInfo &bufferInfo, const QString
 
     // Split apart messages at line feeds.  The IRC protocol uses those to separate commands, so
     // they need to be split into multiple messages.
-    QStringList messages = msg.split(QCharLF);
+    QStringList messages = msg.split(QCharLF, QString::SkipEmptyParts);
```
digitalcircuit: *Undo this change, restoring the original…*
```diff
-    QStringList messages = msg.split(QCharLF, QString::SkipEmptyParts);
+    QStringList messages = msg.split(QCharLF);
```

*But keep the removal of `.trimmed()` below.*

src/core/coreuserinputhandler.cpp
```diff
@@ -529,11 +528,10 @@ void CoreUserInputHandler::handleNotice(const BufferInfo &bufferInfo, const QStr
     QList<QByteArray> params;
     // Split apart messages at line feeds.  The IRC protocol uses those to separate commands, so
     // they need to be split into multiple messages.
-    QStringList messages = msg.section(' ', 1).split(QCharLF);
+    QStringList messages = msg.section(' ', 1).split(QCharLF, QString::SkipEmptyParts);
```
digitalcircuit: *Undo this change, restoring the original…*
```diff
-    QStringList messages = msg.section(' ', 1).split(QCharLF, QString::SkipEmptyParts);
+    QStringList messages = msg.section(' ', 1).split(QCharLF);
```

*But keep the removal of `.trimmed()` below.*

src/core/coreuserinputhandler.cpp
```diff
@@ -604,11 +602,10 @@ void CoreUserInputHandler::handleQuery(const BufferInfo &bufferInfo, const QStri
     QString target = msg.section(' ', 0, 0);
     // Split apart messages at line feeds.  The IRC protocol uses those to separate commands, so
     // they need to be split into multiple messages.
-    QStringList messages = msg.section(' ', 1).split(QCharLF);
+    QStringList messages = msg.section(' ', 1).split(QCharLF, QString::SkipEmptyParts);
```
digitalcircuit: *Undo this change, restoring the original…*
```diff
-    QStringList messages = msg.section(' ', 1).split(QCharLF, QString::SkipEmptyParts);
+    QStringList messages = msg.section(' ', 1).split(QCharLF);
```

*But keep the removal of `.trimmed()` below.*

src/core/coreuserinputhandler.cpp
```diff
@@ -656,11 +653,10 @@ void CoreUserInputHandler::handleSay(const BufferInfo &bufferInfo, const QString
 
     // Split apart messages at line feeds.  The IRC protocol uses those to separate commands, so
     // they need to be split into multiple messages.
-    QStringList messages = msg.split(QCharLF);
+    QStringList messages = msg.split(QCharLF, QString::SkipEmptyParts);
```
digitalcircuit: ***Keep this change**, and keep the removal of `.trimmed()` below.*

src/core/coreuserinputhandler.cpp
```diff
@@ -656,11 +653,10 @@ void CoreUserInputHandler::handleSay(const BufferInfo &bufferInfo, const QString
...
 
     foreach (auto message, messages) {
-        // Handle each separated message independently, ignoring any carriage returns
-        message = message.trimmed();
+        // Handle each separated message independently
 #ifdef HAVE_QCA2
         putPrivmsg(bufferInfo.bufferName(), message, encodeFunc,
                    network()->cipher(bufferInfo.bufferName()));
```
digitalcircuit: For completeness sake, the final diff from `master` (on my side) looks something like...
```diff
@@ -460,8 +460,7 @@ void CoreUserInputHandler::handleMe(const BufferInfo &bufferInfo, const QString
     QStringList messages = msg.split(QCharLF);
 
     foreach (auto message, messages) {
-        // Handle each separated message independently, ignoring any carriage returns
-        message = message.trimmed();
+        // Handle each separated message independently
         coreNetwork()->coreSession()->ctcpParser()->query(coreNetwork(), bufferInfo.bufferName(),
                                                           "ACTION", message);
         emit displayMsg(Message::Action, bufferInfo.type(), bufferInfo.bufferName(), message,
@@ -532,8 +531,7 @@ void CoreUserInputHandler::handleNotice(const BufferInfo &bufferInfo, const QStr
     QStringList messages = msg.section(' ', 1).split(QCharLF);
 
     foreach (auto message, messages) {
-        // Handle each separated message independently, ignoring any carriage returns
-        message = message.trimmed();
+        // Handle each separated message independently
         params.clear();
         params << serverEncode(bufferName) << channelEncode(bufferInfo.bufferName(), message);
         emit putCmd("NOTICE", params);
@@ -607,8 +605,7 @@ void CoreUserInputHandler::handleQuery(const BufferInfo &bufferInfo, const QStri
     QStringList messages = msg.section(' ', 1).split(QCharLF);
 
     foreach (auto message, messages) {
-        // Handle each separated message independently, ignoring any carriage returns
-        message = message.trimmed();
+        // Handle each separated message independently
         if (message.isEmpty()) {
             emit displayMsg(Message::Server, BufferInfo::QueryBuffer, target,
                             tr("Starting query with %1").arg(target), network()->myNick(),
@@ -656,11 +653,10 @@ void CoreUserInputHandler::handleSay(const BufferInfo &bufferInfo, const QString
 
     // Split apart messages at line feeds.  The IRC protocol uses those to separate commands, so
     // they need to be split into multiple messages.
-    QStringList messages = msg.split(QCharLF);
+    QStringList messages = msg.split(QCharLF, QString::SkipEmptyParts);
 
     foreach (auto message, messages) {
-        // Handle each separated message independently, ignoring any carriage returns
-        message = message.trimmed();
+        // Handle each separated message independently
 #ifdef HAVE_QCA2
         putPrivmsg(bufferInfo.bufferName(), message, encodeFunc,
                    network()->cipher(bufferInfo.bufferName()));
```

kinnison:
I apologise for the delay, but I believe I've pushed the correct delta.  Thank you very much for working with me on this, it has really impressed me and encouraged me wrt. contributing in the future.

Hopefully this one will turn out to be right :-)

src/core/coreuserinputhandler.cpp
digitalcircuit: **This code is now correct, ignore previous comments on it.**

I'm not sure if I made a mistake or Github did something weird with my review; the previous comment shouldn't be here.

kinnison:
Hi, this change has been idle for nearly 3 weeks, just wondering if anyone is around who might be able to merge it (or tell me why it can't be merged) ?

digitalcircuit:
@kinnison There's likely nothing wrong with your change, just limited time from the project maintainers.  Usually they go in order from oldest pull requests to newest, asides from major changes that other pull requests will depend on, alongside fixing other things beyond the capacity of most contributors (*most recently `UiStyle` changes, currently `Feature` flag future-proofing*).

If this regression fix doesn't get in near time for 0.13 beta, I'll poke @Sput42 again :)

And apologies for the long delay!  Someone replied to this question in Freenode/`#quassel`, and I forgot to follow up here too.

PR #332, opened 2018-02-14
Persist Blowfish keys in the database

mamarley:
Add a new text field called "cipher" and store the hex-encoded
cipher key there whenever a new key is set or exchanged.  Also,
when each network is initialized, load the ciphers out of the
database and initialize the in-memory hashmap.  Then, the existing
behavior of each CoreIrcNetwork automatically using these keys
upon construction occurs.

Additionally, this makes PM buffer ciphers persistent both across
destruction/construction and across core restarts.

Note that the existing "key" field in the database is confusingly
named.  It does not contain any sort of cryptographic key but
instead holds channel passwords.

Closes #1473

### DIFF
--- a/src/core/SQL/PostgreSQL/select_buffer_ciphers.sql
+++ b/src/core/SQL/PostgreSQL/select_buffer_ciphers.sql
@@ -1,0 +1,3 @@
+SELECT buffername, cipher
+FROM buffer
+WHERE userid = :userid AND networkid = :networkid AND buffertype IN (2, 4)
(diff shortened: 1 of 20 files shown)


AlD:
Is there any specific advantage to storing this as a blob instead of some ascii representation?

mamarley:
I stored it as a blob because Quassel already stores it as a QByteArray internally.  In all the cases I have seen, this ends up being an ASCII string, but I couldn't guarantee that and I didn't want to risk encoding issues potentially corrupting the key.

AlD:
Fair point, but I think it makes sense to persist this as base64 or something. I think this will reduce hassle on DB migrations, etc.

mamarley:
Fixed.  It now uses Base64-encoded data and TEXT fields.

getpanoramic:
Sorry to drop in, but any update on the status of this Pull Request? This is something I'm really missing from quassel.

Sput42:
@digitalcircuit If you could do your testing magic here too with a focus on database changes, that would be swell :)

mamarley:
I am closing this since it has been integrated into https://github.com/quassel/quassel/pull/357.

PR #333, opened 2018-02-18
Store highlight status per buffer coreside

justjanne:
## In Short
* Add highlightCount to buffersyncer, same as with buffer activities
* Sync this with clients

## Status
- [x] Implement database backend
- [x] Integrate the highlight count tracking into the buffersyncer
- [x] Implement the actual sync of the activity between core and client
- [x] Have the client make use of this feature

## Rationale
Currently a client needs to request backlog for all channels, to know if events have occured on that channel (and potentially unhide it, show it differently in the UI, etc). This patch aims at removing that requirement by allowing the client to know the number of unseen highlights that have occured in a channel.

## Disadvantages
The core can not take ignore list items into account, so under some circumstances a channel might be marked as having new highlights, but the client might not show them once the channel is activated. Additionally, this can not process local highlights.

### DIFF
--- a/src/client/client.cpp
+++ b/src/client/client.cpp
@@ -394,6 +394,7 @@ void Client::setSyncedToCore()
     connect(bufferSyncer(), SIGNAL(buffersPermanentlyMerged(BufferId, BufferId)), _messageModel, SLOT(buffersPermanentlyMerged(BufferId, BufferId)));
     connect(bufferSyncer(), SIGNAL(bufferMarkedAsRead(BufferId)), SIGNAL(bufferMarkedAsRead(BufferId)));
     connect(bufferSyncer(), SIGNAL(bufferActivityChanged(BufferId, const Message::Types)), _networkModel, SLOT(bufferActivityChanged(BufferId, const Message::Types)));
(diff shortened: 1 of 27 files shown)


justjanne:
Simple way to test this:

1. Join a channel.
2. Ping yourself from another account until the channel is shown in orange
3. Set backlog fetching to 1
4. Send 2 or more messages that will not trigger a highlight
5. Reconnect
6. See the channel being marked in orange, while having no highlighted messages in the loaded backlog.

Nevcairiel:
How does marking such highlights as read work? If I have multiple highlights, will all be cleared when I open the buffer and scroll up a bit to find the first one? That wouldn't be ideal.

In fact, best would be if the client can know how much backlog it needs to fetch to get all outstanding highlights, and do this on the first manual backlog request.

justjanne:
@Nevcairiel That depends on what the client sends. Currently, the client indicates that it has seen all messages up to a point (the lastSeenMsgId), and the core will then clear all highlights.

Later on, we can obviously improve on that – but I need this feature, as minimalistic as it is now, in 0.13, because I want to be able to see all highlights in the buffer list, without loading backlog (very important for QuasselDroid).

AlD:
FWIW, the client supports fetching all messages up to the lastSeendMsgId via shortcut. This can of course be more than the earliest unseen highlight.

PR #335, opened 2018-02-27
Change HTTP links to HTTPS for quassel-irc.org and subdomains

phuzion:
I've updated any links to quassel-irc.org and any subdomains I could find from HTTP to HTTPS.


digitalcircuit:
I'm so happy this day has come :)

PR #336, opened 2018-02-28
common: Polish core-side highlights

digitalcircuit:
## In short

* Unify highlight and ignore channel matching
  * Add `!` syntax to ignore list scope matching
  * Switch highlight rule scope matching to same as ignore list
  * Make `RegEx`, `CS` also apply to `Sender` and `Channel`
  * Breaks existing advanced highlight rules, but local rules are **migrated on upgrade**
  * Almost no functionality lost, see examples
* Highlight current nick by default on upgrade/new configuration
  * For remote highlights, specify `CurrentNick` as default
  * For local highlights, specify `NoNick` as default
* Fix up importing local highlights
  * Prompt before importing, show success afterwards
  * Fix import button not working
* Explain highlights and Local vs. Remote highlights
  * Add tool-tips to the highlight/ignore list box and other elements
  * Indicate 'local' is for this device only, encourage monolithic to use remote
  * Warn when remote highlights aren't supported
* Align list boxes when switching between `Highlight Rules` and `Highlight Ignore Rules`
  * Add a `QGroupBox` to `Highlight Ignore Rules`, labeled `Never Highlight For`
* Disable `Case sensitive` checkbox when highlight nicks is `None`
  * Reduces confusion as the checkbox does nothing if nicks aren't highlighted

Criteria | Rank | Reason
Impact | ★★★ *3/3* | User-facing polish, consistency, fixes bugs
Risk | ★★★ *3/3* | Changes highlight rule matching behavior, possible confusion
Intrusiveness | ★★☆ *2/3* | UI, string, core changes, may interfere with other PRs

*Depends on [pull request #317](https://github.com/quassel/quassel/pull/317 ) being merged.*

# Rationale
Quassel provides two different locations to reference channel names (technically buffer names), `Ignore List` and `Highlights` (both remote and local).  Before now, Quassel treated channel names differently according to where you configured them.

Instead, Quassel should unify the handling of channel names to reduce confusion.  In order to still provide regular expression functionality, the `RegEx` option should also apply to the channel column.  This will break any existing advanced highlight rules, but it simplifies the easier cases and no functionality is lost.

## Breaking changes
* `Channel` and `Sender` no longer use regular expressions by default
  * Change `Channel` and `Sender` to the simpler wildcard format
  * Check the `RegEx` box, modify `Rule` to also be a regular expression
  * **Existing rules automatically converted to `RegEx` format** if `Channel` is specified
* Case sensitivity now applies to `Channel` and `Sender`, not just `Rule`
  * Un/check the `CS` box to get all case-sensitive or all case-insensitive
  * **No longer possible to have a case-sensitive rule and case-insensitive `Channel` or `Sender`**

# Examples
### Unified Model (this pull request)
* Advantages
  * Makes common cases easy, already learned behavior
  * Shared logic between ignore list and highlight rules
* Disadvantages
  * No way to escape `;` or `!` in non-`RegEx` rules
  * Use of regular expressions tied to `RegEx`, requiring `Rule` to also be regex or not
  * Breaks existing highlight rule channel names, but they are automatically migrated
```
#quassel*; #ircv3; !#quasseldroid
al!*; Sput!*@quassel/developer/sput; Carol*!*; !Caroline!*
```
***Channel:*** *matches `#quassel...`, `#ircv3`, but not `#quasseldroid`*
***Sender:*** *matches `al`, `Sput` with hostmask `quassel/developer/sput`, and `Carol...` but not `Caroline`*
```
!#quassel*; !#ircv3
!Announce*!*; !Wheatley!aperture@*
```
***Channel:*** *matches anything except for `#quassel...` or `#ircv3` (implicit wildcard)*
***Sender:*** *matches anything except for `Announce...` or `Wheatley` with ident `aperture` (implicit wildcard)*

### Ignore List (before)
* Advantages
  * Simpler to manage common cases
  * Already learned behavior
* Disadvantages
  * No invert-match `!`
  * No advanced regular expressions
  * No way to escape `;` in rules
```
#quassel*; #ircv3
```

### Highlight Rules (before)
*Includes highlight ignore rules and local highlights*
```
(#quassel.*|#ircv3)
!#quasseldroid
```

### Local Highlights
*In Settings → Interface → Local Highlights, information bar at bottom and "Details…" dialog*
Settings - Local Highlights - Details dialog, showing Local Highlight vs. Remote Highlights
> *:information_source: Local Highlights apply to this device only*
> 
> Dialog:
> **Local highlights apply to this device only**
> Highlights configured on this page only apply to your current device.
> Configure highlights for all of your devices in *Remote Highlights*.

### Local Highlights (Monolithic)
**For the monolithic build only**
*In Settings → Interface → Local Highlights, information bar at bottom and "Details…" dialog*
Settings - Local Highlights - Details dialog, showing Local Highlight vs. Remote Highlights
> *:information_source: Local Highlights are replaced by Remote Highlights*
> 
> Dialog:
> **Local Highlights are replaced by Remote Highlights**
> These highlights will keep working for now, but you should move to the improved highlight rules when you can.
> Configure the new style of highlights in *Remote Highlights*.

### Remote Highlights unsupported
*In Settings → Interface → Remote Highlights, warning bar at top and "Details…" dialog*
Settings - Remote Highlights - Details dialog, showing when Remote Highlights aren't supported
> *:warning: Your Quassel core is too old to support remote highlights*
> 
> Dialog:
> **Your Quassel core is too old to support remote highlights**
> You need a Quassel core v0.13.0 or newer to configure remote highlights.
> You can still configure highlights for this device only in *Local Highlights*.

### Remote Highlights importing Local
*In Settings → Interface → Remote Highlights, clicking the Import Local button*
Settings - Remote Highlights - Import Local dialogs, showing the stages of an import

*Asking to import*
> Import all highlight rules from *Local Highlights*?
> No | Yes

*Import done*
> `number` highlight rules successfully imported.

*Nothing to import*
> No highlight rules in *Local Highlights*.

### Highlight Rules tooltips
Settings - Remote Highlights - showing different tooltips

> * Enabled
>   * Enable/disable this rule
> * Rule
>   * Phrase to match
> * RegEx
>   * **RegEx**: This option determines if the highlight rule, *Sender*, and *Channel* should be interpreted as a **regular expressions** or just as keywords.
> * CS
>   * **CS**: This option determines if the highlight rule, *Sender*, and *Channel* should be interpreted as **case sensitive**.
> * Import Local (button at bottom)
>   * Import highlight rules configured in *Local Highlights*

Settings - Remote Highlights - showing sender tooltip
*This one is long and shares most of the content from the channel tooltip below*

> **Sender**: Semicolon separated list of *nick!ident@host* names, leave blank to match any nickname.
> 
> *Example:*
> *Alice!\*; Bob!\*@example.com; Carol\*!\*; !Caroline!\**
> would match on *Alice*, *Bob* with hostmask *example.com*, and any nickname starting with *Carol* except for *Caroline*
> 
> If only inverted names are specified, it will match anything except for what's specified (implicit wildcard).
> 
> *Example:*
> *!Announce\*!\*; !Wheatley!aperture@\**
> would match anything except for *Wheatley* with ident *aperture* or any nickname starting with *Announce*

Settings - Remote Highlights - showing channel tooltip
*This one is long and shares most of the content from the updated tooltip in the Ignore List window*

> **Channel**: Semicolon separated list of channel names, leave blank to match any name.
> 
> *Example:*
> *#quassel\*; #foobar; !#quasseldroid*
> would match on #foobar and any channel starting with *#quassel* except for *#quasseldroid*
> 
> If only inverted names are specified, it will match anything except for what's specified (implicit wildcard).
> 
> *Example:*
> *!#quassel\*; !#foobar*
> would match anything except for #foobar or any channel starting with *#quassel*

---

**To be done in a later pull request/future time:**
* Switch to a view model for highlight rules
  * Requires a `highlightrulemodel.cpp`, mimicking [`aliasesmodel.cpp`](https://github.com/quassel/quassel/blob/934fa4151f48fd12549fe0772cfbc995c6e63034/src/qtui/settingspages/aliasesmodel.cpp#L54-L74 ) and [`ignorelistmodel.cpp`](https://github.com/quassel/quassel/blob/934fa4151f48fd12549fe0772cfbc995c6e63034/src/qtui/settingspages/ignorelistmodel.cpp#L55-L86 )
* Hide `Local Highlights` page behind a `Manage local rules` button
  * Make `Local Highlights` into a dialog-able window, mimicking ignores
  * Move `Import` button into dialog behind `Manage local rules` button
  * Avoids confusing new users
  * Allows power-users to access old functionality
  * *Better naming?*

### DIFF
--- a/src/client/clientignorelistmanager.cpp
+++ b/src/client/clientignorelistmanager.cpp
@@ -50,8 +50,8 @@ QMap<QString, bool> ClientIgnoreListManager::matchingRulesForHostmask(const QStr
     QMap<QString, bool> result;
     foreach(IgnoreListItem item, ignoreList()) {
         if (item.type == SenderIgnore && pureMatch(item, hostmask)
-            && ((network.isEmpty() && channel.isEmpty()) || item.scope == GlobalScope || (item.scope == NetworkScope && scopeMatch(item.scopeRule, network))
-                || (item.scope == ChannelScope && scopeMatch(item.scopeRule, channel)))) {
+            && ((network.isEmpty() && channel.isEmpty()) || item.scope == GlobalScope || (item.scope == NetworkScope && scopeMatch(network, item.scopeRule))
+                || (item.scope == ChannelScope && scopeMatch(channel, item.scopeRule)))) {
             result[item.ignoreRule] = item.isActive;
 //      qDebug() << "matchingRulesForHostmask found: " << item.ignoreRule << "is active: " << item.isActive;
         }
--- a/src/client/clientsettings.cpp
+++ b/src/client/clientsettings.cpp
@@ -320,7 +320,7 @@ void NotificationSettings::setHighlightNick(NotificationSettings::HighlightNickT
 
 NotificationSettings::HighlightNickType NotificationSettings::highlightNick()
 {
-    return (NotificationSettings::HighlightNickType)localValue("Highlights/HighlightNick", CurrentNick).toInt();
+    return (NotificationSettings::HighlightNickType)localValue("Highlights/HighlightNick", NoNick).toInt();
 }
 
 
--- a/src/common/highlightrulemanager.cpp
+++ b/src/common/highlightrulemanager.cpp
@@ -145,7 +145,8 @@ bool HighlightRuleManager::match(const QString &msgContents,
         if (!rule.isEnabled)
             continue;
 
-        if (!rule.chanName.isEmpty() && !scopeMatch(rule.chanName, bufferName)) {
+        if (!rule.chanName.isEmpty()
+                && !scopeMatch(bufferName, rule.chanName, rule.isRegEx, rule.isCaseSensitive)) {
             // A channel name rule is specified and does NOT match the current buffer name, skip
             // this rule
             continue;
@@ -163,12 +164,8 @@ bool HighlightRuleManager::match(const QString &msgContents,
         if (rule.sender.isEmpty()) {
             senderMatch = true;
         } else {
-            if (rule.isRegEx) {
-                rx = QRegExp(rule.sender, rule.isCaseSensitive ? Qt::CaseSensitive : Qt::CaseInsensitive);
-            } else {
-                rx = QRegExp(rule.sender, Qt::CaseInsensitive, QRegExp::Wildcard);
-            }
-            senderMatch = rx.exactMatch(msgSender);
+            // A sender name rule is specified, match according to scope rules.
+            senderMatch = scopeMatch(msgSender, rule.sender, rule.isRegEx, rule.isCaseSensitive);
         }
 
         if (nameMatch && senderMatch) {
--- a/src/common/highlightrulemanager.cpp
+++ b/src/common/highlightrulemanager.cpp
@@ -141,17 +141,10 @@ bool HighlightRuleManager::match(const QString &msgContents,
         if (!rule.isEnabled)
             continue;
 
-        if (rule.chanName.size() > 0 && rule.chanName.compare(".*") != 0) {
-            if (rule.chanName.startsWith("!")) {
-                QRegExp rx(rule.chanName.mid(1), Qt::CaseInsensitive);
-                if (rx.exactMatch(bufferName))
-                    continue;
-            }
-            else {
-                QRegExp rx(rule.chanName, Qt::CaseInsensitive);
-                if (!rx.exactMatch(bufferName))
-                    continue;
-            }
+        if (!rule.chanName.isEmpty() && !scopeMatch(rule.chanName, bufferName)) {
+            // A channel name rule is specified and does NOT match the current buffer name, skip
+            // this rule
+            continue;
         }
 
         QRegExp rx;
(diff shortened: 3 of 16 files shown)


digitalcircuit:
## 2018-3-5 update

Both `Sender` and `Channel` now make use the `scopeMatch` function, allowing you to specify sender `nick!ident@host` matches using the simpler semicolon wildcard form.

`RegEx` (regular expression) and `CS` (case sensitivity) now apply to `Sender` and `Channel`, not just `Rule`.  This restores the ability to use regular expressions, but at the cost of needing to adapt `Rule` into regular expression too.

Added `Rationale` with a section on `Breaking changes` to clarify changes for review and to explain how to upgrade existing highlight rules.

Updated tooltips.

digitalcircuit:
## 2018-5-13 update

Existing local highlight rules that specify `Channel` are now automatically migrated to `RegEx` format, making sure **most local highlight rules keep working on upgrade**.  Rules that don't specify a `Channel` field are kept as non-`RegEx` since they'll keep working either way.

This remains compatible even when downgrading the client afterwards.

Core-side highlight rules need manually migrated.  They're harder to detect, and core-side highlights haven't yet been released, anyways.

The matter of case-sensitivity applying to the entire rule still exists.

digitalcircuit:
## 2018-5-28 update

Fixed a bug where inverted `RegEx` sender/channel rules would never match, i.e. when the sender/channel did **not** match the rule, it failed to return true to treat it as a match.

This also simplified the logic a tiny bit.  Yay.

PR #338, opened 2018-03-02
common: Polish core session UI

digitalcircuit:
## In short

* Always show `End Session`, set UI and tooltip according to status
  * Avoids confusion over why buttons sometimes visible
  * Points to outdated core or client as appropriate
* Automatically refresh core session UI on client join/leave
  * Reduces confusion when disconnecting clients
  * Matches expectations set by the ticking "uptime" timer
* Fix crash when repeatedly attempting to disconnect a peer
  * Prevent `peerById` from inserting invalid IDs, which causes crashes elsewhere

*Many thanks to @mamarley and @justjanne for handling the core session synchronization, and for fixing/finding the `peerById` crash bug.*

Criteria | Rank | Reason
Impact | ★★★ *3/3* | User-facing polish, less confusion over features, core crash fix
Risk | ★★☆ *2/3* | Session UI might mislead, unexpected protocol side-effects
Intrusiveness | ★★☆ *2/3* | Protocol changes, could interfere with other PRs

### Explanation tooltips
*In File → Core Info…*
File - Core Info, showing two sessions, one that can be ended, another that cannot
*Regular tooltip*
> End the client's session, disconnecting it

*Client lacks `RemoteDisconnect` feature flag*
> End the client's session, disconnecting it
> 
> **This client does not support being remotely disconnected**

File - Core Info, showing an outdated core that's not able to list any sessions
*Core does not transmit the expected `sessionConnectedClientData`*
> *:warning: Your Quassel core is too old to show active sessions*
> 
> Dialog:
> **Your Quassel core is too old to show active sessions**
> You need a Quassel core v0.13.0 or newer to view and
> disconnect other connected clients.

### End Session reaction
*In File → Core Info…, clicking `End Session` on a session*
File - Core Info, after clicking 'End Session', button relabeled to 'Ending session...'
*`End Session` button is disabled and relabeled as `Ending session...`*

**NOTE:**  On a fast connection, this message won't be shown.  However, it's still useful as immediate feedback for slow connections, and for disconnecting the active session.


digitalcircuit:
I'm having trouble figuring out where to add new signals/slots in order to detect when clients join/part, and to update the "Connected Clients" counter.  If anyone can point me in the right direction, I'd appreciate it!

**Rationale:**  At least to me, it's not at all obvious that you have to close and re-open that dialog to see any changes.  The very second thing I tried when this feature landed was to join a client and wait for it to show up in the dialog, especially since the `Uptime` keeps ticking (*generated by timer*).. and I'd rather not have anyone else filing bugs on this, too :)

justjanne:
I appreciate your enthusiams, but...

> Core lacks RemoteDisconnect feature flag

> **End the client's session, disconnecting it**
> Your Quassel core does not support this feature
> You need a Quassel core v0.13.0 or newer in order to end connected sessions.

This message is unnecessary and will never be shown (so we don’t need to add or translate it).

Only since 0.13 does the CoreInfo data contain the list of currently active sessions anyway

digitalcircuit:
@justjanne Good point, thank you for catching this!  I entirely overlooked that transmitting the active list of core sessions is a 0.13 feature.

Instead, I should handle an empty list of active sessions when connected client count > 0, and show a message in place of the active session list (akin to remote highlights, but information rather than warning).  I'll add that to the pull request TODO list.

digitalcircuit:
## 2018-5-8 update
Core unsupported message has been fixed to show for older cores, thanks to @justjanne's remark.

Core session list now dynamically updates, also thanks to @justjanne and to @mamarley.

PR #339, opened 2018-03-02
[Requires #340, #351] Support core-side filtering of backlog by type

justjanne:
## In Short
* Introduces new sync calls into the backlogmanager:
  * `requestBacklogFiltered(BufferId bufferId, MsgId first, MsgId last, int limit, int additional, int type)`
  * `requestBacklogAllFiltered(MsgId first, MsgId last, int limit, int additional, int type)`
* Introduces a new version of each backlog loading query, which filters by type (with `backlog.type & :type != 0`)

## Status
- [x] Implement a new sync call
- [x] Implement SQL to support this call
- [ ] Make use of this call in the desktop client, optionally.

## Rationale
Currently when loading backlog with lots of types filtered in a channel with many joins/quits but little talk we need to load all the messages that will be filtered anyway. This is acceptable on desktop, but especially on mobile this leads to users being confused, or assuming the client is broken, when the client loads for many minutes without being able to show any loaded messages (as they are all filtered).

For this use case it is useful to support filtering messages on the core.

## Impact
All SQL changes are read-only, no schema upgrade is required. All changes are backwards compatible. No UI or client changes.

### DIFF
--- a/src/core/postgresqlstorage.cpp
+++ b/src/core/postgresqlstorage.cpp
@@ -1449,7 +1449,11 @@ bool PostgreSqlStorage::logMessage(Message &msg)
         return false;
     }
 
-    QSqlQuery getSenderIdQuery = executePreparedQuery("select_senderid", msg.sender(), db);
+    QVariantList senderParams;
+    senderParams << msg.sender()
+                 << msg.realName()
+                 << msg.avatarUrl();
+    QSqlQuery getSenderIdQuery = executePreparedQuery("select_senderid", senderParams, db);
     int senderId;
     if (getSenderIdQuery.first()) {
         senderId = getSenderIdQuery.value(0).toInt();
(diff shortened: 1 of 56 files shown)


romibi:
QFlags does not support unsigned enums under MSVC
See: https://code.woboq.org/qt5/qtbase/src/corelib/global/qflags.h.html#61 and line 106
This compiles (but probably is suboptimal): https://github.com/romibi/quassel/commit/bc2f5511da27466b2939a5f7c4a4a3b048bf4ea7
See here: https://ci.appveyor.com/project/romibi/quassel/build/280
(Note: I did not wait for it to compile on appveyor and went to bed)

src/core/postgresqlstorage.cpp
```diff
@@ -1693,7 +1752,64 @@ QList<Message> PostgreSqlStorage::requestAllMsgs(UserId user, MsgId first, MsgId
...
+
+    safeExec(query);
+    if (!watchQuery(query)) {
+        db.rollback();
+        return messagelist;
+    }
+
+    QDateTime timestamp;
+    for (int i = 0; i < limit && query.next(); i++) {
+        timestamp = query.value(1).toDateTime();
```
mamarley: I think the incorrect index is used for the date value here; it looks like it should be 2.

justjanne:
This PR needs to be rewritten as soon as 351 is merged, as they’d have merge conflicts. Until then, this PR is on gold.

PR #340, opened 2018-03-03
Avoid Y2038 problem

justjanne:
## In Short
* Add new feature flag for 64-bit time
* Serialize timestamp of messages as milliseconds since epoch

## Rationale
Currently, quassel will run into the year 2038 problem within of the next 20 years, this patch avoids that issue. Additionally, it allows higher accuracy of message times, which is something users have requested.

## Impact
This includes a major protocol change, guarded behind a feature flag. No UI or core changes are included.

### DIFF
--- a/src/common/message.cpp
+++ b/src/common/message.cpp
@@ -53,9 +53,15 @@ Message::Message(const QDateTime &ts, const BufferInfo &bufferInfo, Type type, c
 QDataStream &operator<<(QDataStream &out, const Message &msg)
 {
     // We do not serialize the sender prefixes until we have a new protocol or client-features implemented
-    out << msg.msgId()
-        << (quint32) msg.timestamp().toTime_t()
-        << (quint32) msg.type()
+    out << msg.msgId();
+
+    if (SignalProxy::current()->targetPeer()->hasFeature(Quassel::Feature::LongMessageTime)) {
+        out << (quint64) msg.timestamp().toMSecsSinceEpoch();
+    } else {
+        out << (quint32) msg.timestamp().toTime_t();
+    }
+
+    out << (quint32) msg.type()
         << (quint8) msg.flags()
         << msg.bufferInfo()
         << msg.sender().toUtf8();
@@ -72,9 +78,15 @@ QDataStream &operator>>(QDataStream &in, Message &msg)
 {
     in >> msg._msgId;
 
-    quint32 timeStamp;
-    in >> timeStamp;
-    msg._timestamp = QDateTime::fromTime_t(timeStamp);
+    if (SignalProxy::current()->sourcePeer()->hasFeature(Quassel::Feature::LongMessageTime)) {
+        quint64 timeStamp;
+        in >> timeStamp;
+        msg._timestamp = QDateTime::fromMSecsSinceEpoch(timeStamp);
+    } else {
+        quint32 timeStamp;
+        in >> timeStamp;
+        msg._timestamp = QDateTime::fromTime_t(timeStamp);
+    }
 
     quint32 type;
     in >> type;
--- a/src/common/quassel.h
+++ b/src/common/quassel.h
@@ -127,6 +127,7 @@ public:
         SenderPrefixes,           ///< Show prefixes for senders in backlog
         RemoteDisconnect,         ///< Allow this peer to be remotely disconnected
         ExtendedFeatures,         ///< Extended features
+        LongMessageTime,          ///< Serialize message time as 64-bit
     };
     Q_ENUMS(Feature)
 


src/common/message.cpp
```diff
@@ -72,9 +77,15 @@ QDataStream &operator>>(QDataStream &in, Message &msg)
 {
     in >> msg._msgId;
 
-    quint32 timeStamp;
-    in >> timeStamp;
-    msg._timestamp = QDateTime::fromTime_t(timeStamp);
+    if (SignalProxy::current()->targetPeer()->features().testFlag(Quassel::Feature::LongMessageTime)) {
```
mamarley: I'm pretty sure this needs to be "if (SignalProxy::current()->sourcePeer()->features().testFlag(Quassel::Feature::LongMessageTime)) {".  As is, the core crashes whenever a client connects due to targetPeer() being null here.

src/core/postgresqlstorage.cpp
```diff
@@ -1632,9 +1632,7 @@ QList<Message> PostgreSqlStorage::requestMsgs(UserId user, BufferId bufferId, Ms
 
     QDateTime timestamp;
     while (query.next()) {
-        timestamp = query.value(1).toDateTime();
-        timestamp.setTimeSpec(Qt::UTC);
-        Message msg(timestamp,
+        Message msg(QDateTime::fromMSecsSinceEpoch(query.value(1).toLongLong() * 1000),
```
mamarley: This is incorrect.  Calling .toLongLong() on the value returned from PostgreSQL results in 0, causing all messages to appear as if they were sent at the UNIX epoch.  The existing code converts it directly from QVariant to QDateTime, which is 64-bit and does not lose precision or overflow.

src/core/postgresqlstorage.cpp
```diff
@@ -1685,9 +1683,7 @@ QList<Message> PostgreSqlStorage::requestAllMsgs(UserId user, MsgId first, MsgId
 
     QDateTime timestamp;
     for (int i = 0; i < limit && query.next(); i++) {
-        timestamp = query.value(1).toDateTime();
-        timestamp.setTimeSpec(Qt::UTC);
-        Message msg(timestamp,
+        Message msg(QDateTime::fromMSecsSinceEpoch(query.value(1).toLongLong() * 1000),
```
mamarley: Same here

src/core/sqlitestorage.cpp
```diff
@@ -1749,7 +1749,7 @@ QList<Message> SqliteStorage::requestMsgs(UserId user, BufferId bufferId, MsgId
         watchQuery(query);
 
         while (query.next()) {
-            Message msg(QDateTime::fromTime_t(query.value(1).toInt()),
+            Message msg(QDateTime::fromMSecsSinceEpoch(query.value(2).toLongLong() * 1000),
```
mamarley: It looks like the index here got accidentally changed from 1 to 2.

src/common/message.cpp
```diff
@@ -72,9 +77,15 @@ QDataStream &operator>>(QDataStream &in, Message &msg)
...
+        quint64 timeStamp;
+        in >> timeStamp;
+        msg._timestamp = QDateTime::fromMSecsSinceEpoch(timeStamp, Qt::TimeSpec::UTC);
```
mamarley: Qt4 doesn't have a fromMSecsSinceEpoch(qint64 msecs, const QTimeZone &timeZone) function, but I think you could do the same thing using .setTimeSpec().

mamarley: In fact, you probably want to do it that way anyway since according to the documentation, the function you used wasn't added until Qt5.8.

src/core/postgresqlstorage.cpp
justjanne: The existing code only works if we use seconds from epoch, I want to use milliseconds from epoch.

src/common/message.cpp
justjanne: That's intentional - this patch is designed for post-0.13, when we only have modern Qt5.

src/core/postgresqlstorage.cpp
mamarley: Fair enough, but what you have put there doesn't work at all.  It always returns the UNIX epoch.

mamarley: On further investigation, are you sure about this?  I built a core that prints "query.value(1).toDateTime().toMSecsSinceEpoch()" after retrieving each message and it prints such values as "1522214171310" and "1522214171311" that appear to be in milliseconds.

justjanne: Interesting, then I probably did something wrong during my early testing. Thanks.

src/common/message.cpp
Sput42: As a side remark: The Qt5 version we intend to target for 0.14 is Qt 5.5... due to availability in major distros.

PR #341, opened 2018-03-03
Improve containerization

justjanne:
## In Short
* Move CoreState / Active Sessions into the database
  * New database table "core_state"
  * Read state from quasselCore.conf and database
  * Write exclusively to database
* Add new CLI switch to enable environment configuration
* Add new environment variables to configure database and authenticator

## Rationale
In containerized and other automated improvements it’s usually common to configure software through environment variables or config files automatically. Quassel’s existing binary config format does not allow this.

## Impact
No changes for normal setup flow, new database table

## Changes
The CLI switch --config-from-environment enables loading of configuration from environment, unless select-backend or a similar option is given.  
There are global configuration options DB_BACKEND and AUTH_AUTHENTICATOR to select a database and authentication backend. Additionally, each backend has its own additional environment variables.

**PostgreSQL Backend**
 * `DB_PGSQL_USERNAME`
 * `DB_PGSQL_PASSWORD`
 * `DB_PGSQL_HOSTNAME`
 * `DB_PGSQL_PORT`
 * `DB_PGSQL_DATABASE`

**SQLite Backend**
No options

**LDAP Authenticator**
 * `AUTH_LDAP_HOSTNAME`
 * `AUTH_LDAP_PORT`
 * `AUTH_LDAP_BIND_DN`
 * `AUTH_LDAP_BIND_PASSWORD`
 * `AUTH_LDAP_BASE_DN`
 * `AUTH_LDAP_FILTER`
 * `AUTH_LDAP_UID_ATTRIBUTE`

**Database Authenticator**
No options

### DIFF
--- a/src/common/main.cpp
+++ b/src/common/main.cpp
@@ -162,6 +162,7 @@ int main(int argc, char **argv)
     cliParser->addSwitch("syslog", 0, "Log to syslog");
 #endif
     cliParser->addOption("logfile", 'l', "Log to a file", "path");
+    cliParser->addSwitch("config-from-environment", 0, "Load configuration from environment variables");
     cliParser->addOption("select-backend", 0, "Switch storage backend (migrating data if possible)", "backendidentifier");
     cliParser->addOption("select-authenticator", 0, "Select authentication backend", "authidentifier");
     cliParser->addSwitch("add-user", 0, "Starts an interactive session to add a new core user");
--- a/src/core/SQL/PostgreSQL/insert_core_state.sql
+++ b/src/core/SQL/PostgreSQL/insert_core_state.sql
@@ -1,0 +1,2 @@
+INSERT INTO core_state (key, value)
+VALUES (:key, :value)
--- a/src/core/SQL/PostgreSQL/migrate_write_corestate.sql
+++ b/src/core/SQL/PostgreSQL/migrate_write_corestate.sql
@@ -1,0 +1,2 @@
+INSERT INTO core_state (key, value)
+VALUES (?, ?)
--- a/src/core/SQL/PostgreSQL/select_core_state.sql
+++ b/src/core/SQL/PostgreSQL/select_core_state.sql
@@ -1,0 +1,3 @@
+SELECT value
+FROM core_state
+WHERE key = :key
--- a/src/core/SQL/PostgreSQL/setup_140_corestate.sql
+++ b/src/core/SQL/PostgreSQL/setup_140_corestate.sql
@@ -1,0 +1,5 @@
+CREATE TABLE core_state (
+    key TEXT NOT NULL,
+    value bytea,
+    PRIMARY KEY (key)
+)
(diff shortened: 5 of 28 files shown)


justjanne:
Should we give the options each a prefix of the type?

e.g. `AUTH_LDAP_HOSTNAME`, or `DB_PGSQL_PASSWORD`?

AlD:
Thanks, great initiative. I would like to encourage you to take it a step further even.

I would much prefer these to be command line flags than environment variables. Could you abstract this in a way that allows it to be integrated into cliParser? This would also make it usable for other things, besides storage.
It should support usage both via environment and flags. Flags should always win over environment.
IMHO we should get rid of the binary config, at least for these settings. Alternatively we could also introduce a new human readable config, but the binary core config should be deprecated sooner rather than later.

`--help` should show the flags, the corresponding environment variable, and the default value. If a flag / environment variable is already set, this should be reflected by the shown default value.

Very rough example:
```
cliParser->addOption("db-pgsql-username", "PostgreSQL user name", pg_db_username, environmentOrConfigFileOrDefault("DB_PGSQL_USERNAME", "defaultuser"));
```

Of course, the check can be integrated into cliParser directly. Environment variables and flags and config option keys could also be derived from the valuename (but allowing customization).

justjanne:
@AlD No. I will not implement these as flags, and there’s an important reason for this.

The entire command invocation, including flags, is stored, and available to be seen by all processes on the system – including potentially those in other containers.

Today these other processes see `/usr/local/bin/quasselcore --configdir=/var/lib/quassel --syslog --loglevel=Info --port=4242 --require-ssl`, but with your suggestion they’d see `/usr/local/bin/quasselcore --configdir=/var/lib/quassel --syslog --loglevel=Info --port=4242 --require-ssl --db-pgsql-username quassel --db-pgsql-password 3DTd4h37f3L7HjxwN3UBaYbWS9j9pwiJkTJMMWNtno76nQay7JcPTX6Ej563lAuiwQtI8SDSaAB8SzNpzFBTGg==`

AlD:
Thanks, that's a fair point, but it's not withstanding the fact that it would be better to have a single source of truth for (core) settings, that supports env variables, flags and a human readable config file.

This is just a suggestion, not a blocker from my side, though, so if you don't want to take this on, that's perfectly fine.

justjanne:
A single source of truth would be ideal, and we still need to do this, but that's a larger engineering effort.

But in either case, command line flags aren't going to be an option.

justjanne:
So, I looked into the code, and started rewriting this all to be nicer, and clearer, and quickly realized that it makes no sense to rewrite this part before 0.13.

    // Initialize CLI arguments
    // NOTE: We can't use tr() at this point, since app is not yet created
    // TODO: Change this once we get rid of KDE4 and can initialize the parser after creating the app

I’d prefer to get this PR merged before 0.13, though. So after 0.13 is out, I’ll have to rewrite this code to make it cleaner (and handle translation properly as well, as many other features we had to throw out now), but for now, doing so would be a wasted endeavor.

AlD:
Ok, thanks for looking into it.

mamarley:
The description says that the settings should be read both from the config file and the DB, but it seems that the configuration file is only read from the DB.  This causes the active sessions to be lost and the core to not reconnect to IRC automatically the first time it is started after this patch is applied.

PR #343, opened 2018-03-13
Use 64-bit IDs for messages

mamarley:
Increase the length of message IDs from 32 to 64 bits, allowing for
many more total messages be stored.  Because messages are synced to
the client, this requires a protocol change.  The protocol change is
guarded behind a new feature flag that will retain backwards
compatibility with existing clients, at least until the IDs go above
the maximum for a 32-bit number.  At this point, legacy clients will
behave incorrectly or perhaps crash, but this is an acceptable risk
because in the the current implementation, this event would cause
the core to behave incorrectly or crash.  The client is also still
limited to 32 bits worth of backlog per buffer displayed at once
(due to Qt limitations), but it would be highly unlikely to hit
this limit because the client becomes unusably slow with a small
fraction of that amount of backlog loaded in a single buffer.

For SQLite, no schema change was necessary as SQLite makes no
differentiation between the bitnesses of integers.

For PostgreSQL, this patch also changes senderId to 64-bit, but
this has no effect outside the PostgreSQL database driver.

This patch only changes the message ID to 64-bit.  All other IDs
are left at 32 bits because it is either exceedingly unlikely or
technically infeasible to have more than 32 bits worth of entries
in any of the other tables.  If, at some point, it is decided that
other IDs should also be 64-bit, the way in which the feature flag
affects (de)serialization will need to be changed.  This is because
the flag LongMsgId currently acts upon all IDs inheriting from
SignedId64 and it would be necessary to independently control the
(de)serialization of each ID if other IDs became 64-bit in order to
continue to maintain backwards compatibility.  One possible method
of doing this would be to subclass the SignedId64 class for each
ID that inherits from it to allow for separately controlling
(de)serialization for each of the types.  I have not implemented
this, however, because of the unlikelihood of any of the other IDs
ever being made 64-bit.

### DIFF
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -38,6 +38,7 @@ set(SOURCES
     syncableobject.cpp
     transfer.cpp
     transfermanager.cpp
+    types.cpp
     util.cpp
 
     serializers/serializers.cpp
--- a/src/common/quassel.h
+++ b/src/common/quassel.h
@@ -134,6 +134,7 @@ public:
 #if QT_VERSION >= 0x050500
         EcdsaCertfpKeys,          ///< ECDSA keys for CertFP in identities
 #endif
+        LongMessageId,            ///< 64-bit IDs for messages
     };
     Q_ENUMS(Feature)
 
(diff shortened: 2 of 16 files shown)


mamarley:
I am closing this since it has been integrated into https://github.com/quassel/quassel/pull/357.

PR #345, opened 2018-03-26
Represent core/client features as string list in the protocol

Sput42:
The previous way of representing all supported features as a bit-wise
combination of flags is reaching its limits, due to the fact that the
old Features enum didn't have a defined width, and thus the compiler
can (and does) choose to represent it as a 16 bit value even though
the serialization in the protocol is defined as a 32 bit value.

In order to solve this problem, and to make feature negotiation more
robust, a new field "FeatureList" is added to both "ClientInit"
and "ClientInitAck" handshake messages that holds a string list with
the supported features.

Clients still send both "Features" and "FeatureList" fields when
registering, in order to support older cores. The core now omits
the legacy "CoreFeatures" field if the client indicates support
for extended features. Support for extended features is indicated
as the legacy feature 0x8000.

### DIFF
--- a/src/client/client.cpp
+++ b/src/client/client.cpp
@@ -56,7 +56,6 @@
 #include <stdlib.h>
 
 QPointer<Client> Client::instanceptr = 0;
-Quassel::Features Client::_coreFeatures = 0;
 
 /*** Initialization/destruction ***/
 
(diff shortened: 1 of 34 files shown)


mamarley:
I have tested this and found that it works properly.  Thanks!

PR #348, opened 2018-04-03
Add strict-oidentd mode

mamarley:
This mode is disabled by default.  When specified, this restricts
authoratative ident responses to be the authuser account name.

This is a cut-down version of,
https://github.com/quassel/quassel/pull/164 which was originally
authored by @esainane.  @AlD requested that the database schema
change be removed, but the request fell dormant and was eventually
closed.  This version removes the schema change as requested,
always using the user's core username as the ident when enabled.

Co-authored-by: Sai Nane <esainane+github@gmail.com>
Co-authored-by: Michael Marley <michael@michaelmarley.com>

### DIFF
--- a/src/core/oidentdconfiggenerator.cpp
+++ b/src/core/oidentdconfiggenerator.cpp
@@ -18,10 +18,10 @@
  *   51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.         *
  ***************************************************************************/
 
-#include "oidentdconfiggenerator.h"
-#include "corenetwork.h"
-
 #include <QString>
+
+#include "corenetwork.h"
+#include "oidentdconfiggenerator.h"
 
 OidentdConfigGenerator::OidentdConfigGenerator(bool strict, QObject *parent) :
     QObject(parent),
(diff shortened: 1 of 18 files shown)


src/core/oidentdconfiggenerator.cpp
```diff
@@ -65,11 +69,20 @@ bool OidentdConfigGenerator::init()
     return _initialized;
 }
 
+const QString OidentdConfigGenerator::sysidentForIdentity(const CoreIdentity *identity) {
+    const QString &desiredIdent = identity->ident();
```
AlD: Just skip this one-time use variable and return directly below.

src/core/oidentdconfiggenerator.h
```diff
@@ -58,20 +59,26 @@ class OidentdConfigGenerator : public QObject
 {
     Q_OBJECT
 public:
-    explicit OidentdConfigGenerator(QObject *parent = 0);
+    /**
+     * @param strict If false, any identity a user chooses is reported to servers as authoritative.
+     *               If true, an explicit entry in the database must be present to use a sysident other than the authusername.
```
AlD: This seems to be referring to the old version of this patch still.

mamarley:
I updated the comment and removed the one-time-use variables as requested.  Thanks for having a look!

esainane:
What remains to be done to make this mergeworthy?

mamarley:
I don't imagine anything else major will need to be done; I think we are just waiting for Sput to have time.

PR #351, opened 2018-04-10
Coreside Realname/Avatar URL Storage

justjanne:
## In short
- Store avatar URL (currently just "", as we don’t implement METADATA yet) for each sender in database
- Store realname for each sender in database
- Add feature RichMessages to control serialization of these
- Transmit them over the network to clients in messages

### Impact
| Criteria | Rank | Reason |
| - | - | - |
| Impact | *★☆☆&nbsp;1/3* | No user-facing changes, only visible to users in third-party clients |
| Risk | *★★★&nbsp;2/3* | Has to recreate, migrate, and drop the entire senders table |
| Intrusiveness | *★☆☆&nbsp;1/3* | No breaking core changes, interference with other PRs limited to Quassel::Features allocation and sql.qrc |

## Rationale
To show the realname on clients, as well as the user avatar, we need to store these with messages. These features are available in Slack, Discord and IRCCloud, and we’d like to provide them as well.

## Examples
[Screenshot showing realnames and avatars in action](https://i.k8r.eu/cxqY2Q)  
In this screenshot, Quasseldroid-NNG uses the e-mail in the realname to show a fallback avatar (through Gravatar, this feature is optional).

**To be done in the future:**
* Add quasselclient UI to use these features
* Add actual avatar URL functionality with IRCv3 METADATA

### DIFF
--- a/src/common/message.cpp
+++ b/src/common/message.cpp
@@ -26,24 +26,31 @@
 
 #include <QDataStream>
 
-Message::Message(const BufferInfo &bufferInfo, Type type, const QString &contents, const QString &sender, const QString &senderPrefixes, Flags flags)
+Message::Message(const BufferInfo &bufferInfo, Type type, const QString &contents, const QString &sender,
+                 const QString &senderPrefixes, const QString &realName, const QString &avatarUrl, Flags flags)
     : _timestamp(QDateTime::currentDateTime().toUTC()),
     _bufferInfo(bufferInfo),
     _contents(contents),
     _sender(sender),
     _senderPrefixes(senderPrefixes),
+    _realName(realName),
+    _avatarUrl(avatarUrl),
     _type(type),
     _flags(flags)
 {
 }
 
 
-Message::Message(const QDateTime &ts, const BufferInfo &bufferInfo, Type type, const QString &contents, const QString &sender, const QString &senderPrefixes, Flags flags)
+Message::Message(const QDateTime &ts, const BufferInfo &bufferInfo, Type type, const QString &contents,
+                 const QString &sender, const QString &senderPrefixes, const QString &realName,
+                 const QString &avatarUrl, Flags flags)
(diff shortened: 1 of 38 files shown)


src/common/message.cpp
```diff
@@ -63,6 +69,12 @@ QDataStream &operator<<(QDataStream &out, const Message &msg)
     if (SignalProxy::current()->targetPeer()->hasFeature(Quassel::Feature::SenderPrefixes))
         out << msg.senderPrefixes().toUtf8();
 
+    if (SignalProxy::current()->targetPeer()->hasFeature(Quassel::Feature::RichMessages))
```
mamarley: Both of these conditionals are the same.  I think it would be cleaner and probably more efficient to just use one if() statement with {}.

src/common/message.cpp
```diff
@@ -95,6 +107,16 @@ QDataStream &operator>>(QDataStream &in, Message &msg)
         in >> senderPrefixes;
     msg._senderPrefixes = QString::fromUtf8(senderPrefixes);
 
+    QByteArray realName;
+    if (SignalProxy::current()->sourcePeer()->hasFeature(Quassel::Feature::RichMessages))
```
mamarley: Same here with the if() statements

src/common/message.cpp
justjanne: Mhm, I can see where you're coming from — that said, I was trying to make it a bit easier to read the conditions for each field.

If you wish, I can change this, though.

mamarley: Stylistically I don't really care one way or the other, but the efficiency aspect of doing the same comparison twice in a row like that bothers me.

Sput42:
@digitalcircuit If you find time, it would be greatly appreciated if you could unleash your testing routine on this, with a focus on the database setup and migration stuff!

PR #353, opened 2018-04-12
Force Qt 5.9.4 for OSX builds on Travis CI

Sput42:
Qt 5.10.1 is broken.


Sput42:
This isn't going to work, so we'll have to live with 5.10.1 until a newer version is release.

PR #362, opened 2018-05-30
common: Add '$i:identd' verified ident alias, polish

digitalcircuit:
## In short
* Add `$i:identd` parameter to aliases
  * Represents `ident` if verified, or `*` if unknown or unverified
  * Allows ban rules to respect identd, e.g. `/mode $channel +b *!$1:identd@$1:hostname; /kick $1`
* Replace empty nickname parameters with `*`
  * Using the variables for nicknames before `WHO`'ng them still provides results
  * May be dangerous if using above ban example on nickname without hostname available
* Update alias configuration tooltips
  * Add `identd`
  * Organize everything into a table
* Update comments

Criteria | Rank | Reason
Impact | ★★☆ *2/3* | User-facing moderator improvement, tooltip cleanups
Risk | ★★☆ *2/3* | Possible confusion, small behavior change to existing aliases

*Thanks to `Exterminador` for suggesting the change, and @justJanne for further clarifications!*

### Background
Channel moderators will need to ban troublesome users quickly and effectively, while at the same time catching as few innocent people as possible.

IRC servers can enable `identd` checking to ask any connecting user to verify their identity.  If the check fails, the `ident` gets prefixed with `~`, distinguishing them as unverified.  Many shared bouncers have a strict `ident` policy with an `identd` server running to back it up, while many individual users do not bother with `identd`.

By distinguishing between `ident` for server-verified users and wildcard for unverified, aliases can easily ban either an entire IP address for home users, or just a single ident for shared bouncer users.

Though not perfect, this addresses most common cases.

### Motivation
Quassel should offer a way to distinguish between verified and unverified idents in aliases to allow for easy banning/moderation.

For example...
```
/mode $channel +b *!$1:identd@$1:hostname; /kick $1
```
...would target `*!*@no-identd.example.com` for someone connecting without an `identd` server, and `*!userid@shared.example.com` for someone connecting with an `identd` server.

Furthermore, empty parameters (*e.g. if no `/who` has been run on a user already in a channel*) should be replaced with `*` to allow them to still be usable.  **This may be dangerous if using the example ban alias without checking if the user has a hostmask first.**

*Note: Quassel currently returns `*` for any nonexistent nick, so if someone quits between typing a command and running it, this situation could happen before this pull request.  Alternatively, `$i:hostmask` could be made inconsistent with the rest of the parameters, kept as empty when empty, relying on the IRC server to return an error.  And [there's a suggestion for customizable fallbacks](https://github.com/quassel/quassel/pull/362#issuecomment-393972258 ), a potentially rather complex thing.*

### Testing alias
*Alias `/testalias` created with text...*
*`/say nick: $1, account: $1:account, hostname: $1:hostname, ident: $1:ident, identd: $1:identd`*

Nickname | Reported ident | Configuration
---------|----------------|--------------
`justJanne` | `kuschku` | Verified, runs an `identd` server
`dcircuit_dev` | `~quasselde`| Unverified, does not run an `identd` server
`somenick` | *Empty* | No available information as no `/who` run

```
/testalias justJanne
nick: justJanne, account: justJanne, hostname: lithium.kuschku.de, ident: kuschku, identd: kuschku
/testalias dcircuit_dev
nick: dcircuit_dev, account: *, hostname: […], ident: ~quasselde, identd: *
/testalias anatolik
nick: somenick, account: *, hostname: *, ident: *, identd: *
```

### Tooltips
*Tooltips in `Settings` → `Configure Quassel…` → `IRC` → `Aliases` → hover over `Expansion` column*

**Before**
'Configure Aliases' window in Quassel settings showing the tooltip from hovering over the 'Expansion' column, before the changes

> **The string the shortcut will be expanded to**
> **special variables**
> \- **$i** represents the i'th parameter.
> \- **$i..j** represents the i'th to j'th parameter separated by spaces.
> \- **$i..** represents all parameters from i on separated by spaces.
> \- **$i:hostname** represents the hostname of the user identified by the i'th parameter or a * if unknown.
> \- **$i:ident** represents the ident of the user identified by the i'th parameter or a * if unknown.
> \- **$i:account** represents the account of the user identified by the i'th parameter or a * if logged out or unknown.
> \- **$0** the whole string
> \- **$nick** your current nickname
> \- **$channel** the name of the selected channel
> 
> Multiple commands can be separated with semicolons
> 
> **Example:** "Test $1; Test $2; Test All $0" will be expanded to three separate messages "Test 1", "Test 2" and "Test All 1 2 3" when called like /test 1 2 3

**After**
'Configure Aliases' window in Quassel settings showing the tooltip from hovering over the 'Expansion' column, after the changes

> **The string the shortcut will be expanded to**
> \[Following is centered\] **Special variables**
> 
> \[Column header invisible\] | \[Column header invisible\]
> ----------------------------|----------------------------
> \[Two-column cell\] | *Parameter variables*
> **$i** | i'th parameter
> **$i..j** | i'th to j'th parameter separated by spaces
> **$i..** | all parameters from i on separated by spaces
> \[Two-column cell\] | *Nickname parameter variables*
> **$i:account** | account of the user identified by the i'th parameter, or a '*' if logged out or unknown
> **$i:hostname** | hostname of the user identified by the i'th parameter, or a '*' if unknown
> **$i:ident** | ident of the user identified by the i'th parameter, or a '*' if unknown
> **$i:identd** | ident of the user identified by the i'th parameter if verified, or a '*' if unknown or unverified (prefixed with '~')
> \[Two-column cell\] | *General variables*
> **$0** | the whole string
> **$nick** | your current nickname
> **$channel** | the name of the selected channel
> 
> Multiple commands can be separated with semicolons
> 
> **Example:** "Test $1; Test $2; Test All $0"
> ...will be expanded to three separate messages "Test 1", "Test 2" and "Test All 1 2 3" when called like */test 1 2 3*

### DIFF
--- a/src/common/aliasmanager.cpp
+++ b/src/common/aliasmanager.cpp
@@ -187,10 +187,43 @@ void AliasManager::expand(const QString &alias, const BufferInfo &bufferInfo, co
         }
 
         for (int j = params.count(); j > 0; j--) {
+            // Find the referenced IRC user...
             IrcUser *ircUser = net->ircUser(params[j - 1]);
-            command = command.replace(QString("$%1:hostname").arg(j), ircUser ? ircUser->host() : QString("*"));
-            command = command.replace(QString("$%1:ident").arg(j), ircUser ? ircUser->user() : QString("*"));
-            command = command.replace(QString("$%1:account").arg(j), ircUser ? ircUser->account() : QString("*"));
+            // ...and replace components, using short-circuit evaluation as ircUser might be null
+
+            // Account, or "*" if blank/nonexistent/logged out
+            command = command.replace(
+                        QString("$%1:account").arg(j),
+                        (ircUser && !ircUser->account().isEmpty()) ? ircUser->account()
+                                                                   : QString("*"));
+
+            // Hostname, or "*" if blank/nonexistent
+            command = command.replace(
+                        QString("$%1:hostname").arg(j),
+                        (ircUser && !ircUser->host().isEmpty()) ? ircUser->host() : QString("*"));
+
+            // Identd
+            // Ident if verified, or "*" if blank/unknown/unverified (prefixed with "~")
+            //
+            // Most IRC daemons have the option to prefix an ident with "~" if it could not be
+            // verified via an identity daemon such as oidentd.  In these cases, it can be handy to
+            // have a way to ban via ident if verified, or all idents if not verified.  If the
+            // server does not verify idents, it usually won't add "~".
+            //
+            // Identd must be replaced before ident to avoid being treated as "$i:ident" + "d"
+            command = command.replace(
+                        QString("$%1:identd").arg(j),
+                        (ircUser && !ircUser->user().isEmpty()
+                         && !ircUser->user().startsWith("~"))
+                        ? ircUser->user() : QString("*"));
+
+            // Ident, or "*" if blank/nonexistent
+            command = command.replace(
+                        QString("$%1:ident").arg(j),
+                        (ircUser && !ircUser->user().isEmpty()) ? ircUser->user() : QString("*"));
+
+            // Nickname
+            // Must be replaced last to avoid interferring with more specific aliases
             command = command.replace(QString("$%1").arg(j), params[j - 1]);
         }
         command = command.replace("$0", msg);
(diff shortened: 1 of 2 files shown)


digitalcircuit:
@romibi had the nice idea of having aliases let you specify the fallback content.

So `$i:hostname:foo` becomes `hostname` if known, `foo` if not.

This is not implemented in this pull request as of now, but I can try to add it if preferred.

I'm a little concerned about complexity since this would need some form of escaping.  Otherwise...
```
/cmd nospaces$i:hostname:continuedcommand
```
...would be interpreted as...
```
/cmd nospaces[$i:hostname:continuedcommand]
```
...instead of...
```
/cmd nospaces[$i:hostname]:continuedcommand
```

PR #366, opened 2018-06-04
Properly fix the SettingsDlg layout

Sput42:
We've suffered from the page tree in the settings dialog being
squashed horizontally for a long time. A previous commit (65c463f0)
tried to fix this by calculating a minimum size and explicitly
resizing the dialog, but this didn't properly work at least on HIDPI
displays.

Instead, fix layouting properly by setting the correct size policy
for the tree widget, and removing both the afforementioned workaround
and a previous hack that tried to set the minimum width manually
(and failed due to a still-hidden widget reporting wrong size hints).

### DIFF
--- a/src/qtui/settingsdlg.cpp
+++ b/src/qtui/settingsdlg.cpp
@@ -45,22 +45,6 @@ SettingsDlg::SettingsDlg(QWidget *parent)
     connect(Client::instance(), SIGNAL(coreConnectionStateChanged(bool)), SLOT(coreConnectionStateChanged()));
 
     setButtonStates();
-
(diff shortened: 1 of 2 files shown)


digitalcircuit:
Seems to work here!  One caveat, though...

There's no way to make the Settings dialog smaller horizontally.  We'll need to be careful with settings pages to make sure they never push the dialog past the maximum usable size for devices with small screens (*thankfully less of an issue now than it was 2 years ago*).

Sput42:
Just for the record: It couldn't be made much smaller before this change (only the settings tree would allow shrinking), although obviously it's a bit wider now since we display the tree correctly.

I haven't found a way to allow the page area to be shrunk; maybe would have to nest it in a scroll area or something...

PR #370, opened 2018-06-10
Cleans protocol for highlight rules

justjanne:
# In Short
- Sync highlightNick and nicksCaseSensitive as actual Q_PROPERTY
- Identify rules by their id, not their name
- This allows multiple rules to have the same rule, but different scopes
  and senders

# Impact
This breaks the protocol with existing 0.13 clients. They can still connect, but cannot configure highlight rules.

### DIFF
--- a/src/common/highlightrulemanager.cpp
+++ b/src/common/highlightrulemanager.cpp
@@ -20,12 +20,14 @@
 
 #include "highlightrulemanager.h"
 
-#include "util.h"
-
 #include <QDebug>
 
+#include "util.h"
+
 INIT_SYNCABLE_OBJECT(HighlightRuleManager)
-HighlightRuleManager &HighlightRuleManager::operator=(const HighlightRuleManager &other) {
+
+HighlightRuleManager &HighlightRuleManager::operator=(const HighlightRuleManager &other)
+{
     if (this == &other)
         return *this;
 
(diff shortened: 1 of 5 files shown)


digitalcircuit:
For others wondering why, this is useful for ignore rules, to allow multiple `.*` matches (e.g. ignoring bot highlights), but also is handy for other purposes.

A quick test shows this as working with monolithic and new client/core!  Even after restart and such.  Default nick highlight correctly still appears to be `CurrentNick`, too.

**NOTE:  This will also erase all existing core-side highlights.**  As this a 0.13-only feature, this may be an acceptable trade-off to avoid hassle of migration code.

If time permits, I will try to fit in more thorough testing.  This **needs applied before 0.13 beta due to protocol break.**  This was not a thorough test.

### Examples

**Monolithic, highlight rules**
Custom highlights showing a list of "Bitlbee" with sender a, duplicated, then a final one with sender "root!*"

**Monolithic, above rules working**
Chat log from &bitlbee showing the "Bitlbee" highlight rule working for sender "root!*"

**Client + core, highlight rules**
Custom highlights showing a list of "Bitlbee" with sender a, duplicated, then a final one with sender "root!*"

**Client + core, above rules working**
Chat log from &bitlbee showing the "Bitlbee" highlight rule working for sender "root!*"

PR #371, opened 2018-06-14
client: Option for highest sender mode

digitalcircuit:
## In short

* Optionally show only the highest sender mode prefix
  * `All modes`, `Highest mode` (**new default**), `No modes` (**migrated default**)
  * Keep existing users on no modes to match old behavior
  * Use highest mode for new configuration to match other clients
  * *`0.13-master` users will need to re-enable sender prefixes after this*
* Clean up and improve settings page handling
  * Fix `SettingsPage::hasChanged()` widget-specific handling, support `FontSelector`
  * `ChatViewSettings` check if needing to reload stylesheet before doing it
  * Fix settings migration unintentional fallthrough

Criteria | Rank | Reason
Impact | ★★☆ *2/3* | User-facing requested feature, fixes bugs
Risk | ★★☆ *2/3* | More options to manage, possible settings regressions
Intrusiveness | ★★☆ *2/3* | UI, string changes, may interfere with other PRs

## Rationale
Quassel now properly handles multiple prefixes at any given time.  However, in most cases, one only needs to know the highest mode of any given user - if someone is an operator, it stands to reason they may have voice, too.

Thus, Quassel should offer a way to only show the highest sender mode, reducing clutter, while still giving the option to see all modes should someone need it, as well as turn off sender modes for those who find them unncessary.

### Sender prefix display modes
**Setup**
1.  Join an empty channel on a `multi-prefix` capable server, e.g. Freenode
2.  `/mode +v your_nick`, so you have `@` operator and `+` voice
3.  Send a message

**No modes**
*This is the default for people upgrading existing Quassel setups.*
Chat view showing a test message from nick `dcircuit_dev` with no prefix

```
<dcircuit_dev> test
<dcircuit_dev> This is a test message.  Test successful.
```

**Highest mode**
*This is the default for new installations of Quassel.*
Chat view showing a test message from nick `dcircuit_dev` with '@' prefix

```
<@dcircuit_dev> test
<@dcircuit_dev> This is a test message.  Test successful.
```

**All modes**
Chat view showing a test message from nick `dcircuit_dev` with `@+` prefixes

```
<@+dcircuit_dev> test
<@+dcircuit_dev> This is a test message.  Test successful.
```

### Chat View settings for sender modes
*In Settings → Interface → Chat View*
Chat View settings dialog showing the configuration dropdown for sender modes

*Content*
> Show sender modes before nicknames: \[dropdown box here\]

*Tooltip*
> **Sender modes:**
> 
> *No modes:* Don't show any modes.
> *Example:* \<nickname\>
> 
> *Highest mode:* Show only the highest active mode.
> *Example:* \<@ nickname\> \[space added to disable GitHub auto-link\]
> 
> *All modes:* Show all active modes.
> *Example:* \<\@+nickname\>

### Sender prefixes not supported by core
*In Settings → Interface → Chat View*
Chat View settings dialog showing the out of date core tooltip

*Tooltip*
> **Your Quassel core does not support this feature**
> 
> You need a Quassel core v0.13.0 or newer in order to show sender modes before nicknames.

### DIFF
--- a/src/qtui/settingspages/chatviewsettingspage.ui
+++ b/src/qtui/settingspages/chatviewsettingspage.ui
@@ -93,23 +93,49 @@
     </widget>
    </item>
    <item>
-    <widget class="QCheckBox" name="showSenderPrefixes">
-     <property name="toolTip">
-      <string>Shows the modes of senders before their name (e.g. @, +)</string>
-     </property>
-     <property name="text">
-      <string>Show sendermodes in front of nicknames</string>
-     </property>
-     <property name="checked">
-      <bool>true</bool>
-     </property>
-     <property name="defaultValue" stdset="0">
-      <bool>true</bool>
-     </property>
-     <property name="settingsKey" stdset="0">
-      <string notr="true">ShowSenderPrefixes</string>
-     </property>
-    </widget>
+    <layout class="QHBoxLayout" name="horizontalLayout_2">
+     <item>
+      <widget class="QLabel" name="label">
+       <property name="text">
+        <string>Show sender modes before nicknames:</string>
+       </property>
+      </widget>
+     </item>
+     <item>
+      <widget class="QComboBox" name="senderPrefixComboBox">
+       <property name="toolTip">
+        <string>&lt;!DOCTYPE HTML PUBLIC &quot;-//W3C//DTD HTML 4.0//EN&quot; &quot;http://www.w3.org/TR/REC-html40/strict.dtd&quot;&gt;
+&lt;html&gt;&lt;head&gt;&lt;meta name=&quot;qrichtext&quot; content=&quot;1&quot; /&gt;&lt;style type=&quot;text/css&quot;&gt;
+p, li { white-space: pre-wrap; }
+&lt;/style&gt;&lt;/head&gt;&lt;body style=&quot;font-weight:400; font-style:normal;&quot;&gt;
+&lt;p&gt;&lt;span style=&quot;font-weight:600;&quot;&gt;Sender modes:&lt;/span&gt;&lt;/p&gt;
+&lt;p&gt;&lt;span style=&quot;text-decoration: underline;&quot;&gt;No modes:&lt;/span&gt; Don't show any modes&lt;br/&gt;&lt;span style=&quot;font-style: italic;&quot;&gt;Example:&lt;/span&gt; &amp;lt;nickname&amp;gt;&lt;/p&gt;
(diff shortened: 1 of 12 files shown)


src/uisupport/uistyle.h
```diff
@@ -162,6 +163,15 @@ class UiStyle : public QObject
         NumRoles // must be last!
     };
 
+    /// Display of sender prefix modes
+    enum class SenderPrefixMode {
```
Sput42: enum class is not supported in Q_ENUMS for Qt4. Should be a normal old-style enum for now.

src/uisupport/uistyle.h
```diff
@@ -335,7 +345,7 @@ private slots:
     static bool _useCustomTimestampFormat;        /// If true, use the custom timestamp format
     static QString _systemTimestampFormatString;  /// Cached copy of system locale timestamp format
     static QString _timestampFormatString;        /// Timestamp format string
-    static bool _showSenderPrefixes;              /// If true, show prefixmodes before sender names
+    static UiStyle::SenderPrefixMode _senderPrefixDisplay; /// Display of prefix modes before sender
```
Sput42: I just notice this now: To make such comments work properly in Doxygen, it needs to be `///<` as opposed to `///`. The lt character means "applies to entity left of the comment".
Feel free to decide if you want to adapt the others in this class, or just leave as-is; just pointing it out for the future.

src/qtui/settingspages/chatviewsettingspage.ui
```diff
@@ -93,23 +93,49 @@
...
+      <widget class="QLabel" name="label">
+       <property name="text">
+        <string>Show sender modes before nicknames:</string>
+       </property>
+      </widget>
+     </item>
+     <item>
+      <widget class="QComboBox" name="senderPrefixComboBox">
+       <property name="toolTip">
+        <string>&lt;!DOCTYPE HTML PUBLIC &quot;-//W3C//DTD HTML 4.0//EN&quot; &quot;http://www.w3.org/TR/REC-html40/strict.dtd&quot;&gt;
```
Sput42: I'm still not sure if full-blown HTML is going to make it easy for translators - do common translation tools (or Transifex?) handle this?

src/uisupport/uistyle.h
digitalcircuit: Noted!  I'll add a comment for `TODO Qt5`, too.

As it turns out, I remembered to use the `///<` format here, but missed it elsewhere as you commented... :man_facepalming:

digitalcircuit:
Cleaned up according to requests, and fixed an unnecessary translation effort in the core-side highlights settings page, too.

I used separate commits for reference, so in the future if someone adds HTML via Qt's Rich Text editor, this might give an idea of how to fix it.  Though the commits could be merged together more if desired.

PR #381, opened 2018-06-18
client: Handle legacy CoreInfo, fix sync, disconnect

digitalcircuit:
## In short

* Synchronize core info before first connect
  * Removes warning about missing receiver for core info signal
* Refresh legacy core info whenever opening Core Info dialog
  * On legacy cores, fetch core information by recreating `CoreInfo` object
  * Add feature flag `SyncedCoreInfo`, don't poll on modern cores
  * Fixes information only refreshing once when connecting
* Indicate in core info dialog when disconnected
  * Reduces confusion about state of the core
  * Before, core info dialog would simply stop updating

Criteria | Rank | Reason
Impact | ★★☆ *2/3* | User-facing polish, better client experience for old cores
Risk | ★★☆ *2/3* | Core Info might stop being refreshed, become corrupted
Intrusiveness | ★★☆ *2/3* | Protocol changes, could interfere with other PRs

### Disconnected from core
*In File → Core Info…*
*Happens if you disconnect the current session, or if the current session gets disconnected, either remotely or due to server outage.*

**Before**
File - Core Info, showing state after remote disconnect on previous version
*The Core Information dialog stops updating, with `Ending session...` stuck on the button and nothing else working.*

**After**
File - Core Info, showing state after remote disconnect

> Item | Contents
> -----|---------
> Version: | Disconnected from core
> Version Date: | Not available
> Uptime: | Not available
> Connected Clients: | 0

### Sync warning
Connect to modern core.
**Before**
```
2018-06-17 22:12:48 Warning: "no registered receiver for sync call: CoreInfo::setCoreData (objectName=\"\"). Params are:" (QVariant(QVariantMap, QMap(("quasselBuildDate", QVariant(QString, "Sun Jun 17 17:10:58 2018"))("quasselVersion", QVariant(QString, "v0.13-pre (0.12.0+660 git-<a href=\"https://github.com/quassel/quassel/commit/5e4bfcea8cf62cded326d385acd47528c1974b09\">5e4bfce</a>)"))[...]("sessionConnectedClients", QVariant(int, 1))("startTime", QVariant(QDateTime, QDateTime(2018-06-18 03:12:43.229 UTC Qt::TimeSpec(UTC)))))))
```

**After**
*No warnings printed*

### Legacy cores
1.  Connect to legacy core
2.  Open Core Info dialog
3.  Connect other client to legacy core
4.  Close and re-open Core Info dialog in the first client

**Before**
Connected Clients would not update.

**After**
Connected Clients shows the correct value.

### DIFF
--- a/src/client/client.cpp
+++ b/src/client/client.cpp
@@ -101,7 +101,7 @@ Client::Client(QObject *parent)
     _backlogManager(new ClientBacklogManager(this)),
     _bufferViewManager(0),
     _bufferViewOverlay(new BufferViewOverlay(this)),
-    _coreInfo(nullptr),
+    _coreInfo(new CoreInfo(this)),
     _dccConfig(0),
     _ircListHelper(new ClientIrcListHelper(this)),
     _inputHandler(0),
@@ -174,6 +174,9 @@ void Client::init()
     connect(backlogManager(), SIGNAL(messagesReceived(BufferId, int)), _messageModel, SLOT(messagesReceived(BufferId, int)));
 
     coreAccountModel()->load();
+
+    // Attach CoreInfo
+    p->synchronize(coreInfo());
 
     connect(coreConnection(), SIGNAL(stateChanged(CoreConnection::ConnectionState)), SLOT(connectionStateChanged(CoreConnection::ConnectionState)));
     coreConnection()->init();
@@ -407,11 +410,6 @@ void Client::setSyncedToCore()
     SignalProxy *p = signalProxy();
     p->synchronize(bufferSyncer());
 
-    // create CoreInfo
-    Q_ASSERT(!_coreInfo);
-    _coreInfo = new CoreInfo(this);
-    p->synchronize(coreInfo());
-
     // create a new BufferViewManager
     Q_ASSERT(!_bufferViewManager);
     _bufferViewManager = new ClientBufferViewManager(p, this);
@@ -486,6 +484,27 @@ void Client::requestInitialBacklog()
 }
 
 
+void Client::requestLegacyCoreInfo()
+{
+    // On older cores, the CoreInfo object was only synchronized on demand.  Synchronize now if
+    // needed.
+    if (isConnected() && !isCoreFeatureEnabled(Quassel::Feature::SyncedCoreInfo)) {
+        // Delete the existing core info object (it will always exist as client is single-threaded)
+        _coreInfo->deleteLater();
+        // No need to set to null when creating new one immediately after
+
+        // Create a fresh, unsynchronized CoreInfo object, emulating legacy behavior of CoreInfo not
+        // persisting
+        _coreInfo = new CoreInfo(this);
+        // Synchronize the new object
+        signalProxy()->synchronize(_coreInfo);
+
+        // Let others know signal handlers have been reset
+        emit coreInfoResynchronized();
+    }
+}
+
+
 void Client::disconnectFromCore()
 {
     if (!coreConnection()->isConnected())
@@ -512,10 +531,7 @@ void Client::setDisconnectedFromCore()
         _bufferSyncer = 0;
     }
 
-    if (_coreInfo) {
-        _coreInfo->deleteLater();
-        _coreInfo = nullptr;
-    }
+    _coreInfo->reset();
 
     if (_bufferViewManager) {
         _bufferViewManager->deleteLater();
@@ -708,6 +724,12 @@ void Client::markBufferAsRead(BufferId id)
 {
     if (bufferSyncer() && id.isValid())
         bufferSyncer()->requestMarkBufferAsRead(id);
+}
+
+
+void Client::refreshLegacyCoreInfo()
+{
+    instance()->requestLegacyCoreInfo();
 }
 
 
--- a/src/client/client.h
+++ b/src/client/client.h
@@ -149,6 +149,17 @@ public:
     static void mergeBuffersPermanently(BufferId bufferId1, BufferId bufferId2);
     static void purgeKnownBufferIds();
 
+    /**
+     * Requests client to resynchronize the CoreInfo object for legacy (pre-0.13) cores
+     *
+     * This provides compatibility with updating core information for legacy cores, and can be
+     * removed after protocol break.
+     *
+     * NOTE: On legacy (pre-0.13) cores, any existing connected signals will be destroyed and must
+     * be re-added after calling this, in addition to checking for existing data in coreInfo().
+     */
+    static void refreshLegacyCoreInfo();
+
     static void changePassword(const QString &oldPassword, const QString &newPassword);
     static void kickClient(int peerId);
 
@@ -177,6 +188,17 @@ signals:
     void connected();
     void disconnected();
     void coreConnectionStateChanged(bool);
+
+    /**
+     * Signals that core information has been resynchronized, removing existing signal handlers
+     *
+     * Whenever this is emitted, one should re-add any handlers for CoreInfo::coreDataChanged() and
+     * apply any existing information in the coreInfo() object.
+     *
+     * Only emitted on legacy (pre-0.13) cores.  Generally, one should use the
+     * CoreInfo::coreDataChanged() signal too.
+     */
+    void coreInfoResynchronized();
 
     //! The identity with the given ID has been newly created in core and client.
     /** \param id The ID of the newly created identity.
@@ -257,6 +279,17 @@ private:
     void init();
 
     void requestInitialBacklog();
+
+    /**
+     * Deletes and resynchronizes the CoreInfo object for legacy (pre-0.13) cores
+     *
+     * This provides compatibility with updating core information for legacy cores, and can be
+     * removed after protocol break.
+     *
+     * NOTE: On legacy (pre-0.13) cores, any existing connected signals will be destroyed and must
+     * be re-added after calling this, in addition to checking for existing data in coreInfo().
+     */
+    void requestLegacyCoreInfo();
 
     static void addNetwork(Network *);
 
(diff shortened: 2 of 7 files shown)


src/client/client.cpp
```diff
@@ -486,6 +484,24 @@ void Client::requestInitialBacklog()
...
 
+void Client::requestLegacyCoreInfo()
+{
+    // On older cores, the CoreInfo object was only synchronized on demand.  Synchronize now if
+    // needed.
+    if (isConnected() && !isCoreFeatureEnabled(Quassel::Feature::SyncedCoreInfo)) {
+        // Delete the existing core info object if it exists
+        if (_coreInfo) {
+            _coreInfo->deleteLater();
+            _coreInfo = nullptr;
```
Sput42: No need to null this if you're creating a new one immediately after.

src/client/client.cpp
```diff
@@ -513,8 +529,11 @@ void Client::setDisconnectedFromCore()
     }
 
     if (_coreInfo) {
-        _coreInfo->deleteLater();
-        _coreInfo = nullptr;
+        // coreInfo may be deleted as part of refreshing legacy core info data.  If this happens,
+        // there's no need to reset the data as an entirely new object is created.
+        //
+        // This shouldn't happen, but better check to be safe.
+        _coreInfo->reset();
```
Sput42: _coreInfo is now always valid, so no need to check for null...

src/client/client.cpp
```diff
@@ -711,6 +730,14 @@ void Client::markBufferAsRead(BufferId id)
 }
 
 
+void Client::refreshLegacyCoreInfo()
+{
+    if (instance()) {
```
Sput42: instance() always returns a valid pointer, so no need to check.

src/client/client.h
```diff
@@ -258,6 +266,17 @@ private slots:
 
     void requestInitialBacklog();
 
+    /**
+     * Deletes and resynchronizes the CoreInfo object for legacy (pre-0.13) cores
+     *
+     * CAUTION: This will delete the CoreInfo object for a brief period of time.  Any use of
+     * Client::coreInfo() should check that the reference is valid (not null) before using it.
```
Sput42: The entire client is single-threaded, so this cannot happen. If we ever wanted to make the client thread-safe, we'd have to do much more than checking for null, due to synchronization...

src/qtui/coreinfodlg.cpp
```diff
@@ -28,10 +28,18 @@
...
-    connect(coreInfo, SIGNAL(coreDataChanged(const QVariantMap &)), this, SLOT(coreInfoChanged(const QVariantMap &)));
 
-    coreInfoChanged(coreInfo->coreData());
+    // Update legacy core info for Quassel cores earlier than 0.13.  This does nothing on modern
+    // cores.
+    Client::refreshLegacyCoreInfo();
+
+    CoreInfo *coreInfo = Client::coreInfo();
+    // Guard against opening dialog during CoreInfo reset
+    if (coreInfo) {
```
Sput42: Can't be null now.

src/qtui/coreinfodlg.cpp
```diff
@@ -80,14 +104,28 @@ void CoreInfoDlg::coreInfoChanged(const QVariantMap &coreInfo) {
...
+    // Update uptime for immediate display, don't wait for the timer
+    updateUptime();
 }
 
 
 void CoreInfoDlg::updateUptime() {
     CoreInfo *coreInfo = Client::coreInfo();
-    if (coreInfo) {
+
+    if (!coreInfo) {
```
Sput42: Can't happen.

src/client/client.cpp
digitalcircuit: Good catch!  I still need to call `_coreInfo->deleteLater();` , correct?

(And also remove the null check on `_coreInfo` since it's always not null)

src/client/client.h
digitalcircuit: We do *eventually* want to make the client thread-safe, I think, for the sake of much-improved loading times (*see: Quasseldroid-ng*), but that's a task for after 0.13 at minimum.

Noted!

src/qtui/coreinfodlg.cpp
```diff
@@ -28,10 +28,18 @@
...
-    coreInfoChanged(coreInfo->coreData());
+    // Update legacy core info for Quassel cores earlier than 0.13.  This does nothing on modern
+    // cores.
+    Client::refreshLegacyCoreInfo();
+
+    CoreInfo *coreInfo = Client::coreInfo();
+    // Guard against opening dialog during CoreInfo reset
+    if (coreInfo) {
+        connect(coreInfo, SIGNAL(coreDataChanged(const QVariantMap &)),
+                this, SLOT(coreInfoChanged(const QVariantMap &)));
```
Sput42: This connection gets deleted if coreInfo gets destroyed by the client. So you need to somehow update the dialog, if you want to handle that, otherwise information from the new coreInfo will never arrive in the dialog.

src/client/client.cpp
```diff
@@ -486,6 +484,24 @@ void Client::requestInitialBacklog()
...
+    if (isConnected() && !isCoreFeatureEnabled(Quassel::Feature::SyncedCoreInfo)) {
+        // Delete the existing core info object if it exists
+        if (_coreInfo) {
+            _coreInfo->deleteLater();
+            _coreInfo = nullptr;
+        }
+        // Create a new one
+        _coreInfo = new CoreInfo(this);
+        // Synchronize the new object
+        signalProxy()->synchronize(_coreInfo);
```
Sput42: If you want to update an open CoreInfoDlg, you need to emit a signal here to make it aware of the change...

src/client/client.h
Sput42: I don't see how making the client multithreaded would speed up time (still need to do all the work), so I don't expect this to happen. We'd probably want to rework the complete architecture in that case, to get rid of the central singleton, and add workers for things like backlog parsing... in any case, a null check wouldn't be sufficient, because in the multithreaded case, the object could be destroyed just after you checked for it not being null!

src/client/client.cpp
Sput42: Yes, deleteLater() is required to actually clean up the old instance. What that does is send an event to schedule object deletion, so next time the event loop runs, it's deleted.

src/client/client.cpp
digitalcircuit: I had mimicked the behavior of e.g. `Client::markBufferAsRead()`, with `if (bufferSyncer()) [...]`, not paying attention to the matter that `_bufferSyncer` is deleted on disconnect...

Thanks for catching this!

src/client/client.h
digitalcircuit: Mainly, multi-core CPUs.  But agreed, there's some bigger, easier tasks first, like background parsing of backlog, selective synchronization of `IrcUser` objects (*needed for Quasseldroid to connect faster*)...

Good point on the null check; I've got a lot to learn with regards to multi-threading in complex projects (*e.g. probably would need a lock*) :slightly_smiling_face:

src/qtui/coreinfodlg.cpp
digitalcircuit: True, but this connection is only made after the `coreInfo` object gets recreated.  There's a call to `coreInfoChanged(coreInfo->coreData());` right afterwards that applies any existing information, too.

I'll double-check with a `0.12.2` core (*I had used a `0.13` core from before the signal magic was added*)...

src/qtui/coreinfodlg.cpp
digitalcircuit: Good catch, fixed.

I also took this opportunity to have it print `Unknown` instead of an invalid date-time (*all zeros*) if `coreInfo()` somehow gets emptied while still connected.

src/client/client.cpp
digitalcircuit: Given the single-threaded nature, won't the synchronization finish before this function returns, and thus `CoreInfoDlg` will call `coreInfoChanged(coreInfo->coreData());` [as noted in other comment](https://github.com/quassel/quassel/pull/381#discussion_r196198225 ) and receive the new data?

At least, this appears to work in testing.

That said, I should probably emit a signal here, anyways.

src/client/client.h
Sput42: Multithreading is hard, which is why Qt users try to get around it (event-based is so much simpler and safer). Notwithstanding worker threads, which for example could handle tasks like backlog parsing, but the synchronization points should be kept to a minimum...

src/client/client.cpp
digitalcircuit: After further thought, adding a signal is somewhat involved for a legacy-only matter.

If you find this acceptable, I'll merely add a note to the function (*including the public variant*)...

```cpp
    /**
     * Deletes and resynchronizes the CoreInfo object for legacy (pre-0.13) cores
     *
     * This provides compatibility with updating core information for legacy cores, and can be
     * removed after protocol break.
     *
     * NOTE: On legacy (pre-0.13) cores, any existing connected signals will be destroyed and must
     * be re-added after calling this, in addition to checking for existing data in coreInfo().
     */
    void requestLegacyCoreInfo();
```

It's worthwhile noting that any connection to signals with a modern (`0.13`+) core will **also** need to check for existing data on first show.  This function won't destroy the `_coreInfo` object on modern cores, so updates will continue to work.

It'd be cleaner to use signals for handling resynchronization.. but is it worth adding that for legacy cores when signals already work without resynchronization in modern usage?

digitalcircuit: Err, nevermind this, I'll try it.

digitalcircuit:
Updated to address review changes.  Not yet ready, I intend to implement a periodic timer refresh in `CoreInfoDlg`.  Just letting you confirm these changes while I go address the other pull request review :slightly_smiling_face:

digitalcircuit:
Should be ready for review!  Legacy cores now have the `Connected Clients` count (*and the rest of the unchanging core info*) updated every 15 seconds.  Polling stops the moment the dialog is closed.

Modern cores (*with `SyncedCoreInfo` flag*) do not use polling, as before.

PR #387, opened 2018-06-19
/mode: Prepend current nickname if buffer has no name

genius3000:
Currently when running `/mode ...` without having a valid channel name or our current IRC nick as the first parameter, the current buffer name is prepended. This works great when running from a channel buffer or a user query (typically changing modes of other users will fail, but that's not a Quassel problem).  
When ran from the status buffer, the buffer name returned is an empty string, so we can check for this and prepend our current IRC nick to be able to shortcut (for example: `/mode +R`) from the status buffer. This does not effect the channel/user prepending as mentioned before.

### Testing
**Before:**
From channel: `/mode +R` ; Mode +R is set on the channel
From user query: `/mode +R` ; Server returns `Can't change mode for other users`
From server query: `/mode +R` ; Server returns `<server name>: No such nick/channel`
From status buffer: `/mode +R` ; Server returns `+R: No such nick/channel`

**After:**
From channel: `/mode +R` ; Mode +R is set on the channel
From user query: `/mode +R` ; Server returns `Can't change mode for other users`
From server query: `/mode +R` ; Server returns `<server name>: No such nick/channel`
From status buffer: `/mode +R` ; Mode +R is set on myself

Also tested this with `/m` aliased to both `/mode $0` and `/mode $channel $0` with the same results throughout.

### DIFF
--- a/src/core/coreuserinputhandler.cpp
+++ b/src/core/coreuserinputhandler.cpp
@@ -474,7 +474,6 @@ void CoreUserInputHandler::handleMode(const BufferInfo &bufferInfo, const QStrin
     Q_UNUSED(bufferInfo)
 
     QStringList params = msg.split(' ', QString::SkipEmptyParts);
-    // if the first argument is neither a channel nor us (user modes are only to oneself) the current buffer is assumed to be the target
     if (!params.isEmpty()) {
         if (params[0] == "-reset" && params.count() == 1) {
             network()->resetPersistentModes();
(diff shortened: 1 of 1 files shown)


src/core/coreuserinputhandler.cpp
```diff
@@ -483,7 +483,7 @@ void CoreUserInputHandler::handleMode(const BufferInfo &bufferInfo, const QStrin
             return;
         }
         if (!network()->isChannelName(params[0]) && !network()->isMyNick(params[0]))
-            params.prepend(bufferInfo.bufferName());
+            params.prepend(!bufferInfo.bufferName().isEmpty() ? bufferInfo.bufferName() : network()->myNick());
```
digitalcircuit: I'd suggest adding a comment here to explain what's happening, and wrapping to account for the 100-character line width.

For example (*you can write something different!*)...

```diff
         if (!network()->isChannelName(params[0]) && !network()->isMyNick(params[0]))
-            params.prepend(!bufferInfo.bufferName().isEmpty() ? bufferInfo.bufferName() : network()->myNick());
+            // Prepend the buffer name if available, if not (e.g. status buffer), add nick instead
+            params.prepend(!bufferInfo.bufferName().isEmpty() ?
+                               bufferInfo.bufferName() : network()->myNick());
         if (network()->isMyNick(params[0]) && params.count() == 2)
             network()->updateIssuedModes(params[1]);
```

genius3000: I noticed the previous comment a bit higher up that really belonged down here; moved it with a bit of re-wording and line wrapping, added a new comment and wrapped the `params.prepend()` as well. Thanks.

PR #391, opened 2018-06-20
client: Convert /list #chan to advanced search, auto-search, polish

digitalcircuit:
## In short

* Pass `/list <params>` to channel list dialog when opening
  * Allows for `/list #channel` to search
  * Restores behavior similar to current stable release
* Immediately search when called via `/list` and `/list <params>`
  * Mimics behavior of non-UI `/quote list`
* Polish up the experience
  * Add more keyboard friendliness with input focus
  * Sort channels in ascending order by default
* Prompt when `/list` called without valid network selected
  * Fixes `/list` before any network selected resulting in useless channel list dialog
  * Happens when `/list` after connecting to core without any connected networks

Criteria | Rank | Reason
Impact | ★★☆ *2/3* | User-facing regression fix, UI polish/keyboard readiness
Risk | ★☆☆ *1/3* | Channel list may not show at correct times
Intrusiveness | ★☆☆ *1/3* | Minor changes, shouldn't interfere with other PRs

## Examples
**CAUTION: If testing, do not use `Freenode`.  Use a network that has a smaller channel list.**

### Before
*`/list` and `/list #test` behave the same way*
Dialog shown when running '/list' on previous client, no channel results shown, 'Show Channels' must be clicked, sorting in descending channel name
*No content is shown, same as right-click network → `Show Channel List`, channel names Z-A*

### After
**Showing channel list via network right-click menu, keyboard shortcut**
*Behaves the same as before, no change*

**Showing channel list via `/list`**
Dialog shown when running '/list' on new client, channel results shown without any extra steps, sorted in ascending channel name
*`Channel List` dialog immediately shows channel listing, no clicking buttons required, channel names A-Z*

**Searching for a specific channel via `/list #test`**
Dialog shown when running '/list #test' on new client, specific channel result shown without any extra steps, sorted in ascending channel name
*`Channel List` dialog immediately shows specific channel listing, enabling advanced mode, no clicking buttons required, channel names A-Z*

### No network selected
*Connect to core that has no networks connected, so Quassel keeps the welcome logo visible.  Run `/list` without selecting a network.*
Dialog shown when running '/list' and no valid network is selected

> \[dialog box with title `No network selected`\]
>
> :information_source: **No network selected**
>
> Select a network before trying to view the channel list.
>
> \[OK button\]

### DIFF
--- a/src/client/client.h
+++ b/src/client/client.h
@@ -172,12 +172,14 @@ public:
      *
      * @see Client::showChannelList()
      *
-     * @param networkId       Network ID for associated network
-     * @param channelFilters  Partial channel name to search for, or empty to show all
+     * @param networkId        Network ID for associated network
+     * @param channelFilters   Partial channel name to search for, or empty to show all
+     * @param listImmediately  If true, immediately list channels, otherwise just show dialog
      */
-    void displayChannelList(NetworkId networkId, const QString &channelFilters = {})
+    void displayChannelList(NetworkId networkId, const QString &channelFilters = {},
+                            bool listImmediately = false)
     {
-        emit showChannelList(networkId, channelFilters);
+        emit showChannelList(networkId, channelFilters, listImmediately);
     }
 
 signals:
@@ -190,10 +192,13 @@ signals:
      *
      * @see MainWin::showChannelList()
      *
-     * @param networkId       Network ID for associated network
-     * @param channelFilters  Partial channel name to search for, or empty to show all
+     * @param networkId        Network ID for associated network
+     * @param channelFilters   Partial channel name to search for, or empty to show all
+     * @param listImmediately  If true, immediately list channels, otherwise just show dialog
      */
-    void showChannelList(NetworkId networkId, const QString &channelFilters = {});
+    void showChannelList(NetworkId networkId, const QString &channelFilters = {},
+                         bool listImmediately = false);
+
     void showIgnoreList(QString ignoreRule);
 
     void connected();
--- a/src/client/client.h
+++ b/src/client/client.h
@@ -167,15 +167,33 @@ public:
         emit showIgnoreList(ignoreRule);
     }
 
-    void displayChannelList(NetworkId networkId) {
-        emit showChannelList(networkId);
+    /**
+     * Request to show the channel list dialog for the network, optionally searching by channel name
+     *
+     * @see Client::showChannelList()
+     *
+     * @param networkId       Network ID for associated network
+     * @param channelFilters  Partial channel name to search for, or empty to show all
+     */
+    void displayChannelList(NetworkId networkId, const QString &channelFilters = {})
+    {
+        emit showChannelList(networkId, channelFilters);
     }
 
 signals:
     void requestNetworkStates();
 
     void showConfigWizard(const QVariantMap &coredata);
-    void showChannelList(NetworkId networkId);
+
+    /**
+     * Request to show the channel list dialog for the network, optionally searching by channel name
+     *
+     * @see MainWin::showChannelList()
+     *
+     * @param networkId       Network ID for associated network
+     * @param channelFilters  Partial channel name to search for, or empty to show all
+     */
+    void showChannelList(NetworkId networkId, const QString &channelFilters = {});
     void showIgnoreList(QString ignoreRule);
 
     void connected();
--- a/src/qtui/channellistdlg.h
+++ b/src/qtui/channellistdlg.h
@@ -66,6 +66,11 @@ private:
     void enableQuery(bool enable);
     void setAdvancedMode(bool advanced);
 
+    /**
+     * Update the focus of input widgets according to dialog state
+     */
+    void updateInputFocus();
+
     Ui::ChannelListDlg ui;
 
     bool _listFinished;
--- a/src/qtui/channellistdlg.h
+++ b/src/qtui/channellistdlg.h
@@ -50,8 +50,13 @@ public:
      */
     void setChannelFilters(const QString &channelFilters);
 
-protected slots:
+public slots:
+    /**
+     * Request a channel listing using any parameters set in the UI
+     */
     void requestSearch();
+
+protected slots:
     void receiveChannelList(const NetworkId &netId, const QStringList &channelFilters, const QList<IrcListHelper::ChannelDescription> &channelList);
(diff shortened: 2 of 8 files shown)


genius3000:
Gave this a quick test, results match with the above. I'll also add that search patterns of `>3 or <4` (user count) also work from a direct `/list >3`. I know Insp has another server side search pattern or two, didn't test those.

src/client/client.h
```diff
@@ -174,15 +174,37 @@ class Client : public QObject
 #endif
     static inline const QString &debugLog() { return instance()->_debugLogBuffer; }
 
-    void displayChannelList(NetworkId networkId) {
-        emit showChannelList(networkId);
+    /**
+     * Request to how the channel list dialog for the network, optionally searching by channel name
```
genius3000: Typo: how -> show

digitalcircuit: Thanks for catching this!  Fixed in latest push (upcoming).

digitalcircuit:
Thanks for testing, and reviewing!  Glad to know the multi-channel search feature works; I wasn't sure what network to try that on (*the networks I've tried only work with a single channel*).

digitalcircuit:
Scratch that, `<#` and `>#` actually works on the network I tested.  The more you know...  :smile:

src/client/client.h
```diff
@@ -174,15 +174,37 @@ class Client : public QObject
...
+    /**
+     * Request to show the channel list dialog for the network, optionally searching by channel name
+     *
+     * @see Client::showChannelList()
+     *
+     * @param networkId        Network ID for associated network
+     * @param channelFilters   Partial channel name to search for, or empty to show all
+     * @param listImmediately  If true, immediately list channels, otherwise just show dialog
+     */
+    void displayChannelList(NetworkId networkId, const QString channelFilters = QString(),
```
Sput42: const QString & (and instead of QString(), you can just use {} to indicate default value in C++11)

src/client/client.h
```diff
@@ -174,15 +174,37 @@ class Client : public QObject
...
+    /**
+     * Request to show the channel list dialog for the network, optionally searching by channel name
+     *
+     * @see MainWin::showChannelList()
+     *
+     * @param networkId        Network ID for associated network
+     * @param channelFilters   Partial channel name to search for, or empty to show all
+     * @param listImmediately  If true, immediately list channels, otherwise just show dialog
+     */
+    void showChannelList(NetworkId networkId, const QString channelFilters = QString(),
```
Sput42: Same here (& is missing)

src/qtui/channellistdlg.h
```diff
@@ -40,8 +40,23 @@ class ChannelListDlg : public QDialog
...
-protected slots:
+    /**
+     * Set the channel search string, enabling advanced mode if needed
+     *
+     * Sets the channel name search text to the specified string, enabling advanced mode.  If search
+     * string is empty, advanced mode will be automatically hidden.
+     *
+     * @param channelFilters Partial channel name to search for, or empty to not filter by name
+     */
+    void setChannelFilters(QString channelFilters);
```
Sput42: const QString &

src/qtui/mainwin.h
```diff
@@ -121,7 +121,17 @@ private slots:
...
-    void showChannelList(NetworkId netId = NetworkId());
+
+    /**
+     * Show the channel list dialog for the network, optionally searching by channel name
+     *
+     * @param networkId        Network ID for associated network
+     * @param channelFilters   Partial channel name to search for, or empty to show all
+     * @param listImmediately  If true, immediately list channels, otherwise just show dialog
+     */
+    void showChannelList(NetworkId netId = NetworkId(), QString channelFilters = QString(),
```
Sput42: const QString & (and {}, also instead of NetworkId())

src/uisupport/networkmodelcontroller.h
```diff
@@ -154,7 +154,17 @@ protected slots:
...
+    /**
+     * Request to show the channel list dialog for the network, optionally searching by channel name
+     *
+     * @see MainWin::showChannelList()
+     *
+     * @param networkId        Network ID for associated network
+     * @param channelFilters   Partial channel name to search for, or empty to show all
+     * @param listImmediately  If true, immediately list channels, otherwise just show dialog
+     */
+    void showChannelList(NetworkId, QString, bool);
```
Sput42: const QString &

digitalcircuit:
Thank you for the feedback!  I think I fixed all of these, and better still, this should help me learn modern practices.

A quick test shows stuff as still working, too, which is always a good thing after rebasing...

src/qtui/mainwin.cpp
```diff
@@ -206,8 +206,12 @@ void MainWin::init()
     connect(Client::instance(), SIGNAL(networkRemoved(NetworkId)), SLOT(clientNetworkRemoved(NetworkId)));
     connect(Client::messageModel(), SIGNAL(rowsInserted(const QModelIndex &, int, int)),
         SLOT(messagesInserted(const QModelIndex &, int, int)));
-    connect(GraphicalUi::contextMenuActionProvider(), SIGNAL(showChannelList(NetworkId)), SLOT(showChannelList(NetworkId)));
-    connect(Client::instance(), SIGNAL(showChannelList(NetworkId)), SLOT(showChannelList(NetworkId)));
+    connect(GraphicalUi::contextMenuActionProvider(),
+            SIGNAL(showChannelList(NetworkId, const QString &, bool)),
+            SLOT(showChannelList(NetworkId, const QString &, bool)));
```
Sput42: Just a remark: Old style SIGNAL/SLOT macros don't require cv qualifiers (const &); these are basically just parsed as a string. In fact, there's a normalization pass on the list of arguments that removes all such things and the spaces. This extra step is skipped if the given parameter list is already normalized. So the most efficient way here would be:
`SIGNAL(showChannelList(NetworkId,QString,bool))`

Don't bother too much, it's not consistent, and we're going to replace SIGNAL and SLOT by functor-based connections, which don't suffer from this problem (they don't take parameter lists at all). So take this as an information why we do things the way we do :)

Sput42:
Fixed merge conflicts, and also changed the signature of the connections mentioned above while I was at it.

digitalcircuit:
Thank you for tidying up those loose ends!

PR #393, opened 2018-06-20
logger: Refactor the logging framework

Sput42:
- Rename the previous Logger class to LogMessage (since that is what
      it is), and consolidate logging functionality, including the
      various output methods, setup of log file etc., into a new
      Logger class.
    
    - Hold the "debug log", i.e. the collected log messages, in the Logger
      class instead of in the Client. Adapt the DebugLogWidget accordingly.
      Messages will not be kept in quasselcore, as we wouldn't have a way
      to display them otherwise (and we don't want to accumulate lots of
      data over months of the core running).
    
    - Support early log messages. Instantiate the Logger right at the
      start of main(), and keep all log messages until the Quassel class
      is initialized and we can rely on command-line options to configure
      backends accordingly. Accumulated messages will be output once the
      Logger is configured.


Sput42:
@digitalcircuit: Hmm, shouldn't have been *that* broken really... well, please check if the newer version works for you. Certainly works for me...

genius3000:
Just ran a quick test on client only (core is ~1 day old master) and running via `./bin/quasselclient` and `./bin/quasselclient -d -L Debug` (built with type Debug), I would receive a single error message (but not prefixed with Quassel's log prefix) and then nothing else until I closed the client:
```
Hspell: can't open /usr/share/hspell/hebrew.wgz.sizes.
2018-06-21 16:56:54 [Warn ] HSpellDict::HSpellDict: Init failed
2018-06-21 16:56:54 [Debug] WARNING: bool Phonon::FactoryPrivate::createBackend() phonon backend plugin could not be loaded
2018-06-21 16:56:59 [Debug] Enabling compression...
2018-06-21 16:56:59 [Debug] Using the DataStream protocol...
2018-06-21 16:56:59 [Debug] Starting encryption...
```
Note the initial error, then nothing else print to stderr until closing the client (timestamps are correct to time of log). This seems to match @digitalcircuit's previous findings regarding the client.

src/common/logger.cpp
```diff
@@ -18,128 +18,239 @@
...
+    // Use signal connection to make this method thread-safe
+    emit messageLogged({QDateTime::currentDateTime(), level, logString += msg});
 }
+
+
+void Logger::onMessageLogged(const LogEntry &message)
+{
+    if (_keepMessages) {
+        _messages.push_back(message);
+    }
```
digitalcircuit: *Quoting the commit message...*
> (and we don't want to accumulate lots of data over months of the core running)

We already have issues with memory fragmentation and such that require restarting the client eventually.  Will this add anything significant to memory-usage, e.g. if a lot of debug messages get stored while the client's open for a long time?  Especially for monolithic as debug messages from both client and core are stored.

Would it make sense to add an upper bound based on estimated memory usage of a log message, and never store more than, say, 1 million messages, removing the oldest at that point?

Sput42: Nah, I'll rework logging to file to include the stderr case (we only log to stderr if we don't have file logging enabled anyway, so we can just as well just put the file on stderr and get rid of the special casing). And make that unbuffered.

I wouldn't be concerned about storing log size on the client, at least I hope we're not logging a million messages? On the core it's slightly more of an issue simply because there's potentially hundreds of users and database logs can be verbose, plus the uptime. But mostly I don't store on the core because there's no way to retrieve/view those logs anyway.

Compared to current implementation (which _already_ keeps all log messages), we actually save a bit of RAM, simply because we're now storing the timestamps as a QDateTime rather than a verbose string.

digitalcircuit: Noted on buffering!

Ack, I keep forgetting Quassel's existing implementation already kept every log message.  Thank you for clarifying yet again; this approach makes sense.

Sput42:
Yeah, so the issues you've seen are not related to buffering, but to some screw-up and refactoring remnants that would create two Logger instances. Logging through the standard Qt debug macros (rather than our own qu* versions) would go into the bogus, uninitialized instance, and thus only dumped on shutdown. Fixed this by removing the bogus instance, and also removed the useless instance ptr in the Logger class (also a refactoring artifact), instead going through Quassel::instance().

PR #395, opened 2018-06-22
common: Fix buffer search wrongly showing on upgrade

digitalcircuit:
## In short

* Fix initialization of default values when upgrading `BufferViewConfig`
  * Fixes wrongly enabling persistent buffer search on upgrade
  * Matches defaults for fresh configuration
  * Avoids cluttering the UI for upgrades, less likely to get complaints
  * Can still enable in the settings UI as before, or use keyboard shortcut
* Add more documentation

Criteria | Rank | Reason
Impact | ★★☆ *2/3* | Avoids user-facing unexpected UI changes on upgrade
Risk | ★☆☆ *1/3* | Possible wrong `BufferViewConfig` defaults, broken loading

*Polishes [changes made in pull request #190](https://github.com/quassel/quassel/pull/190).*

### Steps
1.  Set up Quassel core `0.12.5`
2.  Connect with a client, add a network
3.  Upgrade Quassel core to latest git commit
4.  Connect with a `0.13` client

### Before
`Show search` is wrongly activated on all buffer views.

### After
`Show search` is kept disabled, following the default from a fresh (non-upgrade) configuration.

### DIFF
--- a/src/common/bufferviewconfig.h
+++ b/src/common/bufferviewconfig.h
@@ -23,6 +23,7 @@
 
 #include "syncableobject.h"
 
+#include "bufferinfo.h"
 #include "types.h"
 
 class BufferViewConfig : public SyncableObject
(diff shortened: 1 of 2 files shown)


digitalcircuit:
*Small aside: @sandsmark , after mentioning this to you months ago, I finally figured it out!  Thanks again for getting the feature to a merge-able state in the first place :slightly_smiling_face:*

sandsmark:
could probably default initialize the rest of the members as well, though

src/common/bufferviewconfig.h
```diff
@@ -132,14 +133,18 @@ public slots:
     int _bufferViewId;
     QString _bufferViewName;
     NetworkId _networkId;
```
Sput42: Agreeing with @sandsmark here, this property is never initialized. Please just initialize to 0, while you're at it.

digitalcircuit: Noted!  Setting the `int` to 0, the others to `{}`, default constructor.

digitalcircuit:
Should be fixed now!

src/common/bufferviewconfig.h
```diff
@@ -129,17 +130,21 @@ public slots:
...
-    bool _addNewBuffersAutomatically;
-    bool _sortAlphabetically;
-    bool _hideInactiveBuffers;
-    bool _hideInactiveNetworks;
-    bool _disableDecoration;
-    int _allowedBufferTypes;
-    int _minimumActivity;
-    bool _showSearch;
+    int _bufferViewId = 0;
+    QString _bufferViewName = {};
```
Sput42: This wouldn't be necessary (QString is default-initialized to an empty string), but it doesn't really hurt, either.

digitalcircuit: This is specific to QString, right?  The original bug arose with `bool _showSearch` *not* default-initializing to `false`.. which is because it's a primitive and not an object-type, right?  Or is there something else at play here?

*Aside, I'll force-push this evening now that AppVeyor's fixed.  Might as well add the comments to these newly default-initialized variables, too.*

digitalcircuit:
Updated code comments, and now that AppVeyor's fixed (*thanks again to TheOneRing for all the maintenance!*), build is green, too :slightly_smiling_face:

PR #397, opened 2018-06-22
core: Fix highlight nick config save/load

digitalcircuit:
## In short

* Modernize `CoreHighlightRuleManager` via `SyncableObject` functions, following `DccConfig`
  * Fixes highlight nick configuration being reset on core/monolithic restart
  * Simplifies code
  * Fixes regression from [highlight protocol pull request #370](https://github.com/quassel/quassel/pull/370 )
* Save remote highlight rules if changed before importing local
  * Fixes losing unsaved rules when importing

Criteria | Rank | Reason
Impact | ★★★ *3/3* | User-facing common functionality - remembering highlight nick options
Risk | ★★☆ *2/3* | Core-side change, revertible with configuration loss

## Rationale
Persistence for **`Highlight Nicks`** is needed, for otherwise restarting the core **or the monolithic client** will reset nickname highlighting options, re-enabling if disabled and limiting to current nick if more had been chosen.

### Breaking changes
Anyone currently using the `master` branch will have to re-set-up their core-side highlight rules one more time (*sorry*).  Afterwards, settings persist as expected.

Upgrading from `0.12.5` or earlier works fine.

### Steps
1.  Set up core and client
2.  In `Configure Quassel…` → `Remote Highlights`, change **`Highlight Nicks`** to something non-default
    * E.g. `All Nicks from Identity`, `Case sensitive`
3.  Apply settings, shutdown core
4.  Launch core, reconnect
5.  Open configuration to `Remote Highlights` page

### Before
**`Highlight Nicks`** settings reset to default on every load.

### After
Even after core and client restarts, **`Highlight Nicks`** settings persist and have effect on core highlight behavior.

### DIFF
--- a/src/core/corehighlightrulemanager.cpp
+++ b/src/core/corehighlightrulemanager.cpp
@@ -23,35 +23,30 @@
 #include "core.h"
 #include "coresession.h"
 
+constexpr auto settingsKey = "HighlightRuleList";
+
 INIT_SYNCABLE_OBJECT(CoreHighlightRuleManager)
-CoreHighlightRuleManager::CoreHighlightRuleManager(CoreSession *parent)
-    : HighlightRuleManager(parent)
+CoreHighlightRuleManager::CoreHighlightRuleManager(CoreSession *session)
+    : HighlightRuleManager(session)
+    , _coreSession{session}
 {
-    CoreSession *session = qobject_cast<CoreSession*>(parent);
-    if (!session) {
-        qWarning() << "CoreHighlightRuleManager: unable to load HighlightRuleList. Parent is not a Coresession!";
-        //loadDefaults();
-        return;
-    }
+    // Load config from database if it exists
+    auto configMap = Core::getUserSetting(session->user(), settingsKey).toMap();
+    if (!configMap.isEmpty())
+        update(configMap);
+    // Otherwise, we just use the defaults initialized in the base class
 
-    initSetHighlightRuleList(Core::getUserSetting(session->user(), "HighlightRuleList").toMap());
-
-    // we store our settings whenever they change
+    // We store our settings whenever they change
     connect(this, SIGNAL(updatedRemotely()), SLOT(save()));
 }
 
-void CoreHighlightRuleManager::save() const
+void CoreHighlightRuleManager::save()
 {
-    CoreSession *session = qobject_cast<CoreSession *>(parent());
-    if (!session) {
-        qWarning() << "CoreHighlightRuleManager: unable to save HighlightRuleList. Parent is not a Coresession!";
-        return;
-    }
-
-    Core::setUserSetting(session->user(), "HighlightRuleList", initHighlightRuleList());
+    Core::setUserSetting(_coreSession->user(), settingsKey, toVariantMap());
 }
 
-bool CoreHighlightRuleManager::match(const RawMessage &msg, const QString &currentNick, const QStringList &identityNicks)
+bool CoreHighlightRuleManager::match(const RawMessage &msg, const QString &currentNick,
+                                     const QStringList &identityNicks)
 {
     return match(msg.text, msg.sender, msg.type, msg.flags, msg.target, currentNick, identityNicks);
 }
--- a/src/core/corehighlightrulemanager.h
+++ b/src/core/corehighlightrulemanager.h
@@ -25,6 +25,11 @@
 class CoreSession;
 struct RawMessage;
 
+/**
+ * Core-side specialization for HighlightRuleManager.
+ *
+ * Adds the ability to load/save the settings from/to the database.
+ */
 class CoreHighlightRuleManager : public HighlightRuleManager
 {
     Q_OBJECT
@@ -33,7 +38,12 @@ class CoreHighlightRuleManager : public HighlightRuleManager
     using HighlightRuleManager::match;
 
 public:
-    explicit CoreHighlightRuleManager(CoreSession *parent);
+    /**
+     * Constructor.
+     *
+     * @param[in] session Pointer to the parent CoreSession (takes ownership)
+     */
+    explicit CoreHighlightRuleManager(CoreSession *session);
 
     virtual const QMetaObject *syncMetaObject() const override { return &HighlightRuleManager::staticMetaObject; }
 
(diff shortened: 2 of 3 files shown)


digitalcircuit:
I did try the method from `CoreDccConfig`, of `toVariantMap()` and `update()`.  On upgrade, it didn't apply the right default settings (*picked `None` instead of `Current nick` despite header having initialized values*).  On a blank/fresh configuration, the core crashed.

I almost certainly was doing it wrong, just probably comes down to time to figure things out.

src/core/corehighlightrulemanager.cpp
```diff
@@ -34,24 +37,49 @@ CoreHighlightRuleManager::CoreHighlightRuleManager(CoreSession *parent)
...
+    setNicksCaseSensitive(highlightConfig["nicksCaseSensitive"].toBool());
+}
+
+
+QVariantMap CoreHighlightRuleManager::localInitExtraConfig() {
+    // TODO: Clean this up in a way that properly handles settings
+    QVariantMap highlightConfig;
+
+    highlightConfig["highlightNick"] = highlightNick();
+    highlightConfig["nicksCaseSensitive"] = nicksCaseSensitive();
```
Sput42: I would appreciate keys in maps starting with capital letter. I know we're not consistent everywhere (CoreInfo comes to mind), but where we can change it without breaking compat, or when introducing new maps, we should do it right.

src/core/corehighlightrulemanager.cpp
```diff
@@ -34,24 +37,49 @@ CoreHighlightRuleManager::CoreHighlightRuleManager(CoreSession *parent)
...
 }
 
 bool CoreHighlightRuleManager::match(const RawMessage &msg, const QString &currentNick, const QStringList &identityNicks)
 {
     return match(msg.text, msg.sender, msg.type, msg.flags, msg.target, currentNick, identityNicks);
 }
+
+
+void CoreHighlightRuleManager::localInitSetExtraConfig(const QVariantMap &highlightConfig) {
+    // TODO: Clean this up in a way that properly handles settings
```
Sput42: I'm not quite sure I understand the comment...

src/core/corehighlightrulemanager.h
```diff
@@ -50,5 +50,21 @@ public slots:
...
+     * @param highlightConfig Additional highlight configuration as a variant map
+     */
+    void localInitSetExtraConfig(const QVariantMap &highlightConfig);
+
+    /**
+     * Get additional configuration not contained within highlight rules
+     *
+     * @return Additional highlight configuration as a variant map
+     */
+    QVariantMap localInitExtraConfig();
```
Sput42: Both these methods are very confusingly named IMHO. The init* and initSet* methods have a very specific purpose in the context of syncable objects, but for "local" setters we should just be using normal names such as setHighlightConfig() or whatever. Maybe I'm not understanding correctly what you plan to do in the final version, though?

src/core/corehighlightrulemanager.cpp
```diff
@@ -34,24 +37,49 @@ CoreHighlightRuleManager::CoreHighlightRuleManager(CoreSession *parent)
...
+    Core::setUserSetting(session->user(), settingsKeyRules, initHighlightRuleList());
 }
 
 bool CoreHighlightRuleManager::match(const RawMessage &msg, const QString &currentNick, const QStringList &identityNicks)
 {
     return match(msg.text, msg.sender, msg.type, msg.flags, msg.target, currentNick, identityNicks);
 }
+
+
+void CoreHighlightRuleManager::localInitSetExtraConfig(const QVariantMap &highlightConfig) {
```
Sput42: Linebreak

src/core/corehighlightrulemanager.cpp
```diff
@@ -34,24 +37,49 @@ CoreHighlightRuleManager::CoreHighlightRuleManager(CoreSession *parent)
...
+    // TODO: Clean this up in a way that properly handles settings
+
+    // Make sure the default for _highlightNick is "CurrentNick" if not set
+    setHighlightNick(HighlightNickType(
+                         highlightConfig.value("highlightNick",
+                                               HighlightNickType::CurrentNick).toInt()));
```
Sput42: Is this weirdly indented, or does the Github UI fool me?

src/core/corehighlightrulemanager.cpp
```diff
@@ -34,24 +37,49 @@ CoreHighlightRuleManager::CoreHighlightRuleManager(CoreSession *parent)
...
+
+    setNicksCaseSensitive(highlightConfig["nicksCaseSensitive"].toBool());
+}
+
+
+QVariantMap CoreHighlightRuleManager::localInitExtraConfig() {
```
Sput42: Linebreak

src/core/corehighlightrulemanager.cpp
Sput42: I retract this request, I now understand that you plan to initialize the properties from the map using fromVariantMap(). In this case, the keys need to match the property names exactly, so lowerCamelCase is ok here.

digitalcircuit:
Thank you for the comments and review!

As discussed on IRC, this was more of a hackish "get something working for a beta deadline" approach in case `update()` for loading/`toVariantMap()` for saving can't be used [as it is in `CoreDccConfig`](https://github.com/quassel/quassel/blob/master/src/core/coredccconfig.cpp ).  My initial efforts looking into that resulted in crashes, but I'll try again.

As to why?  It'd be a lot cleaner to get rid of the manual `initSet`/`init` calls since the `update`/`toVariantMap` should automatically call them *and* automatically include all `Q_PROPERTY` items.  I'll have to test upgrading and clean install configuration to make sure the right default settings are being used.

I'll keep in mind the corrections you provided where applicable, and should we need a separate save/load method for nick configuration after all, I'll certainly apply the rest of the corrections to this, too.

digitalcircuit:
I've reworked this with the `update()`/`toVariantMap()` options, and everything on Linux seems to work!

In all cases below, I started/set up core, connected client, configured, shut down core, then started up core, checked settings persisted.

* Fresh configuration: **pass**
* Upgrade from faux-`0.12.5`: **pass**
  * I just deleted `HighlightRuleList` from the `user_setting` table on a `0.13` build
* Upgrade from `master`: **pass**
  * All existing core-side highlights are deleted, just this once

More extensive testing could be done, but this seems to at least make the situation better in the basic tests I used.

PR #398, opened 2018-07-12
Change the timer method for /LIST to handle large lists better

genius3000:
The current timer of 10 seconds from starting the channel list query can timeout with larger lists or strict rate-limiting. This will cause an early 'end of list' which populates the dialog in the client with what was received up till then and the client dumps the rest to the Status Buffer. The timer also isn't killed when a proper 'end of list' is received and a new query ran quickly after another can be timed out by a previous timer.  
I found `QBasicTimer` to be a better fit here as calling `start(ms)` will restart the timer if it's already running. This allows us to restart the timer after each call to `addChannel()`, letting the query take a little longer if needed but not timing out while still receiving data. The timer is now removed when a proper 'end of list' is received and a timeout will still call 'end of list'.

I found starting the timer at 5 seconds and restarting each time at 5 seconds to be a good values as rate limiting or "fakelag" can sometimes add multiple seconds to replies from the server.
#### Testing
* Small channel lists: quick retrieval of list as expected; timer is started, restarted, stopped correctly.
* Faked large channel list: retrieval of list takes a bit but doesn't over-run to the Status Buffer; timer is started, restarted, stopped correctly.
* Freenode's large channel list: again, retrieval takes a bit but doesn't over-run; timer is started, restarted, stopped correctly.

Fixes #1471

### DIFF
--- a/src/core/coreirclisthelper.cpp
+++ b/src/core/coreirclisthelper.cpp
@@ -23,6 +23,8 @@
 #include "corenetwork.h"
 #include "coreuserinputhandler.h"
 
+constexpr auto kTimeoutMs = 5000;
+
 INIT_SYNCABLE_OBJECT(CoreIrcListHelper)
 QVariantList CoreIrcListHelper::requestChannelList(const NetworkId &netId, const QStringList &channelFilters)
 {
@@ -45,6 +47,9 @@ bool CoreIrcListHelper::addChannel(const NetworkId &netId, const QString &channe
         return false;
 
     _channelLists[netId] << ChannelDescription(channelName, userCount, topic);
+    if (_queryTimeoutByNetId.contains(netId))
+        _queryTimeoutByNetId[netId]->start(kTimeoutMs, this);
+
     return true;
 }
 
@@ -55,7 +60,12 @@ bool CoreIrcListHelper::dispatchQuery(const NetworkId &netId, const QString &que
     if (network) {
         _channelLists[netId] = QList<ChannelDescription>();
         network->userInputHandler()->handleList(BufferInfo(), query);
-        _queryTimeout[startTimer(10000)] = netId;
+
+        auto timer = std::make_shared<QBasicTimer>();
+        timer->start(kTimeoutMs, this);
+        _queryTimeoutByNetId[netId] = timer;
+        _queryTimeoutByTimerId[timer->timerId()] = netId;
+
         return true;
     }
     else {
@@ -66,6 +76,12 @@ bool CoreIrcListHelper::dispatchQuery(const NetworkId &netId, const QString &que
 
 bool CoreIrcListHelper::endOfChannelList(const NetworkId &netId)
 {
+    if (_queryTimeoutByNetId.contains(netId)) {
+        // If we recieved an actual RPL_LISTEND, remove the timer
+        int timerId = _queryTimeoutByNetId.take(netId)->timerId();
+        _queryTimeoutByTimerId.remove(timerId);
+    }
+
     if (_queuedQuery.contains(netId)) {
         // we're no longer interessted in the current data. drop it and issue a new request.
         return dispatchQuery(netId, _queuedQuery.take(netId));
@@ -92,8 +108,14 @@ bool CoreIrcListHelper::endOfChannelList(const NetworkId &netId)
 
 void CoreIrcListHelper::timerEvent(QTimerEvent *event)
 {
-    int timerId = event->timerId();
-    killTimer(timerId);
-    NetworkId netId = _queryTimeout.take(timerId);
+    if (!_queryTimeoutByTimerId.contains(event->timerId())) {
+        IrcListHelper::timerEvent(event);
+        return;
+    }
+
+    NetworkId netId = _queryTimeoutByTimerId.take(event->timerId());
+    _queryTimeoutByNetId.remove(netId);
+
+    event->accept();
     endOfChannelList(netId);
 }
--- a/src/core/coreirclisthelper.h
+++ b/src/core/coreirclisthelper.h
@@ -21,10 +21,13 @@
 #ifndef COREIRCLISTHELPER_H
 #define COREIRCLISTHELPER_H
 
+#include <memory>
+
 #include "irclisthelper.h"
 
 #include "coresession.h"
 
+class QBasicTimer;
 class QTimerEvent;
 
 class CoreIrcListHelper : public IrcListHelper
@@ -58,7 +61,8 @@ private:
     QHash<NetworkId, QString> _queuedQuery;
     QHash<NetworkId, QList<ChannelDescription> > _channelLists;
     QHash<NetworkId, QVariantList> _finishedChannelLists;
-    QHash<int, NetworkId> _queryTimeout;
+    QHash<int, NetworkId> _queryTimeoutByTimerId;
+    QHash<NetworkId, std::shared_ptr<QBasicTimer>> _queryTimeoutByNetId;
 };
 
 


src/core/coreirclisthelper.h
```diff
@@ -25,8 +25,6 @@
 
 #include "coresession.h"
 
-class QTimerEvent;
```
Sput42: You should probably not remove this, because you're still using the type in the signature for timerEvent()...

src/core/coreirclisthelper.h
```diff
@@ -58,7 +56,8 @@ public slots:
     QHash<NetworkId, QString> _queuedQuery;
     QHash<NetworkId, QList<ChannelDescription> > _channelLists;
     QHash<NetworkId, QVariantList> _finishedChannelLists;
-    QHash<int, NetworkId> _queryTimeout;
+    QHash<int, NetworkId> _queryTimeoutByTimerId;
+    QHash<NetworkId, QBasicTimer*> _queryTimeoutByNetId;
```
Sput42: Since you're using this type here, you should also forward-declare it.

src/core/coreirclisthelper.cpp
```diff
@@ -55,7 +59,12 @@ bool CoreIrcListHelper::dispatchQuery(const NetworkId &netId, const QString &que
     if (network) {
         _channelLists[netId] = QList<ChannelDescription>();
         network->userInputHandler()->handleList(BufferInfo(), query);
-        _queryTimeout[startTimer(10000)] = netId;
+
+        QBasicTimer *timer = new QBasicTimer;
```
Sput42: This is scary, because it might leak (e.g. when CoreIrcListHelper is destroyed while there are still active timers). I would suggest using a std::shared_ptr, which would also make some of your logic simpler (i.e. you can then just erase the shared_ptr from your hashes, no need to explicitly stop and delete the QBasicTimer instance).

src/core/coreirclisthelper.h
genius3000: I did wonder about this, but the docs for QBasicTimer don't mention/imply needing to include this class nor the example in the docs which uses the same timerEvent() function. Looking at `coreuserinputhandler`, it's using the same timerEvent() without this `class QTimerEvent` line; though I also see I forgot to include the passing of the Event if it doesn't belong to CoreIrcListHelper. I'm not against leaving it for safety or properness, just trying to not leave remnants of previous code behind.

src/core/coreirclisthelper.cpp
genius3000: Valid concern, I will need to do some reading/grepping and then implement your advice.

genius3000:
Thanks for the review and requests, I'll dig into it tomorrow. I've never used std::shared_ptr before so any pointers to a good reference/doc would be greatly appreciated; though I'll read up on cppreference and Google as well :relaxed:

src/core/coreirclisthelper.cpp
genius3000: Is this implementation valid:
**(header)**
`typedef std::shared_ptr<QBasicTimer> _queryTimeout;`
then use `_queryTimeout` in the hash.
**(source)**
`_queryTimeout timer(new QBasicTimer)` then reference with `_queryTimeout timer = <hash>.take()` and so forth?

Current usage of shared_ptr in Quassel is minimal and a bit different so I went with the more standard looking implementation from cppreference and some examples out in the wild. It does compile and appear to work...I'm not sure how to test the CoreIrcListHelper class being destroyed though.

src/core/coreirclisthelper.h
Sput42: Well, usually the necessary includes are not specifically documented in Qt. And most of the time, things just work through transitive includes (e.g. QObject including QBasicTimer). However, that is bad practice, and we have to fix issues due to this on a regular basis, because a new Qt version removes a public include...

So, for new code I'd prefer this to be done properly, such that all the types used are also directly included or forward-declared. The legacy code we'll update as we go; you may have noticed that I keep adding includes quite often in code I touch :)

src/core/coreirclisthelper.cpp
Sput42: This looks weird and ugly :) Especially the typedef thing. Just use `QHash<int, std::shared_ptr<QBasicTimer>> _queryTimerById` etc. in the header. In the source you can then just do things like `_queryTimerById[id]->start()`, and if you want to delete it, just remove/take it from the hash. No need to explicitly test for destruction then, because std::shared_ptr guarantees to delete its pointee as soon as the last reference to it is gone (that's the whole point of using this instead of raw pointers). So if the hash is destroyed, all its items are too.

Sput42: To clarify a bit more: Instead of using `QBasicTimer *`, you use `std::shared_ptr<QBasicTimer>`, which can be used exactly like a raw pointer, only that you don't need to delete it. And for creating the instance, instead of `new QBasicTimer` you use `std::make_shared<QBasicTimer>()`.

src/core/coreirclisthelper.h
genius3000: Understood, thanks for the explanation.

src/core/coreirclisthelper.cpp
genius3000: Ah that does sound simpler and cleaner, thank you again for the advice and explanation.

genius3000:
Updated with the requested changes and re-tested successfully.

src/core/coreirclisthelper.h
```diff
@@ -25,6 +25,7 @@
 
 #include "coresession.h"
```
Sput42: We're missing an `#include <memory>` for the shared_ptr.

src/core/coreirclisthelper.cpp
```diff
@@ -55,7 +58,11 @@ bool CoreIrcListHelper::dispatchQuery(const NetworkId &netId, const QString &que
     if (network) {
         _channelLists[netId] = QList<ChannelDescription>();
         network->userInputHandler()->handleList(BufferInfo(), query);
-        _queryTimeout[startTimer(10000)] = netId;
+
+        _queryTimeoutByNetId[netId] = std::make_shared<QBasicTimer>();
+        _queryTimeoutByNetId[netId]->start(5000, this);
+        _queryTimeoutByTimerId[_queryTimeoutByNetId[netId]->timerId()] = netId;
```
Sput42: It would probably make sense to store the new pointer in a local variable `auto timer = std::make_shared<QBasicTimer>()` and use that to avoid the repeated lookups in `_queryTimeoutByNetId`.

src/core/coreirclisthelper.cpp
```diff
@@ -92,8 +105,13 @@ bool CoreIrcListHelper::endOfChannelList(const NetworkId &netId)
 
 void CoreIrcListHelper::timerEvent(QTimerEvent *event)
 {
-    int timerId = event->timerId();
-    killTimer(timerId);
-    NetworkId netId = _queryTimeout.take(timerId);
+    if (!_queryTimeoutByTimerId.contains(event->timerId())) {
+        QObject::timerEvent(event);
```
Sput42: You should call the direct base here, i.e. `IrcEventHelper::timerEvent()`, just in case someone adds a timer there at some point.

src/core/coreirclisthelper.cpp
```diff
@@ -92,8 +105,13 @@ bool CoreIrcListHelper::endOfChannelList(const NetworkId &netId)
...
+        return;
+    }
+
+    NetworkId netId = _queryTimeoutByTimerId.take(event->timerId());
+    _queryTimeoutByNetId.remove(netId);
+
```
Sput42: Since we've handled the event, we should call `event->accept()`. This stops it from being propagated to other potential receivers. I don't think it really matters in this particular case, but it's good practice.

src/core/coreirclisthelper.cpp
```diff
@@ -45,6 +45,9 @@ bool CoreIrcListHelper::addChannel(const NetworkId &netId, const QString &channe
         return false;
 
     _channelLists[netId] << ChannelDescription(channelName, userCount, topic);
+    if (_queryTimeoutByNetId.contains(netId))
+        _queryTimeoutByNetId[netId]->start(5000, this);
```
Sput42: Just as a suggestion, I'd make the 5000 a constant at the top of the file, after the includes, such as `auto constexpr kTimeoutMs = 5000`. You're reusing it further down too, so it's nice to have the "magic number" only defined in one place.

src/core/coreirclisthelper.cpp
```diff
@@ -55,7 +58,11 @@ bool CoreIrcListHelper::dispatchQuery(const NetworkId &netId, const QString &que
     if (network) {
         _channelLists[netId] = QList<ChannelDescription>();
         network->userInputHandler()->handleList(BufferInfo(), query);
-        _queryTimeout[startTimer(10000)] = netId;
+
+        _queryTimeoutByNetId[netId] = std::make_shared<QBasicTimer>();
+        _queryTimeoutByNetId[netId]->start(5000, this);
```
Sput42: See above

src/core/coreirclisthelper.h
genius3000: Should I organize/combine the two current includes? I looked through some files for reference of structure and mostly see "std includes, Qt includes, local includes" but also some that separate local includes into "very local" and "quassel local" sort of thing (if that makes sense lol).

genius3000:
Updated as per requests and re-based (no conflicts but best to). I left the other 'include' lines alone for now.  
Thank you again for the review and pointers.

src/core/coreirclisthelper.h
Sput42: I'm still struggling to make this consistent. Right now, my approach is to have the include for the header belonging to a cpp as the very first, then system/std headers, Qt headers, Quassel headers. I'm still debating with myself if the order of the latter should be reversed, i.e. first Quassel, then Qt, then system/std. Seems like both ways have their pros and cons. Google I think prefers the first, but having system headers last may reduce accidental transitive includes...

When we do the big cleanup, I'll come up with a guideline and document it somewhere.

PR #399, opened 2018-07-15
Improve handling of core initialization errors

Sput42:
Since the event loop may not be running while the core is being
initialized, calling exit() does not have the desired effect.
Instead, throw an exception with exit code and error message to
be able to handle this correctly. Exceptions must not be thrown
through the event loop, so ensure that they're caught before in
cases where the event loop was started already.

For the monolithic client, introduce an asynchronous init method
that catches potential exceptions and calls exit() instead. Show
an error popup in the UI if an error message is provided, so the
user gets some feedback before the client is terminated.

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -33,6 +33,7 @@
 #include "quassel.h"
 #include "sqlauthenticator.h"
 #include "sqlitestorage.h"
+#include "types.h"
 #include "util.h"
 
 // Currently building with LDAP bindings is optional.
@@ -100,7 +101,7 @@ Core::~Core()
 }
 
 
-bool Core::init()
+void Core::init()
 {
     _startTime = QDateTime::currentDateTime().toUTC(); // for uptime :)
 
@@ -112,9 +113,7 @@ bool Core::init()
     // so far, we only have 1
     CoreSettings s;
     if (s.version() != 1) {
-        qCritical() << "Invalid core settings version, terminating!";
-        QCoreApplication::exit(EXIT_FAILURE);
-        return false;
+        throw ExitException{EXIT_FAILURE, tr("Invalid core settings version!")};
     }
 
     // Set up storage and authentication backends
@@ -135,7 +134,8 @@ bool Core::init()
     if (config_from_environment) {
         db_backend = environment.value("DB_BACKEND");
         auth_authenticator = environment.value("AUTH_AUTHENTICATOR");
-    } else {
+    }
+    else {
         CoreSettings cs;
 
         QVariantMap dbsettings = cs.storageSettings().toMap();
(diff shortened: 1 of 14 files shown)


genius3000:
I haven't done a **full test**, just a successful test of a few command line options on the core:
* `--select-backend <backend>`
* `--add-user`
* `--change-userpass`

The core properly exits after success or failure of the above 'short run' options.
No issues noted with normal starting and use of the core.

Thanks again for taking care of this @Sput42

src/core/core.cpp
```diff
@@ -165,36 +168,37 @@ bool Core::init()
...
+                if (_configured) {
+                    _configured = initAuthenticator(auth_authenticator, auth_properties, environment, config_from_environment, true);
+                }
+            }
+            catch (ExitException) {
+                _configured = false;
+            }
 
             if (!_configured) {
                 qWarning() << "Cannot configure from environment";
```
genius3000: Missed removal?

Sput42: Worked as expected, but I'm changing it anyway to add the actual error message returned from the backend. This was written when I didn't have an errorString in the exception :) Thanks for noticing.

digitalcircuit:
In a quick test, this works for me.

* Core and monolithic start and work correctly
* `--select-backend SQLite` with SQLite already used
  * Properly sticks to SQLite, exit code of `0`
* `--select-backend PostgreSQL` and configuring incorrectly
  * Exit code of `1`
* Monolithic, modifying `Core::init()` to introduce a fake exception
  * `throw ExitException{EXIT_FAILURE, tr("I don't feel like working!")};`

> **`Quassel encountered a fatal error and is terminated.`**
> `Reason:` *`I don't feel like working!`*

**Note:  The core (not monolithic) doesn't appear to print or log this message anywhere.**  I suspect this is before logging is set up, and logging setup never gets a chance to finish, so it's not possible..?

*For whatever reason, I had to completely empty my build directory due to the use of exceptions, but the build succeeded after that.*

digitalcircuit:
Aha!

`Logger::handleMessage(LogLevel, QString)` depends on the Qt event loop running.  Given `QApplication::exit()` isn't working due to no event loop, this won't work, either.

```cpp
void Logger::handleMessage(LogLevel level, const QString &msg)
{
    QString logString;
    switch (level) {
[...]
    }
    // Use signal connection to make this method thread-safe
    emit messageLogged({QDateTime::currentDateTime(), level, logString += msg});
}
```

`messageLogged` is connected via SIGNAL/SLOT in the initialization…
```cpp
connect(this, SIGNAL(messageLogged(Logger::LogEntry)), this, SLOT(onMessageLogged(Logger::LogEntry)));
```

…so `Logger::onMessageLogged(LogEntry)` never gets called to push messages onto the stack.

Perhaps you need a fallback method to detect when the Qt event loop is not running, and directly call `onMessageLogged()`?

You could create a single-shot event in the logger class to flip a boolean that indicates events are ready, and check that boolean when emitting the signal to either call directly or emit.

See also: https://stackoverflow.com/questions/8876390/qt-is-there-notification-when-event-loop-starts

digitalcircuit:
...okay, nevermind the above.  Somehow the signals do appear to work from the `Core::init()` phase despite never reaching `app.exec();`.

I think my issue was looking at the wrong log file - d'oh.  I was fooled by seeing other recent events due to switching around which profiles I used for testing.  Apologies for the hassle!

Still, it might be worthwhile testing an exception in `Core::init()` if you haven't already.

PR #402, opened 2018-07-16
Post-0.13 modernization

Sput42:
Modernize the code base after 0.13 is released.

- Improve CI support and add new CI targets (Xenial, Bionic with and without KDE Frameworks integration)
- Remove support for Qt4/KDE4
- Remove support for obsolete dependencies: libindicate, Phonon, miniz
- Bump dependency versions: Qt >= 5.5, CMake >= 3.5.1
- Require new dependency: Boost headers
- Clean up and modernize initialization sequence, translation loading, command line parsing
- Modernize CMake build system
- Remove custom CMake modules where possible
- Build shared libraries for the Quassel modules
- Modernize the codebase using clang-tidy checks
- Modernize Qt usage using clazy checks
- Remove almost all uses of old-style connects and the SIGNAL/SLOT macros
- Reformat ALL the source
- Introduce a unit testing framework
- Start refactoring the SignalProxy and SyncableObject to make use of modern language and Qt features and avoid undocumented Qt hacks


CMakeLists.txt
```diff
@@ -386,14 +386,13 @@ endif()
 # Non-Qt-based packages
 
 # zlib for compression, however we can always fall back to miniz
```
digitalcircuit: You might have missed this comment in your cleanup…

```diff
- # zlib for compression, however we can always fall back to miniz
+ # zlib for compression
```

Sput42: Indeed, thanks for noticing!

Sput42: Will just remove the whole comment, as the purpose is already stated in set_package_properties,

cmake/QuasselMacros.cmake
```diff
@@ -43,6 +45,113 @@ function(quassel_add_module _module)
...
+    )
+    set_source_files_properties(${qrc_filepath} PROPERTIES SKIP_AUTORCC ON)
+
+    # RCC sucks and expects the data files relative to the qrc file, with no way to configure it differently.
+    # Only when reading from stdin ("-") it takes the working directory as a base, so we have to use this if
+    # we want to use generated qrc files (which obviously cannot be placed in the source directory).
+    # Since neither autorcc nor qt5_add_resources() support this, we have to invoke rcc manually :(
+    #
+    # On Windows, input redirection apparently doesn't work, however piping does. Use this for all platforms for
+    # consistency, accomodating for the fact that the 'cat' equivalent on Windows is 'type'.
```
digitalcircuit: **Minor:** `accomodating` → `accommodating`

Also: oof.  Not fun to debug, glad you added a comment so the rest of us can learn from your pain :slightly_smiling_face:

Sput42: Wouldn't have figured out the cat/type thing without @TheOneRing...

src/common/expressionmatch.cpp
```diff
@@ -68,31 +62,17 @@ bool ExpressionMatch::match(const QString &string, bool matchEmpty) const
         // Check inverted match rule
 
         // See _matchRegEx section below for explanations of QRegExp vs. QRegularExpression
```
digitalcircuit: **Minor:** Missed this comment.

src/client/clientsettings.cpp
```diff
@@ -25,6 +25,7 @@
 #include <QHostAddress>
 #ifdef HAVE_SSL
 #include <QSslSocket>
+#include <utility>
```
digitalcircuit: Did you mean to include this only within the `#ifdef HAVE_SSL` block?

src/common/signalproxy.cpp
```diff
@@ -18,6 +18,8 @@
  *   51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.         *
  ***************************************************************************/
 
+#include <utility>
```
digitalcircuit: **Minor:** The `#include <utility>` line sometimes appears above Qt includes, sometimes below.  Will this be fixed up with later auto-formatting?

src/qtui/chatscene.cpp
```diff
@@ -33,6 +33,7 @@
 
 #ifdef HAVE_WEBENGINE
 #  include <QWebEngineView>
+#include <utility>
```
digitalcircuit: Did you mean to include this only within the `#ifdef HAVE_WEBENGINE` block?

src/client/bufferviewoverlay.h
```diff
@@ -75,14 +75,14 @@ private slots:
     void updateHelper();
     QSet<BufferId> filterBuffersByConfig(const QList<BufferId> &buffers, const BufferViewConfig *config);
 
-    bool _aboutToUpdate;
+    bool _aboutToUpdate{false};
```
digitalcircuit: **Not an issue, just curiosity:**  Why…
```cpp
bool _aboutToUpdate{false};
```
…instead of…
```cpp
bool _aboutToUpdate = false;
```
Is it to better express intent of default construction, or behavioral changes, too?

Sput42: I think both ways are semantically identical. Not completely sure about this (outside of headers, I believe that equals is calling the copy ctor, while the other is calling the normal ctor), however I didn't manage to find confirmation anywhere. The standard used to call it "brace-or-equal initializer" which suggests them being identical. Unlike writing = vs {} in "normal" code, using it in headers doesn't directly translate to code being executed; it's telling the compiler what to initialize the value with if there isn't another initializer in the ctor itself.

Main reason to do it the way it is here is that clang-tidy does it that way :-)

src/common/signalproxy.cpp
Sput42: Yeah, include order will be made consistent once we reformat the codebase. clang-tidy adds it everywhere, and sometimes gets it wrong...

src/qtui/chatscene.cpp
Sput42: Fixed!

src/common/funchelpers.h
```diff
@@ -62,3 +70,51 @@ struct FuncHelper<R(C::*)(Args...) const> : public FuncHelper<R(C::*)(Args...)>
...
+ * @returns true if the callable could be invoked with the given list of arguments
+ */
+template<typename Callable>
+bool invokeWithArgsList(const Callable& c, const QVariantList& args)
+{
+    using ArgsTuple = typename FunctionTraits<Callable>::ArgsTuple;
+    constexpr auto tupleSize = std::tuple_size<ArgsTuple>::value;
+
+    if (tupleSize != args.size()) {
+        qWarning().nospace() << "Argument count mismatch! Expected: " << tupleSize << ", actual: " << args.size();
```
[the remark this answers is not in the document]
Sput42: Wellll.... it does give you useful information, however it looks like this:

> [ RUN      ] FuncHelpersTest.invokeWithArgsList
> Argument count mismatch! Expected: 2, actual: 3
> Argument count mismatch! Expected: 2, actual: 1
>bool detail::invokeWithArgsList(const Callable&, const QVariantList&, std::index_sequence<Is ...>) [with Callable = FuncHelpersTest_invokeWithArgsList_Test::TestBody()::<lambda(int, const QString&)>; long unsigned int ...Is = {0, 1}; ArgsTuple = std::tuple<int, const QString&>; QVariantList = QList<QVariant>; std::index_sequence<Is ...> = std::integer_sequence<long unsigned int, 0, 1>] Cannot convert parameter 1 from type QVariantList to expected argument type
> [       OK ] FuncHelpersTest.invokeWithArgsList (0 ms)

We really only want the ArgsTuple types, but I don't know how to easily print them explicitly without resorting to manual demangling of typeid()... and not sure if it's worth it. In practice, SignalProxy will log the affected signal name immediately afterwards anyway.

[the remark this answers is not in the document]
digitalcircuit: Oof, that function information is pretty messy.  I agree, not a good idea to include it.

As long as outside of test cases there's more information printed at the same timestamp, as you noted, that should be fine.  I was merely concerned we might get reports from people where the only debugging message printed is `Argument count mismatch! Expected: 2, actual: 3` :slightly_smiling_face:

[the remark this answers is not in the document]
Sput42: i hope we won't get any of those messages at all, because that would be an actual bug (as opposed to unknown properties or whatever which may be caused by old/new client/core)...

Sput42:
Just testing if BitBot announcements work...

PR #406, opened 2018-07-24
Stylesheet support for hovered-upon URLs; disable color codes in highlighted messages

Sput42:
- A new message label "hovered" can now be used in stylesheets to control how a hovered-upon URLs should be rendered. The default stylesheets defines an underline.
- Fix rendering of selected, active clickable. Fixes the issue where a selected URL in a highlighted message would vanished when hovered upon.
- Disable color codes in highlighted messages in the default stylesheet.

### DIFF
--- a/data/stylesheets/default.qss
+++ b/data/stylesheets/default.qss
@@ -14,6 +14,7 @@ ChatLine {
 ChatLine[label="highlight"] {
   foreground: black;
   background: #ff8000;
+  allow-foreground-override: false;
   allow-background-override: false;
 }
 
--- a/data/stylesheets/default.qss
+++ b/data/stylesheets/default.qss
@@ -25,6 +25,10 @@ ChatLine[label="selected"] {
   allow-background-override: false;
 }
 
+ChatLine[label="hovered"] {
+  font-style: underline;
+}
+
 // ChatLine::sender[sender="self"] {
 //   font-style: italic;
 // }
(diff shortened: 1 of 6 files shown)


Sput42:
Wow, that's a fun one. It seems to depend on which characters are actually selected; which is probably why I didn't see this in my (basic) testing, where I of course tested partial select... investigating.

Sput42:
No, it's easily reproducible with other strings. I'm also almost convinced it's a bug in Qt's own rendering, because the formats we give it are correct in both cases - just moving the selection by one character to the left or right will do the correct thing with additionalFormats() which Qt is supposed to overlay over the normal formats, but the line won't be drawn anymore. Another artifact I already found before pushing the patchset was a one-pixel-gap between a selected space preceding the URL, and the URL itself. Also that one sometimes shows up and sometimes doesn't, and it's clear that the input data is the same...

Will have to see if I just do the format merging in Quassel rather than using the Qt feature, or live with the glitch.

Sput42:
So I tried fixing this issue by adding the selection/hover formats to the regular format list of the layout, since additionalFormats() doesn't seem to render the underline always. This almost worked, but since we have to relayout afterwards, some letter combinations will slightly change sometimes, and the text "wobbles" while selecting it. So I'd rather live with partially selected URLs not properly working.

digitalcircuit:
That approach makes sense to me.  Thank you for investigating further!  And if someone else reports the bug, well, we have this to show :slightly_smiling_face:

PR #357, opened 2018-05-08
[Rollup] common: Master of database/protocol PRs

digitalcircuit:
## In short

* **A:** Persist Blowfish keys in the database, [pull request #332](https://github.com/quassel/quassel/pull/332 )
  * Keeps encrypted chats working after core restart/etc
  * Modernizing Quassel's encryption is for a later pull request
* **B:** Store highlight status per buffer coreside, [pull request #333](https://github.com/quassel/quassel/pull/333 )
  * Feature `BufferActivitySync`, see if buffer has highlights without fetching backlog!
  * Needed by QuasselDroid to improve notification handling
* **C:** Avoid Y2038 problem, [pull request #340](https://github.com/quassel/quassel/pull/340 )
  * Implement 64-bit message timestamps, guard protocol behind feature `LongMessageTime`
  * Older clients will work until year 2038
* **D:** Implement sender realname/avatarurl storage, [pull request #351](https://github.com/quassel/quassel/pull/351 )
  * Store sender `realname` and `avatarurl`, guard protocol behind feature `RichMessages`
  * Avatar URLs not stored yet, needs IRCv3 `METADATA`
  * Needed by QuasselDroid for display name and fallback avatar loading
* **E:** Support core-side filtering of backlog by type, [pull request #339](https://github.com/quassel/quassel/pull/339 )
  * Feature `BacklogFilterType`, allows fetching only messages of a type, e.g. if filtering join/part/quit
  * Needed by QuasselDroid for better backlog fetching performance
* **F:** Improve containerization, [pull request #341](https://github.com/quassel/quassel/pull/341 )
  * Move core state to database, allow reading configuration from environment variables
  * Allows disregarding the configuration file entirely
* **G:** Use 64-bit IDs for messages, [pull request #343](https://github.com/quassel/quassel/pull/343 )
  * Allows many more total messages, e.g. for shared cores
  * Guard protocol behind feature `LongMessageId`, old clients still work
* Cleanups
  * PRs modified to merge, fixed bugs, etc
  * Fix remaining use of 32-bit time, hopefully solving all Y2038 bugs
  * Rename `LongMessageTime` to `LongTime` due to other protocol changes
  * Change `Message::Flags`/`Type` cast to `Int`, not `UInt`
* Modify SQLite to store to milliseconds, not seconds
  * Modify all backlog, setting `time = time * 1000`
  * **This may take a long time, mention upgrade delay in release notes**
  * Modify schema upgrade message, adding `This may take a while for major upgrades.`
[...]

PR #411, opened 2018-08-06
client: Fix duplicated query quit messages

digitalcircuit:
## In short

* Fix duplicate `<nick> has quit` messages in queries
  * Search quit message timestamps ±1000 ms, not relying on seconds rounding
  * Fixes existing bug made worse by 64-bit time precision in commit 6a63070246d89aa2a2474e3a9a1035fa889ad77e
* Clean up approach
  * Don't needlessly track nicknames in hashmap
  * Add more documentation

Criteria | Rank | Reason
Impact | ★★☆ *2/3* | Fixes user-facing annoying UI regression from `0.12.5`
Risk | ★☆☆ *1/3* | Possibly hiding quit messages or showing too many

*Fixes regression from [changes made in pull request #357](https://github.com/quassel/quassel/pull/357 ).*

## Implementation
Replace the `.contains()` check with `std::binary_search` on message timestamps, passing a custom comparator that provides ± `1000` ms of fuzzy matching.  If a quit message is received and doesn't exist within this time range, it gets forwarded to the query, otherwise discarded.

Remove the `QMultiHash` method that tracked quit nicknames.  Each query gets a unique `MessageFilter`, so there's no need to store and filter that.  This will need modified if e.g. adding an option for quit messages in the Chat Monitor.

This fixes an existing bug made much worse with millisecond timestamp precision.  Before the 64-bit time changes, quit messages would only be duplicated if they got split over the turn of a second, a rare occurrence.  Afterwards, they'd get duplicated if split over the turn of a millisecond, a much more common occurrence.

Fixes regression from commit 6a63070246d89aa2a2474e3a9a1035fa889ad77e .

### Steps
1.  Connect a `0.13` core and client to network
2.  Join several channels
3.  Start a query with a test nickname in multiple channels
4.  Quit the test nickname
5.  Observe the query inside Quassel

*It's also worthwhile testing setting backlog fetching to `1`, then fetching backlog in reverse order, newest to oldest.*

### Before
*Multiple quit messages shown, one for each shared channel, `5` in this case*
```
[12:00:00 am] - {Day changed to Sunday, August 5, 2018}
[11:21:16 pm] <-- nickname (ident@hostmask) has quit (Quit: Leaving)
[11:21:16 pm] <-- nickname (ident@hostmask) has quit (Quit: Leaving)
[11:21:16 pm] <-- nickname (ident@hostmask) has quit (Quit: Leaving)
[11:21:16 pm] <-- nickname (ident@hostmask) has quit (Quit: Leaving)
[11:21:16 pm] <-- nickname (ident@hostmask) has quit (Quit: Leaving)
[12:00:00 am] - {Day changed to Monday, August 6, 2018}
```

### After
*Single quit message shown for any 1 second block of time*
```
[12:00:00 am] - {Day changed to Sunday, August 5, 2018}
[11:21:16 pm] <-- nickname (ident@hostmask) has quit (Quit: Leaving)
[12:00:00 am] - {Day changed to Monday, August 6, 2018}
```

### DIFF
--- a/src/client/messagefilter.cpp
+++ b/src/client/messagefilter.cpp
@@ -20,6 +20,8 @@
 
 #include "messagefilter.h"
 
+#include <algorithm>
+
 #include "buffersettings.h"
 #include "client.h"
 #include "buffermodel.h"
@@ -81,7 +83,7 @@ void MessageFilter::messageTypeFilterChanged()
 
     if (_messageTypeFilter != newFilter) {
         _messageTypeFilter = newFilter;
-        _filteredQuitMsgs.clear();
+        _filteredQuitMsgTime.clear();
         invalidateFilter();
     }
 }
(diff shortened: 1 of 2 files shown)


src/client/messagefilter.cpp
```diff
@@ -221,17 +223,30 @@ bool MessageFilter::filterAcceptsRow(int sourceRow, const QModelIndex &sourcePar
...
+        QString quiter = nickFromMask(sourceModel()->data(sourceIdx, MessageModel::MessageRole)
+                                      .value<Message>().sender()).toLower();
+
+        // Check that nickname matches query name
         if (quiter != bufferName().toLower())
             return false;
 
-        if (_filteredQuitMsgs.contains(quiter, messageTimestamp))
+        // Check if a quit message was already forwarded within +/- 1000 ms
+        const qint64 MAX_QUIT_DELTA_MS = 1 * 1000;
```
Sput42: Just a quicky before I have to take off for work: This should be constexpr, then it hopefully also compiles with MSVC (otherwise, it needs to be "static constexpr" or defined in global scope near the top of the file).

digitalcircuit: Thanks for catching this!  I scrolled too quickly through the AppVeyor build log since I expected failure (it was a bit odd that the other build worked, turns out this is why).

I'll give these permutations a try until MSVC is happy.

digitalcircuit:
A local `static constexpr` did the trick!

PR #414, opened 2018-08-11
Don't mess with OUTPUT_NAME

floppym:
This resolves an error when building with ninja:

ninja: error: build.ninja:4374: multiple rules generate quasselcore


Sput42:
We still want the binaries to be placed in the build root though.

I suggest adding this property to the three targets:
`RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}`

floppym:
Thanks for the suggestion, that seems to do the trick. I force-pushed an update.

PR #415, opened 2018-08-17
common: Cache Highlight/Ignore/nick rules, port to QRegularExpression, fixes

digitalcircuit:
## In short

* Add `ExpressionMatch` class, unifying regular expression handling
  * Supports `Phrase`, `MultiPhrase`, `Wildcard`, `MultiWildcard`, and `RegEx` patterns
  * Condenses patterns into a single regular expression, reusing it
  * Fixes inability to escape `?`, `*`, `!`, and `;` in `Wildcard`/`MultiWildcard`
  * Backed by partial tests in `ExpressionMatchTests`
* Add `NickHighlightMatcher` class, unifying and caching nickname highlights
  * Track nickname per network, automatically rebuild pattern on change
  * Remove cache on network removal
* Port local/core-side highlights, ignore rules to above new classes
  * Use `ExpressionMatch` for all, `NickHighlightMatcher` for highlights
  * Should significantly improve performance when many rules are enabled
* Fix issue with invalid whitespace-only CTCP ignore rules
  * Check length of sender types before attempting to use value
* Add documentation [to Quassel FAQ on new escape patterns](https://bugs.quassel-irc.org/projects/quassel-irc/wiki/FAQ#How-do-I-ignore-specific-highlights )
* Clean up
  * Properly handle ignore rule scope with newlines
  * Fix invalid resource path in settings page
  * Remove broken `scopeMatch()` function from `util.cpp`

Criteria | Rank | Reason
Impact | ★★★ *3/3* | User-facing performance, ability to escape `*`, `!`, etc
Risk | ★★★ *3/3* | Manual migrations, possible broken highlight/ignore rules
Intrusiveness | ★★☆ *2/3* | Modifes highlight/ignore processing, may conflict with other changes

*Thanks to @sandsmark for the initial `QRegularExpression` efforts!*

## Rationale
Quassel processes potentially hundreds of messages at a time to apply highlight and ignore rules, some retroactively (ignore rules, local highlights).  To improve performance, Quassel should compile down regular expressions to the simplest format, and cache them for reuse until changed.  Similarly, on Qt 5, porting to `QRegularExpression` provides further speedups.

Some IRC channels can begin with `!`.  Quassel scope matching needs a way to escape `!` to match it literally, e.g. for `!channel`.

Spam messages can contain `*` and `?` characters, such as `**** ATTENTION ****`.  Quassel should provide a way to escape these characters to match them literally.

### Breaking changes
* `Wildcard` no longer supports character classes, like `[abc]` to match `a`, `b`, or `c`
  * For highlight rules and ignore rule contents, manually switch to `RegEx` mode and rewrite the rule
  * For ignore rule scope, **no workaround exists** as `Regular expression` does not affect the scope (legacy reasons)
* `\` in `Wildcard` rules must now be escaped
  * `\*` translates to literally matching `*`, `\?` to literal `?`
  * `\\[...]` translates to literally matching `\[...]`
  * Necessary to support escaping `*`, `?`, and `;`
* `!` at the start of `Wildcard` rules now inverts the rule
  * E.g. `!*some message*` sets a rule for everything that does not contain `some message`
  * `\![...]` escapes the `!` to literally match `![...]`
  * `\\![...]` escapes the `\` to literally match `\![...]`

### `ExpressionMatch`
`ExpressionMatch` implements an iterative parser, tracking state while looping through the list.  This is necessary to provide escaping sequences for `Wildcard` and `MultiWildcard`, akin to [how Qt implemented their algorithm](https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/tools/qregexp.cpp#n754 ) but dropping the character class support, and adding handling `;` as a separator.

`ExpressionMatch` will take any input pattern and simplify it into a single regular expression, optimizing for speed.

**To avoid separating documentation from code, I've put most of the implementation remarks into the comments of the code itself.**

It is not easily possible to implement escape sequences with simple string-to-regex substitution.

Test cases are implemented as opt-in `Q_ASSERT()` via `ExpressionMatchTests::runTests()`.  **These are not run by default.**  A quick edit to `quassel.cpp` will add them in for development tests:

```diff
--- a/src/common/quassel.cpp
+++ b/src/common/quassel.cpp
@@ -54,6 +54,8 @@
 
 #include "../../version.h"
 
+#include "expressionmatchtests.h"
+
 Quassel *Quassel::instance()
 {
     static Quassel instance;
@@ -124,6 +126,8 @@ bool Quassel::init()
         return false;
     }
 
+    ExpressionMatchTests::runTests();
+
     // Don't keep a debug log on the core
     return instance()->logger()->setup(runMode() != RunMode::CoreOnly);
 }
```

This is designed to be adaptable to a proper test framework whenever Quassel adopts one.

### `NickHighlightMatcher`
`NickHighlightMatcher` builds upon `ExpressionMatch`, using the `MultiPhrase` mode, and stores nickname-based expression match objects per network.

Whenever a nickname set changes, or a network is removed, the cache is deleted and built anew on demand.

### Legend
Symbol | Meaning
-------|--------
:checkered_flag: | Behavior in `0.12.5` stable release
:ballot_box_with_check: | Behavior in `0.13rc1`
:new: | New or changed in this pull request

### Modes of operation

Nickname matching always uses `MatchMode::MultiPhrase`.

**`Regular expressions` / `RegEx` is NOT checked**

Item | `MatchMode` for patterns | Status
-----|--------------------------|-------
Highlight rule contents | `Phrase` | :checkered_flag: *In stable*
Highlight rule sender/channel | `MultiWildcard` | :ballot_box_with_check: *In `0.13rc1`*
Ignore rule contents | `Wildcard` | :checkered_flag: *In stable*
Ignore rule scope | `MultiWildcard` | :checkered_flag: *In stable*

**`Regular expressions` / `RegEx` is checked**

Item | `MatchMode` for patterns | Status
-----|--------------------------|-------
Highlight rule contents | `RegEx` | :checkered_flag: *In stable*
Highlight rule sender/channel | `RegEx` | :checkered_flag: *In stable*
Ignore rule contents | `RegEx` | :checkered_flag: *In stable*
Ignore rule scope | `MultiWildcard` | :checkered_flag: *In stable*

*Ignore rule scope is fixed to `MultiWildcard` for backwards compatibility.*

### Phrase (`Phrase`)
*This is used for highlight rules when `RegEx` is not checked*

**:checkered_flag: In stable**

```
word
```

* Matches
  * `word`
  * `A word.`
  * `has a word in it`
* Does not match
  * `wording`

**:new: New to this PR**
*There's a leading space; this can be used for more exact matches*

```
 spaced
```

* Matches
  * `, spaced`
  * `is spaced out`
* Does not match
  * `;spaced`

### Multiple phrases (`MultiPhrase`)
*This is used for nickname matching.*

**:checkered_flag: In stable**

```
nick
othernick
test
```

* Matches
  * `nick`
  * `nick[something]`
  * `othernick: ...`
  * `pinging test hi`
* Does not match
  * `nicks`
  * `tests`

*Quassel `0.12.5` also would match `nick[something]`; if this isn't desirable, it can be changed, albeit with some complexity, needing a new `Nickname` matching mode.  Quassel currently assumes non-word characters are not valid in nicknames as per the `\W` in phrase regular expression.*

### Wildcard (`Wildcard`)
*This is used for ignore rule contents when `Regular expression` is not checked*

**:new: Changed in this PR**

```
*Asking questions\?  Nope?
```

* Matches
  * `Asking questions?  Nopea`
  * `Basking questions?  Nope.`
* Does not match
  * `Asking questions.  Nope.`
  * `Asking questions?  Nopes.`

**:new: New to this PR**
*Implicit wildcard is supported, too*

```
!*filter*
```

* Matches
  * *Everything, unless it contains something from the below list*
* Does not match
  * `filter`
  * `filtering`
  * `#nofilter yo`

**:new: New to this PR**
*Exclamation points can be escaped at beginning (not required elsewhere)*

```
\!*filter*
```

* Matches
  * `!filter`
  * `!yes filtering`
* Does not match
  * `filter`

**:new: New to this PR**
*Escape character can be escaped*

```
\\!*filter*
```

* Matches
  * `\!filter`
  * `\!yes filtering`
* Does not match
  * `filter`
  * `!filter`

### Multiple wildcards (`MultiWildcard`)
*This is used for ignore rule scope regardless of `Regular expression` checkbox, and for highlight rule sender/channel when `RegEx` is not checked*

**:new: Changed in this PR**

```
Alice!*; Bob!*@example.com; Carol*!*; !Caroline!*
Dan!*; escaped \; separator; \!not-inverted; \\!slash-prefixed
```

*Newlines and `;` are interchangeable in ignore rule scope; highlight rules treat <kbd>Enter</kbd> as submit*

* Matches
  * `Alice![...]`
  * `Bob![...]@example.com`
  * `Carol[...]![...]` except as noted below
  * `Dan![...]`
  * `escaped ; separator`
  * `!not-inverted`
  * `\!slash-prefixed`
* Does not match
  * `Caroline![...]`
  * `Malice![...]`
  * `John!`

**:ballot_box_with_check: In `0.13rc1`**
*Implicit wildcard is supported, too*

```
!Announce*!*; !Wheatley!aperture@*
```

* Matches
  * *Everything, unless it contains something from the below list*
* Does not match
  * `Announce[...]![...]`
  * `Wheatley!aperture@[...]`

### Regular expression (`RegEx`)
*This is used for ignore rule contents and highlight rule contents/sender/channel when `RegEx` is checked*

**:checkered_flag: In stable**

```
simple.\*escape-match.*
```

* Matches
  * `simpleA*escape-match`
  * `simpleA*escape-matchBBBB`
* Does not match
  * `not above`
  * `simpleA*escape-mat`
  * `simple*escape-match`
  * `simpleABBBBescape-matchBBBB`

**:checkered_flag: In stable**
*Inverted rules are supported, too*

```
!invert.\*escape-match.*
```

* Matches
  * *Everything, unless it matches something from the below list*
* Does not match
  * `invertA*escape-match`
  * `invertA*escape-matchBBBB`

**:new: Changed in this PR**
*Exclamation points can be escaped at beginning (not required elsewhere)*

```
\!simple.\*escape-match.*
```

* Matches
  * `!simpleA*escape-matchBBBB`
* Does not match
  * `simpleA*escape-matchBBBB`

**:new: New to this PR**
*Escape character can be escaped at beginning*

```
\\!simple.\*escape-match.*
```

* Matches
  * `\!simpleA*escape-matchBBBB`
* Does not match
  * `!simpleA*escape-matchBBBB`

---

*Though I checked, I may still have made mistakes in the examples above.  If you want some samples to test with, try out the above.*


digitalcircuit:
**Notes to self:** I'll need to switch this to drop backtracking and use looping instead.  Two dedicated loops should be fine, one for handling escapable `;` splitting in `IgnoreListEditItemSomethingDlg`, and one for `Wildcard` up-conversion in `ExpressionMatch` that handles `;` splitting and `\*` + `\?` reg-ex escape fix-ups.

Also ignore greedy nature, `*` is greedy in shell globbing by default.  That performance can be profiled/fixed later, should be much smaller difference than combing and caching regex instances.

digitalcircuit:
**Notes to self:**

* Maybe splitting first, ignoring the QRegExp::escape(), then wildcardToRegEx upconverting each individual part will work better after all.
  * "!" at start inverts, "\!" at start converts to "!", "!" elsewhere stays "!", "\!" elsewhere stays "\!".
";" splits, "\;" doesn't split and gets converted to ";", "\\;" splits and "\\" stays as "\\" (to be converted into "\" in wildcardToRegex()), "\[...]" stays as "\[...]" where "[...]" represents any other character (not ";")
  * "\n" always splits.
* And the trimming logic of IgnoreListEditDlg::widgetHasChanged() will always ignore "!" (treat it as the same, don't remove it), and never splits on "\n" (as that's splitted separately).
* Add a ::Wildcard vs. ::MultiWildcard mode, so Ignore Rules won't have splitting (just the direct wildcard convert), but the Ignore Rule/Highlight rule scope will.

digitalcircuit:
Tentative working draft is here!  Just need to do all the documentation and testing...

genius3000:
Haven't been able to get fully into testing this but gave it a quick try with both client/core updated. Also to note, my test core doesn't have enough backlog/large enough channels to really test the time to load backlog issue. I'll try to remedy that for more in-depth testing.  

#### Initial good findings:
* Ignore List
  * Regex of `.*(test|foo|bar).*` works
  * Wildcard of `meow*`, `meow`, `meow\*`, `meow\?` work (matching wildcard or literal escaped character, as expected)

* Remote Highlights
  * Simple regexes (ie. `.*meow.*`, `meow.*`) work
  * Literal matching (non-regex) of `meow`, `meow*` work

* Local Highlights (currently same as Remote Highlights)

#### Noted concerns (minor/thoughts):
* Ignore list manager logs a Warning of trying to set an empty rule, when changing options in the widget (and the rule is empty):
`[Warn ] void IgnoreListManager::IgnoreListItem::setContents(const QString&) Asked to set an ignore rule to empty`
Example: Create new rule, change from Dynamic to Permanent, Sender to Message, non-regex to regex, etc.
At a glance, it looks like it would be caused by [this line](https://github.com/quassel/quassel/pull/415/files#diff-67037546156688f7088a5028cff166d2R341) which calls .setContents() whenever the widget is updated. Is it trying to save the rule with every change rather than just upon clicking `Ok`?

* Local/Remote Highlights
I was actually surprised that these didn't support wildcard matching when set non-regex. Had to double check with current 0.13-rc1 to verify intended behavior. I've always used a few regex rules to match/not match specific criteria and never noticed this before. I realize the match allows for a trailing `';', ','` and maybe others. Would it be worth allowing wildcard matching here or just rely on the regex option when people want that ability?

&nbsp;  
Thanks for the awesome work so far with this, it should definitely improve things and I'll try to get more testing done before too late.

digitalcircuit:
@genius3000 I've addressed the warning issue (by removing it).  It's not a crashing issue, it just means that the ignore rule won't ever match.  Quassel never warned about this in the past, so it should be fine to leave as is.

Supporting wildcard for local/remote highlights is a nice idea, and now's the time to do any backwards-incompatible protocol breaking and string changes before `0.13` becomes stable, to provide a `Phrase`, `Wildcard`, and `RegEx` dropdown.  That said, regular expressions work, too, so I feel it might not be worth the extra complexity.  Either way, it'd be a follow-up pull request to this :slightly_smiling_face:

Thank you for your initial testing!  I've got my own testing and documentation to write, then this should be okay...

digitalcircuit:
### 2018-9-1 - More testing is welcome!

I tested the basics (highlight match, nick match, ignore rules), and did more piecemeal tests during development, but I haven't extensively tried patterns outside of `ExpressionMatchTests`.

digitalcircuit:
Made a small typo in a test case and pull request documentation (regex `\\!test*` matches `\!test[...]`), no impact on matching code.  Fixed.

digitalcircuit:
With the unifying changes in this pull request, I have been wondering how little (or much) work is left to give `Local Highlights` all the functionality of `Remote Highlights` - add a settings minor version upgrade, copy the settings page table UI and code (minus import button), copy the `HighlightRule` class, add a few more message checks, change the import mechanics...

We'll see if/when this is merged, maybe.

PR #417, opened 2018-08-22
settings: Fix Chat Monitor defaults

digitalcircuit:
## In short

* Unify `Chat Monitor` defaults
  * Syncs three separate places for setting defaults
  * Defaults to disabling showing read messages, enabling showing own messages
  * Add a comment to explain and hopefully avoid this in the future
* Fix `Settings` default caching wrong values
  * Fix caching of default values overriding different defaults elsewhere
  * Fixes `Chat Monitor` settings showing up as invalid instead of `Opt Out`
  * Cache key persistence, possibly speeding up `localKeyExists()`

Criteria | Rank | Reason
Impact | ★★☆ *2/3* | User-facing defaults fix-up, avoids UI quirk
Risk | ★★☆ *2/3* | Different defaults might be used, new cache mechanics

### Before - Chat Monitor settings
*In `Settings` → `Configure Quassel…` → `Interface` → `Chat Monitor`*
Chat Monitor settings, showing the Operation Mode combobox as having nothing selected

*Chat Monitor's `Operation Mode` combobox is not selected*

### After - Chat Monitor settings
*In `Settings` → `Configure Quassel…` → `Interface` → `Chat Monitor`*
Chat Monitor settings, showing the Operation Mode combobox as having "Opt Out" selected

*Chat Monitor's `Operation Mode` combobox correctly selects `Opt Out` by default*

### Debug log showing differences
*First instance is from `ChatMonitorFilter`, second is from `ChatMonitorSettingsPage`*
```
2018-08-22 16:39:06 [Debug] Settings::localValue()
Returning default value for key "QtUi/ChatView/ChatMonitor/OperationMode" ,
default QVariant(int, 0)

2018-08-22 16:39:08 [Debug] Settings::localValue()
Returning default value for key "QtUi/ChatView/ChatMonitor/OperationMode" ,
default QVariant(int, 2)
```

*Before, the default value from the first would be cached, causing `Settings` to return that value for the second even though it requested a different default.*

### DIFF
--- a/src/common/settings.cpp
+++ b/src/common/settings.cpp
@@ -30,6 +30,7 @@ const int VERSION = 1;              /// Settings version for backwords/forwards 
 const int VERSION_MINOR_INITIAL = 1; /// Initial settings version for compatible changes
 
 QHash<QString, QVariant> Settings::settingsCache;
+QHash<QString, bool> Settings::settingsKeyPersistedCache;
 QHash<QString, SettingsChangeNotifier *> Settings::settingsChangeNotifier;
 
 #ifdef Q_OS_MAC
@@ -186,6 +187,7 @@ void Settings::setLocalValue(const QString &key, const QVariant &data)
     QString normKey = normalizedKey(group, key);
     create_qsettings;
     s.setValue(normKey, data);
+    setCacheKeyPersisted(normKey, true);
     setCacheValue(normKey, data);
     if (hasNotifier(normKey)) {
         emit notifier(normKey)->valueChanged(data);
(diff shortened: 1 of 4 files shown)


digitalcircuit:
This also provides the same fix as in [pull request #396](https://github.com/quassel/quassel/pull/396 ), but does it more elegantly with a separate cache instead of removing the cache entirely for key existence/persistence checks.

I followed the test case in that pull request and this didn't break it, either, so.. hopefully all works :slightly_smiling_face:

PR #422, opened 2018-08-29
Fix issues with TreeModel

Sput42:
Don't send the dataChanged() signal of PropertyMapItems before actually setting the new value.

Don't store the property list dynamically in every item instance.

### DIFF
--- a/src/client/networkmodel.cpp
+++ b/src/client/networkmodel.cpp
@@ -40,7 +40,7 @@
 *  Network Items
 *****************************************/
 NetworkItem::NetworkItem(const NetworkId &netid, AbstractTreeItem *parent)
-    : PropertyMapItem(QList<QString>() << "networkName" << "currentServer" << "nickCount", parent),
+    : PropertyMapItem(parent),
     _networkId(netid),
     _statusBufferItem(0)
 {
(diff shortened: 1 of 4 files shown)


digitalcircuit:
It's hard to test for the absence of a crash one's not entirely sure how to recreate, but in spamming `Connect` and `Disconnect` **I haven't had this crash on me yet.**

However, **there is another tree model bug that existed before this pull request.**  You can try addressing it here, or leave it for another code change if you'd like.

### Steps

1.  Set up a fresh Quassel configuration (`monolithic`, or `core` + `client`)
2.  Create a test identity following the wizard
3.  Connect to a test network following the wizard, **renaming the network to `A` before connecting**
    * *Or you can rename it afterwards, just wanting it to be first alphabetically*
4.  Once connected, set up two additional chat lists in `Configure Quassel…` → `Interface` → `Custom Chat Lists`
    * Add `Channels`, uncheck `Show queries`
    * Add `PMs`, uncheck `Show channels` and **check `Hide inactive networks`**
5.  Save and close the settings window
6.  Close the `All Chats` chat list, leaving only the two new ones visible
7.  Add another test network, **naming it `B`**
8.  Connect to the `B` test network
9.  In the `Channels` chat list, right-click `Disconnect` from network `A`

### Expected
Network `A` disappears from the chat lists.

### Actual
Both network `A` and network `B` disappear from the chat lists.

*Again, this can be addressed in another pull request if needed.  Wasn't sure if you hoped to catch this issue too, here.*

Sput42:
Yeah, I wasn't planning to do deeper debugging into the models. This was just something I stumbled upon when doing unrelated things :)

digitalcircuit:
Agreed.  I've filed the above comment away in a personal TODO list, and I'll (try to) eventually get around to filing a bug report on Quassel IRC's bug tracker.

PR #427, opened 2018-09-04
Fix issues found in static analysis

Sput42:
Running clang-tidy on the codebase uncovered a bunch of issues even with default checks, so fix them before release.

### DIFF
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -16,7 +16,7 @@ set(SOURCES
     event.cpp
     eventmanager.cpp
     expressionmatch.cpp
-    expressionmatchtests.cpp
+    # expressionmatchtests.cpp
     highlightrulemanager.cpp
     identity.cpp
     ignorelistmanager.cpp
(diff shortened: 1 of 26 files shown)


src/common/CMakeLists.txt
```diff
@@ -16,7 +16,7 @@ set(SOURCES
     event.cpp
     eventmanager.cpp
     expressionmatch.cpp
-    expressionmatchtests.cpp
+    # expressionmatchtests.cpp
```
digitalcircuit: Thanks for catching this!  I wasn't sure how to handle it as Qt Creator didn't appear to show the file without it being in `CMakeLists.txt` - now that I know this causes warnings, I'll leave similar faux-test files out in the future.

Looking forward to adopting an actual test framework :slightly_smiling_face:

Sput42: Well, the warnings were about unused variables (ones that were only used within Q_ASSERT). And only in some CI configurations. I guess since we're doing release builds in CI, Q_ASSERT gets optimized out and some compilers don't catch this properly to mark variables as used.

Rather than trying to work around this in the source, I decided to just not compile it :)

PR #441, opened 2018-10-29
allow selecting the search result when searching for buffers

sandsmark:


sandsmark:
basically "fake" currentindex, because currentindex is used as the current selection all over the code and i didn't want to fix it all over (also because a ton of potential for regressions).

Sput42:
@sandsmark: I took the liberty to make your commit message nicer and fix some nitpicks :) I think this should still go into 0.13.0.

PR #448, opened 2018-11-18
cmake: Move libraries into a private directory

Sput42:
Since Quassel's libraries are not meant for public consumption,
move them from libdir into libdir/quassel. Always set RPATH
accordingly, so the executables find their libraries.

On Windows, DLLs count as runtime artifacts rather than libraries,
so they remain unaffected and stay in the bindir.


Sput42:
OK, after consideration and feedback from Debian packaging, we won't do this. Since Quassel should be split in several packages, the libraries should be put in the standard, public location.

PR #450, opened 2018-11-18
serializers: Fix quint16 deserialization, remove code duplication

Sput42:
Fix a bug where a quint16 would be deserialized into a QVariant
of type Int (instead of UShort). This broke the sync of
autoReconnectRetries in the network settings, because the slot's
signature didn't match the type of the QVariant.

The issue was caused by QVariant not having a constructor for
quint16, thus casting it into an int. Fix this by using
QVariant::fromValue(), which handles the type correctly.

Use the opportunity to remove much of the code duplication in the
deserialization code by providing and using a template function,
avoiding potential other issues with using QVariant's constructor.

### DIFF
--- a/src/common/serializers/serializers.cpp
+++ b/src/common/serializers/serializers.cpp
@@ -20,6 +20,21 @@
 
 #include "serializers.h"
 
+namespace {
+
+template<typename T>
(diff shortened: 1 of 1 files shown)


digitalcircuit:
Will this apply cleanly to `0.13`, i.e. if I manually carry over the changes nothing else needs changed?

PR #459, opened 2019-01-03
tests: Verify ExpressionMatch with invalid regex

digitalcircuit:
## In short

* Verify `ExpressionMatch` handles invalid regular expressions correctly
  * Ensure there's no crash and that this won't regress in future changes
  * Matches @justJanne 's [test for Quasseldroid-NG](https://git.kuschku.de/justJanne/QuasselDroid-ng/commit/f1f320782cb0bca3fb7974e7688ce33af19456cf )

Criteria | Rank | Reason
Impact | ★☆☆ *1/3* | Further verifies proper operation of `ExpressionMatch`
Risk | ★☆☆ *1/3* | Adds another development-only test

### `ctest` run

```
[...]/Quassel/quassel-build$ ctest
Test project [...]/Quassel/quassel-build
    Start 1: ExpressionMatchTest
1/3 Test #1: ExpressionMatchTest ..............   Passed    0.00 sec
    Start 2: FuncHelpersTest
2/3 Test #2: FuncHelpersTest ..................   Passed    0.00 sec
    Start 3: SignalProxyTest
3/3 Test #3: SignalProxyTest ..................   Passed    0.01 sec

100% tests passed, 0 tests failed out of 3

Total Test time (real) =   0.02 sec
```

### DIFF
--- a/tests/common/expressionmatchtest.cpp
+++ b/tests/common/expressionmatchtest.cpp
@@ -434,3 +434,22 @@ TEST(ExpressionMatchTest, trimMultiWildcardWhitespace)
         EXPECT_EQ(ExpressionMatch::trimMultiWildcardWhitespace(result), result);
     }
 }
+
+
+TEST(ExpressionMatchTest, testInvalidRegEx)
+{
+    // Invalid regular expression pattern
+    ExpressionMatch invalidRegExMatch =
+            ExpressionMatch("*network", ExpressionMatch::MatchMode::MatchRegEx, false);
+
+    // Assert not valid
+    ASSERT_FALSE(invalidRegExMatch.isValid());
+    // Assert not empty
+    EXPECT_FALSE(invalidRegExMatch.isEmpty());
+    // Assert default match fails
+    EXPECT_FALSE(invalidRegExMatch.match(""));
+    // Assert wildcard match fails
+    EXPECT_FALSE(invalidRegExMatch.match("network"));
+    // Assert literal match fails
+    EXPECT_FALSE(invalidRegExMatch.match("*network"));
+}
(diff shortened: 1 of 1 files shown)


justjanne:
On the Quasseldroid side this is also useful to ensure that we won’t get issues with "Invalid Syntax" exceptions from the regex engine.

tests/common/expressionmatchtest.cpp
```diff
@@ -434,3 +434,22 @@ TEST(ExpressionMatchTest, trimMultiWildcardWhitespace)
...
+
+
+TEST(ExpressionMatchTest, testInvalidRegEx)
+{
+    // Invalid regular expression pattern
+    ExpressionMatch invalidRegExMatch =
+            ExpressionMatch("*network", ExpressionMatch::MatchMode::MatchRegEx, false);
+
+    // Assert not valid
+    ASSERT_FALSE(invalidRegExMatch.isValid());
```
digitalcircuit: Given the difference between `ASSERT` and `EXPECT`, `ExpressionMatch` should *never* crash/cause issues if `isValid()` is false, so all `ASSERT`s (in all tests) could be changed to `EXPECT`.

Here, I just copied the pattern from the other tests.  If preferred, I could change the other tests instead.

Sput42: Well, question is if it makes sense to run the other checks if that assertion fails? I don't care much either way, whatever you think makes more sense ;-)

digitalcircuit: Since I haven't had to deal with these tests failing and needing debugged, I'm really not sure.  If the `isValid()` check is broken, most of the results will be bogus anyways.

I think I'll leave it at an `ASSERT` for now, and if this turns out to be hard to debug in the future, we can always change it then.

(However, I did find a place that *should* be an `ASSERT`, and will fix that!)

PR #460, opened 2019-01-12
Make output to syslog nicer

vehk:
Before: `2019-01-11T22:52:27+00:00 hostname info [Info ] Quitting...`
After:  `2019-01-11T16:19:43+00:00 hostname info quasselcore[4438]: Quitting...`

This patch improves quasselcore's syslog output by adding a program name and instructing the logger to include the process's PID for every message.

Additionally, it removes superfluous log levels in the log message, as syslog is already aware of them.

Potential issues: If you log to a file and syslog at the same time (not sure whether this is possible, didn't check), the log level will not be included in the log file, since the check for _syslogEnabled will override that.

### DIFF
--- a/src/common/logger.cpp
+++ b/src/common/logger.cpp
@@ -37,7 +37,27 @@ namespace {
 
 QByteArray msgWithTime(const Logger::LogEntry& msg)
 {
-    return (msg.timeStamp.toString("yyyy-MM-dd hh:mm:ss ") + msg.message + "\n").toUtf8();
+    QString levelString;
+
+    switch (msg.logLevel) {
+        case Logger::LogLevel::Debug:
+            levelString = "[Debug] ";
+            break;
+        case Logger::LogLevel::Info:
+            levelString = "[Info ] ";
+            break;
+        case Logger::LogLevel::Warning:
+            levelString = "[Warn ] ";
+            break;
+        case Logger::LogLevel::Error:
+            levelString = "[Error] ";
+            break;
+        case Logger::LogLevel::Fatal:
+            levelString = "[FATAL] ";
+            break;
+    }
+
+    return (msg.timeStamp.toString("yyyy-MM-dd hh:mm:ss ") + levelString + msg.message + "\n").toUtf8();
 }
 
 }  // namespace
@@ -151,31 +171,8 @@ void Logger::handleMessage(QtMsgType type, const QString& msg)
 
 void Logger::handleMessage(LogLevel level, const QString& msg)
 {
-    QString logString;
-
-    // Only add the log level to the message if we do not output to syslog
-    if (!_syslogEnabled) {
-        switch (level) {
-        case LogLevel::Debug:
-            logString = "[Debug] ";
-            break;
-        case LogLevel::Info:
-            logString = "[Info ] ";
-            break;
-        case LogLevel::Warning:
-            logString = "[Warn ] ";
-            break;
-        case LogLevel::Error:
-            logString = "[Error] ";
-            break;
-        case LogLevel::Fatal:
-            logString = "[FATAL] ";
-            break;
-        }
-    }
-
     // Use signal connection to make this method thread-safe
-    emit messageLogged({QDateTime::currentDateTime(), level, logString += msg});
+    emit messageLogged({QDateTime::currentDateTime(), level, msg});
 }
 
 void Logger::onMessageLogged(const LogEntry& message)
--- a/src/common/logger.cpp
+++ b/src/common/logger.cpp
@@ -106,6 +106,10 @@ void Logger::setup(bool keepMessages)
 
 #ifdef HAVE_SYSLOG
     _syslogEnabled = Quassel::isOptionSet("syslog");
+
+    // set up options, program name, and facility for later calls to syslog(3)
+    if (_syslogEnabled)
+        openlog("quasselcore", LOG_PID, LOG_USER);
 #endif
 
     _initialized = true;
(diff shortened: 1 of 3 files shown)


digitalcircuit:
Though I'd have to double check when not on mobile, I'm pretty sure it's possible to enable logging to file and syslog at the same time (just pass both options).

If so, perhaps the `logLevel` string could be kept separate, concatenated to all outputs *except* for syslog?

vehk:
I took another look at the code - logs to the logfile and stdout all go through msgWithTime, so I moved the log level string generation there.

This fixes any potential issues, and allows regular logs to files and stdout to be formatted nicely even with output to syslog enabled.

vehk:
I do not have a systemd-based machine ready, so I can't test this explicitly, but even with this patch I am pretty sure that journalctl will keep highlighting warnings and errors (since most probably it looks at the log level itself and does not parse the log message).

digitalcircuit:
I went ahead and [backported this to `0.13` as pull request #472](https://github.com/quassel/quassel/pull/472) for the impending `0.13.1` release.  Let me know if you'd rather have the attribution changed, or anything!

vehk:
Looks good to me, thanks very much!

src/common/logger.cpp
```diff
@@ -106,6 +126,10 @@ void Logger::setup(bool keepMessages)
 
 #ifdef HAVE_SYSLOG
     _syslogEnabled = Quassel::isOptionSet("syslog");
+
+    // set up options, program name, and facility for later calls to syslog(3)
+    if (_syslogEnabled)
+        openlog("quasselcore", LOG_PID, LOG_USER);
```
Sput42: This is wrong, because the process could also be quassel or quasselclient. To correctly determine this, one would probably need to work with Quassel::buildInfo() (and Quassel::runMode() to determine which of its properties to use). I'm not aware of a platform-independent way to determine the current process name generically...

vehk: Would it be acceptable here to use "quassel" for everything? If not, I see the following possibilities:
1. Use RunMode and BuildInfo, like stated.
2. Use [QCoreApplication::applicationFilePath](https://doc.qt.io/qt-5/qcoreapplication.html#applicationFilePath) to get the program name

Either of these requires storing the program name in Logger, since it would otherwise go out of scope.

Sput42: I think a proper solution would require to go with 1). I'm not sure why we'd have to keep the name in Logger, since it's only used when calling openlog(), isn't it?

vehk: openlog() takes a const char*, and results in syslog referring back to that pointer on every call, so simply passing QString.toLocal8Bit().constData() does not guarantee the availability of said data.
In my test code I ended up storing the QByteArray from toLocal8Bit() in Logger itself, such that it was always in scope there. Keep in mind, however, that I have no experience with cpp (or QT programming for that matter), so there may be a cleaner solution.

Sput42: Hmm, my manpage for openlog() says: "If ident is NULL, the program name is used. (POSIX.1-2008 does not specify the behavior when ident is NULL.)" Seems like at least on a modern Linux, we could just pass a null pointer to openlog(), bypassing this problem. Since that behavior isn't specified in POSIX, I don't know how this behaves on other syslog-supporting platforms, however...

In any case, storing the string in the Logger class just to be safe isn't an issue either.

vehk: Yeah, that was a source of initial confusion for me as well, since the manpage goes on to state that, if needed, syslog(3) will call openlog(3) with ident set to NULL, so I should have seen at least the program name on all of my test machines even without my patch.

Turns out at least musl does not automatically set the program name and therefore relies on a call to openlog(3). 

I'll update this branch with a fix for this later today, getting the program name from RunMode and BuildInfo like agreed.

digitalcircuit:
I've updated the [backport pull request #472](https://github.com/quassel/quassel/pull/472 ), and it seems to match exactly, asides from a few code cleanup changes.

digitalcircuit:
Ah, found an issue - this removes the `[Debug]`/etc tags in the Debug Log window.  I'll fix this in the backport, which can be pulled here.

vehk:
Adapted the code & pushed with original author information and message.

digitalcircuit:
I'm unsure of what's happening, either.  I know there was a lot of build system cleanups after 0.13's release, and it's possible Microsoft's updated compiler/changed build flags/etc are not happy with what I did in 0.13.  From quick looking around on mobile, nothing stands out as obviously incorrect.

If I had to guess, `quassel-qtui.lib` needs to declare a reference or dependency on the common code library, somehow..?  After 0.13, Sput split Quassel into shared libraries, but I don't understand why Linux and macOS builds are just fine.

I'm on a long road trip at the moment without access to my development setup, but once I arrive at the destination I'll look into it this when I get the time.  I apologize for the hassle; I really didn't expect this to break, let alone in Exciting New Ways™.

Build error message for easier reference:
```
   Creating library lib\quassel-qtui.lib and object lib\quassel-qtui.exp
debuglogdlg.cpp.obj : error LNK2019: unresolved external symbol "public: class QString __cdecl Logger::LogEntry::toString(void)const " (?toString@LogEntry@Logger@@QEBA?AVQString@@XZ) referenced in function "public: __cdecl DebugLogDlg::DebugLogDlg(class QWidget *)" (??0DebugLogDlg@@QEAA@PEAVQWidget@@@Z)
bin\quassel-qtui.dll : fatal error LNK1120: 1 unresolved externals
ninja: build stopped: subcommand failed.
Action: compile for quassel/quassel:master FAILED
*** Craft all failed: quassel/quassel after 9 minutes 21 seconds ***
fatal error: package quassel/quassel all failed
Command exited with code
```

@Sput42 This doesn't affect the 0.13 backport, though if the C++ code needs changed to fix this, master will diverge (likely not an issue so long as it's equivalent functionality-wise).

Sput42:
> If I had to guess, `quassel-qtui.lib` needs to declare a reference or dependency on the common code library, somehow..? After 0.13, Sput split Quassel into shared libraries, but I don't understand why Linux and macOS builds are just fine.

Welcome to DLL hell. All symbols need to be exported for them to work. While the Logger class is (see the COMMON_EXPORT macro), the subclass LogEntry isn't. Try adding COMMON_EXPORT to that one too; I'm not sure if that works in this nested way. If all else fails, we either need to move the class out of Logger to make it standalone, or introduce a proxy method Logger::toString() that takes a LogEntry to output...

PR #465, opened 2019-01-22
Update pt_BR.po

rodrigorodrigo:
first version, I will keep completing this as soon as I can.


AlD:
Thanks for contributing!
We're managing translations in Transifex though, which prevents us from accepting direct commits to these files.
Could you please head over to https://www.transifex.com/quassel/quassel/translate/#pt_BR/master and submit your translations there?
Thanks!

PR #466, opened 2019-01-23
cmake: add ENABLE_SHARED option, default ON

rkitover:
Add an ENABLE_SHARED option, defaults to ON, when set to OFF
quassel_add_module() will build modules as static (does not affect
platforms where modules would be built as static anyway.)

The option name ENABLE_SHARED is frequently used for this purpose in
cmake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -360,6 +360,12 @@ if (NOT WIN32)
     )
 endif()
 
+# Shared library support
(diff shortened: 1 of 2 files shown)


rkitover:
Added signoff.

Sput42:
Added feature info for the flag and moved around the option so it doesn't look out of place.

PR #467, opened 2019-01-24
lower Boost dep version 1.56 -> 1.54

rkitover:
Require Boost 1.54 instead of 1.56.

Tested to compile and work correctly on OpenSUSE Leap 42.3.

Related issue #1510.

Signed-off-by: Rafael Kitover <rkitover@gmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -330,7 +330,7 @@ endif()
 # Non-Qt-based packages
 #####################################################################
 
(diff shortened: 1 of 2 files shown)


digitalcircuit:
@Sput42 if you agree with this change, this should probably be backported for the 0.13.1 release, too.

Sput42:
Slightly reworded the commit message for automatic ticket closing and proper capitalization in the subject.

PR #477, opened 2019-02-07
Core: only try local auth if a user has the "database" auth provider set

TC01:
This commit attempts to resolve issue #1501; currently, quassel will always attempt to do local authentication, and *then* try LDAP auth. This makes it difficult to migrate a core from local auth to LDAP auth; users who have local passwords can always use those passwords to log in, and there is not a simple way to migrate them to non-local auth.

There is an "authenticator" column in the database for each user, but it is currently only used to stop passwords from being changed for LDAP users. This commit modifies the login flow to first check that database field-- if authenticator is not "Database", we don't try local authentication and proceed directly to non-local authentication.

This is a bit clumsy-- I added a method to the core to look up a user's authenticator by string. But there's no way to map that to the actual authenticator object, because we only store one configured authenticator per core-- so we then just check if that authenticator is "Database" or not. I think this is something that should be improved in the future, but as a quick fix to #1501 this is probably good enough.

NOTE: I have not been able to test this, as I no longer have access to a LDAP server to run a test quassel core. So testing from someone who does would be greatly appreciated! (I did at least confirm database authentication still works).

### DIFF
--- a/src/core/core.h
+++ b/src/core/core.h
@@ -125,6 +125,16 @@ public:
         return instance()->_storage->getUserAuthenticator(userid) == authenticator;
     }
 
+    //! Gets the authenticator configured for a user.
+    /**
+     * \param userid The user's name as a QString.
+     * \return String value corresponding to the user's configure dauthenticator.
+     */
+    static inline QString getUserAuthenticator(const QString& userName)
+    {
+        return instance()->_storage->getUserAuthenticator(instance()->_storage->getUserId(userName));
+    }
+
     //! Change a user's password
     /**
      * \param userId     The user's ID
--- a/src/core/core.h
+++ b/src/core/core.h
@@ -127,12 +127,22 @@ public:
 
     //! Gets the authenticator configured for a user.
     /**
-     * \param userid The user's name as a QString.
+     * \param userName The user's name as a QString.
      * \return String value corresponding to the user's configure dauthenticator.
      */
     static inline QString getUserAuthenticator(const QString& userName)
     {
         return instance()->_storage->getUserAuthenticator(instance()->_storage->getUserId(userName));
+    }
+
+    //! Gets the user ID mapped to a username. This is necessary so that non-database auth methods can log in users properly.
+    /**
+     * \param userName The user's name as a QString.
+     * \return userId  The user's ID.
+     */
+    static inline UserId getUserId(const QString& userName)
+    {
+        return instance()->_storage->getUserId(userName);
     }
 
     //! Change a user's password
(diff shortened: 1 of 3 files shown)


finlstrm:
Finally had some time to test this, much closer, however moving an existing user to LDAP still doesn't work. Once switched to LDAP in the DB, It looks like the existing user is authenticating to LDAP but login fails with `Invalid login attempt from 172.17.0.1 as "finlstrm"`. I do find it odd that it is not giving `(LDAP authentication failed)` as it did previously as [mentioned in 1501](https://bugs.quassel-irc.org/issues/1501). Zeroing out the password gave the same error as previously mentioned. I don't think it's my LDAP configuration as new users are still authenticated to LDAP and added to quassel just fine.

TC01:
@finlstrm Huh. Would you mind pasting the full log from the core?

Can you confirm that you can log in as yourself if the account doesn't already exist in the database after switching to LDAP? I'm sure you've tried this at some point, but it might be good to check if you haven't.

Thanks again for testing! I'll have to think about what else could be causing this.

finlstrm:
@TC01, I have some debug output for you via '-d', please take a look at the below pastebins:

Successful login from DB user: https://pastebin.com/rmDZrCg9
Same user, converted to LDAP, unsuccessful: https://pastebin.com/axcAqhah
New LDAP user successful login: https://pastebin.com/WvjRLWxC

TC01:
@finlstrm I've just pushed another commit-- can you test it now?

I think the problem is that the core was trying to do local auth *after* LDAP auth succeeds-- this was done in order to look up the real quassel UserId of a user that already exists in the database. I passed the empty string (```QString()```) to the ```validateUser``` method in order to check that they had no password; but obviously, when you migrate a user to LDAP auth, there will be something else in the password column and so this comparison will fail.

So I think it is simpler to not do any of this and just have a new method in core.h to retrieve a UserId given a string username. LDAP auth should now succeed regardless of what's stored in a user's "password" column in the database.

If there was a script to do migrations, or if it existed as part of some core management function, it would be nice to somehow clear the password field that was stored in the database when moving a user to LDAP.

finlstrm:
@TC01 Awesome!!! Works great. Merged your code into my local repo with the latest from quassel:master and built core only. Tested another new core initially authenticating with database, then swapped to LDAP. This still requires a manual DB change per user to LDAP, but once that's made existing user is then authenticating to LDAP.

I should note that I did not touch the user's 'password' field, that didn't make quassel happy in my initial testing. I think something still needs to be in that field for LDAP users.

Also, I wrote a quick [bash] script to convert users. I put it up in a [public gist](https://gist.github.com/finlstrm/f8a945322227f9161e44a0d07c62379c).

PR #479, opened 2019-02-11
[WIP] Implements Parsing for IRCv3 message-tags, handles IRCv3 server-time and echo-message

justjanne:
## In Short

* Replaces the parsing code with more efficient, more robust code to also handle IRCv3 message-tags
* Implements the IRCv3 server-time capability
* Implements the IRCv3 echo-message capability

| Criteria | Rank | Reason |
| - | - | - |
| Impact | ★☆☆  _1/3_ | Improves IRC parsing for rare edge cases, adds minor functionality — the impact will come in follow-up PRs |
| Risk | ★★★  _3/3_ | Rewrites the entire IRC parsing, could break Quassel for some users |

## Rationale

IRCv3 message-tags are becoming more and more important for new IRCv3 functionality, and for many of the future features we’ve been planning on implementing, we need to handle them.

At the same time, our current parsing code, due to splitting at `" :"` (which now also appears before the prefix, as the tags come first), can’t handle them at all.

Rewriting the parsing code also made sense to clean it up, make each part more explicit, and clean up the spaghetti code of interactions.

Server-Time allows us to display the exact moment a message was actually received by the IRCd.
Echo-Message additionally allows us to show the message the exact way other users saw it, including splits, potential color stripping, and other modifications the IRCd might do.

## Testing

I’ve tested this on the Oragono Testnet (testnet.oragono.io) and the InspIRCd Testnet (testnet.inspircd.org), and everything has been working. I’ve also tested it in normal usage on Freenode to prevent regressions.

Due to the potential impact, I’d prefer if we could do a lot more testing.

### DIFF
--- a/src/common/irccap.h
+++ b/src/common/irccap.h
@@ -108,6 +108,13 @@ namespace IrcCap {
     const QString SERVER_TIME = "server-time";
 
     /**
+     * Server sending own message back
+     *
+     * https://ircv3.net/specs/extensions/echo-message-3.2.html
+     */
+    const QString ECHO_MESSAGE = "echo-message";
+
+    /**
      * Vendor-specific capabilities
      */
     namespace Vendor {
@@ -149,6 +156,7 @@ namespace IrcCap {
                                               SASL,
                                               USERHOST_IN_NAMES,
                                               SERVER_TIME,
+                                              ECHO_MESSAGE,
                                               Vendor::TWITCH_MEMBERSHIP,
                                               Vendor::ZNC_SELF_MESSAGE};
     // NOTE: If you modify the knownCaps list, update the constants above as needed.
--- a/src/common/irccap.h
+++ b/src/common/irccap.h
@@ -115,6 +115,20 @@ namespace IrcCap {
     const QString ECHO_MESSAGE = "echo-message";
 
     /**
+     * Draft capabilities
+     */
+    namespace Draft
+    {
+
+        /**
+         * Allows updating realname while being connected
+         *
+         * https://github.com/ircv3/ircv3-specifications/pull/361
+         */
+        const QString SETNAME = "draft/setname";
+    }
+
+    /**
      * Vendor-specific capabilities
      */
     namespace Vendor {
@@ -157,6 +171,7 @@ namespace IrcCap {
                                               USERHOST_IN_NAMES,
                                               SERVER_TIME,
                                               ECHO_MESSAGE,
+                                              Draft::SETNAME,
                                               Vendor::TWITCH_MEMBERSHIP,
                                               Vendor::ZNC_SELF_MESSAGE};
     // NOTE: If you modify the knownCaps list, update the constants above as needed.
--- a/src/common/eventmanager.h
+++ b/src/common/eventmanager.h
@@ -113,6 +113,7 @@ public:
         IrcEventWallops,
         IrcEventRawPrivmsg,  ///< Undecoded privmsg (still needs CTCP parsing)
         IrcEventRawNotice,   ///< Undecoded notice (still needs CTCP parsing)
+        IrcEventSetname,
         IrcEventUnknown,     ///< Unknown non-numeric cmd
 
         IrcEventNumeric = 0x00031000,     /* needs 1000 (0x03e8) consecutive free values! */
--- a/src/core/coresessioneventprocessor.cpp
+++ b/src/core/coresessioneventprocessor.cpp
@@ -791,6 +791,21 @@ void CoreSessionEventProcessor::processIrcEventError(IrcEvent* e)
     }
 }
 
+void CoreSessionEventProcessor::processIrcEventSetname(IrcEvent* e)
+{
+    if (checkParamCount(e, 1)) {
+        IrcUser* ircuser = e->network()->updateNickFromMask(e->prefix());
+        if (!ircuser) {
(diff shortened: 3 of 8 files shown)


digitalcircuit:
It'll be a while before I can fully review the IRC parser changes, but, meanwhile - one concern with `echo-message` is server lag will no longer be hidden.  Which is nice for confirmation of being sent, but also makes Quassel *seem* laggier than other clients, as connection to core versus core to IRC latency has the same appearance.

Should this be disabled until `message-ids`/whatever IRCv3 solution (I don't recall right now) is implemented, to distinguish a message between "sending" and "sent"?

Sput42:
Without having really looked at the code, one thing I've already noticed - we're attaching the &* qualifiers to the type name now, I'd appreciate if you could make that consistent in your contributions, too. I realize it's hard to get used to ;-)

arcadedroid:
This is the bug with day change messages:

day change bug

This is the bug with message duplication:

message duplication bug

src/core/corenetwork.cpp
```diff
@@ -1407,7 +1407,9 @@ void CoreNetwork::sendAutoWho()
             // And https://github.com/quakenet/snircd/blob/master/doc/readme.who
             // And https://github.com/hexchat/hexchat/blob/57478b65758e6b697b1d82ce21075e74aa475efc/src/common/proto-irc.c#L752
             putRawLine(serverEncode(
-                QString("WHO %1 n%chtsunfra,%2").arg(serverEncode(chanOrNick), QString::number(IrcCap::ACCOUNT_NOTIFY_WHOX_NUM))));
+                QString("WHO %1 n%chtsunfra,%2")
+                    .arg(chanOrNick, QString::number(IrcCap::ACCOUNT_NOTIFY_WHOX_NUM))
+            ));
```
digitalcircuit: Whoops.  Thanks for catching this!

src/core/ircparser.cpp
```diff
@@ -271,6 +271,13 @@ void IrcParser::processNetworkIncoming(NetworkDataEvent* e)
         qDebug() << "IRC net" << net->networkId() << "<<" << tags << prefix << cmd << params;
     }
 
+    IrcTagKey serverTimeTag{"", "time", false};
+    if (tags.contains(serverTimeTag)) {
+        QDateTime serverTime = QDateTime::fromString(tags[serverTimeTag], "yyyy-MM-ddThh:mm:ss.zzzZ");
```
digitalcircuit: This should check [if the `QDateTime` is valid](https://doc.qt.io/qt-5/qdatetime.html#isValid ) in case [`fromString()`](https://doc.qt.io/qt-5/qdatetime.html#fromString-1 ) fails (e.g. customized/non-compliant server), and fall back to the current date/time (is simply not setting the timestamp enough?), alongside at least a `qDebug()` to log this issue.

Considering that if this is an issue, it's likely to happen for every single message, it probably shouldn't be `qInfo()`.  Might also leak sensitive information.

src/common/irccap.h
```diff
@@ -114,6 +114,13 @@ namespace IrcCap {
      */
     const QString ECHO_MESSAGE = "echo-message";
 
+    /**
+     * Allows updating realname while being connected
+     *
+     * https://github.com/ircv3/ircv3-specifications/pull/361
+     */
+    const QString SETNAME = "draft/setname";
+
```
digitalcircuit: Should we have a separate `Draft` namespace, akin to `Vendor` capabilities?

Ideally everything would be removed from draft over time, so having that might help with tracking all draft specs, but also might be needless code change.

justjanne:
@Sput42 @digitalcircuit This PR is ready for review up until 4a26f75

arcadedroid:
Any chance this could be merged before it goes stale? It provides a lot of useful functionality.

justjanne:
@arcadedroid most of it has already been merged, this PR continues to stay open to continually add new functionality

arcadedroid:
@justjanne I just compiled latest master and I don't see `server-time`, has that functionality been merged or should I still use this branch for that?

justjanne:
@arcadedroid server_time should long be merged, it should work out of the box

arcadedroid:
@justjanne 
```
* Ready to negotiate (found: extended-join, away-notify, userhost-in-names, server-time, multi-prefix, sasl)
* Negotiating capabilities (requesting: sasl, away-notify, extended-join, multi-prefix, userhost-in-names)...
```

Using `v0.13.1 (dist-3778a12)` (https://github.com/quassel/quassel/commit/3778a12912369eb5add886bb65ca74e9df841744).

arcadedroid:
Ah damn, running `quasselcore` ran `/usr/local/bin/quasselcore` but systemd ran my distro's `/usr/bin/quasselcore`. My poop :P

digitalcircuit:
The currently open changes (`echo-message`, `setname`) have been [bundled into a new IRCv3 mega pull request!](https://github.com/quassel/quassel/pull/545 )

digitalcircuit:
With [the IRCv3 pull request #545 merged including `setname` and `echo-message` (disabled by default)](https://github.com/quassel/quassel/pull/545 ), I think this PR can be closed?  Unless you prefer to keep it for future testing (though a new PR might offer a cleaner slate).

Regardless, thank you again for the work you've done, @justJanne!

PR #480, opened 2019-02-11
cmake: fix build on MSYS2

rkitover:
Some minor changes to allow building easily on MSYS2:

- don't check for snorenotify on MSYS2 because it segfaults on startup

- don't use -fstack-protector-strong (libssp) on MinGW because it
currently causes segfaults during static initialization

https://sourceforge.net/p/mingw-w64/bugs/755/

- build libraries as static on Windows, the comments indicate that this
was the intention, but the check was not implemented

- use the `cat` command to read files when using the 'MSYS Makefiles'
generator, and not the cmd.exe `type` command

Tested to build correctly on MSYS2 with both the 'MSYS Makefiles'
generator and the Ninja generator using a native ninja.

Signed-off-by: Rafael Kitover <rkitover@gmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -207,19 +207,23 @@ if (BUILD_GUI)
         PURPOSE     "Required for audio notifications"
     )
 
-    find_package(LibsnoreQt5 0.7.0 QUIET)
-    set_package_properties(LibsnoreQt5 PROPERTIES TYPE OPTIONAL
-        URL "https://projects.kde.org/projects/playground/libs/snorenotify"
-        DESCRIPTION "a cross-platform notification framework"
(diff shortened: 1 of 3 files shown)


rkitover:
This will conflict with #466 but I will rebase whichever one is to be merged second.

Sput42:
Regarding Windows: actually, the comment is wrong here. It was written before I managed to get DLLs to build and link properly on that platform. Since that works now: we should honor the ENABLE_SHARED setting that you introduced, and just remove the reference to Windows from that comment.

rkitover:
@Sput42 rebased and updated function doc.

Thank you for taking a look!

PR #488, opened 2019-03-19
Implement a metrics daemon to expose Prometheus metrics

justjanne:
## In Short

* Adds an HTTP-based metrics API to expose certain metrics for Prometheus
* Adds an HTTP-based healthz API to allow container systems to determine that the core has started

| Criteria | Rank | Reason |
| - | - | - |
| Impact | ★☆☆  _1/3_ | Allows easier monitoring for use in automated deployments |

## Rationale

Exposing metrics allows monitoring of these, which also allows automated alerts in case something unexpected happens (e.g. a user on a shared core connecting to hundreds of networks, or having thousands of clients connected).

Additionally, this opens the way for including more metrics, such as exposing the remaining duration of the TLS certificate (which allows users to set a custom alert to get notified when it’s about to run out), as well as exposing a /healthz endpoint which allows container orchestration systems to figure out if a container is ready to accept connections.

### DIFF
--- a/src/core/metricsserver.cpp
+++ b/src/core/metricsserver.cpp
@@ -1,0 +1,408 @@
+/***************************************************************************
+ *   Copyright (C) 2005-2019 by the Quassel Project                        *
+ *   devel@quassel-irc.org                                                 *
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) version 3.                                           *
+ *                                                                         *
+ *   This program is distributed in the hope that it will be useful,       *
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of        *
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         *
+ *   GNU General Public License for more details.                          *
+ *                                                                         *
+ *   You should have received a copy of the GNU General Public License     *
+ *   along with this program; if not, write to the                         *
+ *   Free Software Foundation, Inc.,                                       *
+ *   51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.         *
+ ***************************************************************************/
+
+#include "metricsserver.h"
+
+#include <utility>
+
+#include <QByteArray>
+#include <QDebug>
+#include <QHostAddress>
+#include <QStringList>
+#include <QTcpSocket>
+
+#include "core.h"
+#include "corenetwork.h"
+
+MetricsServer::MetricsServer(QObject* parent)
+    : QObject(parent)
+{
+    connect(&_server, &QTcpServer::newConnection, this, &MetricsServer::incomingConnection);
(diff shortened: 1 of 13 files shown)


src/core/metricsserver.cpp
```diff
@@ -0,0 +1,271 @@
...
+        return;
+    }
+
+    int index = 0;
+    QString verb;
+    QString requestPath;
+    QString version;
+    QByteArray request;
+    for (int i = 0; i < 5 && verb == ""; i++) {
+        request = socket->readLine(2048);
```
winks: 4k or 8k might be safer, cf
http://nginx.org/en/docs/http/ngx_http_core_module.html#large_client_header_buffers 
http://httpd.apache.org/docs/2.4/mod/core.html#limitrequestfieldsize

justjanne: That would be true if we're expecting a browser to access this, but we know the only tools ever accessing this will be Prometheus and cURL.

src/core/metricsserver.cpp
```diff
@@ -0,0 +1,271 @@
...
+
+    if (requestPath == "/metrics") {
+        if (version == "HTTP/1.1") {
+            socket->write(
+                "HTTP/1.1 200 OK\r\n"
+                "Content-Type: text/plain\r\n"
+                "Connection: close\r\n"
+                "\r\n"
+            );
+        }
```
winks: What's supposed to be in else? HTTP2? No idea who would want to query with `HTTP/1.0` but I don't see anything complicated in the response that it shouldn't work. There's no `Host` header involved, but my memory is hazy on the other differences between 1.0 and 1.1

src/core/metricsserver.cpp
winks: Sure, that's reasonable.

src/core/metricsserver.cpp
justjanne: Fallback for older HTTP versions without any version string being passed is not returning any headers and not reading any headers.

This is based on how existing web servers handle such solutions.

winks: TIL, never stumbled upon anything like this, only more explicit handling.

justjanne:

src/core/metricsserver.cpp
```diff
@@ -0,0 +1,356 @@
...
+        for (const auto& key : _sessions.keys()) {
+            const QString& name = _sessions[key];
+            socket->write(
+                QString("quassel_client_sessions{user=\"%1\"} %2\n")
+                    .arg(name)
+                    .arg(QString::number((float) _clientSessions.value(key, 0)))
```
Sput42: Any particular reason you're casting this to a float? Does this result in a decimal point that is needed for Prometheus?

Btw, QString::arg() accepts numbers directly, so you should probably just write `.arg(_clientSessions.value(key, 0)`. This avoids the construction of a temporary string (which in this frequently called part of the code would be especially annoying).

Same is true below, as well.

justjanne: > Any particular reason you're casting this to a float? Does this result in a decimal point that is needed for Prometheus?

Yes, that’s basically the reason we do this.

PR #489, opened 2019-03-22
qtui: Set desktop file name

heftig:
The XCB platform plugin defaults the WM class to the executable name,
which already matches our desktop file name. Unfortunately, the Wayland
platform plugin prepends the inverted organization domain, resulting in
an app ID of "org.quassel-irc.quasselclient", thus breaking the
association.

Set the desktop file name explicitly so the Wayland platform doesn't get
confused.

### DIFF
--- a/src/main/monoapplication.cpp
+++ b/src/main/monoapplication.cpp
@@ -30,7 +30,11 @@ class InternalPeer;
 
 MonolithicApplication::MonolithicApplication(int& argc, char** argv)
     : QtUiApplication(argc, argv)
-{}
+{
+#if QT_VERSION >= 0x050700
+    QGuiApplication::setDesktopFileName(Quassel::buildInfo().applicationName);
+#endif
+}
 
 void MonolithicApplication::init()
 {
--- a/src/qtui/qtuiapplication.cpp
+++ b/src/qtui/qtuiapplication.cpp
@@ -36,6 +36,9 @@ QtUiApplication::QtUiApplication(int& argc, char** argv)
 #if QT_VERSION >= 0x050600
     QGuiApplication::setFallbackSessionManagementEnabled(false);
 #endif
+#if QT_VERSION >= 0x050700
+    QGuiApplication::setDesktopFileName(Quassel::buildInfo().clientApplicationName);
+#endif
 }
 
 void QtUiApplication::init()


kbroulik:
According to Qt documentation this must not have the `.desktop` suffix
> This property holds the base name of the desktop entry for this application

heftig:
> According to Qt documentation this must not have the `.desktop` suffix
> 
> > This property holds the base name of the desktop entry for this application

Base name means "without directory" not "without extension".

kbroulik:
In Qt context it is "The base name consists of all characters in the file up to (but not including) the first '.' character." and all code I checked expects it to not have a `.desktop` suffix.

heftig:
Look at [this test](https://github.com/qt/qtbase/blob/53d62b8fcbb639bd625777c8f1c01764445fb1c4/tests/auto/gui/kernel/qguiapplication/tst_qguiapplication.cpp#L140-L153):
```c++
void tst_QGuiApplication::desktopFileName()
{
    int argc = 1;
    char *argv[] = { const_cast<char*>("tst_qguiapplication") };
    QGuiApplication app(argc, argv);

    QCOMPARE(QGuiApplication::desktopFileName(), QString());

    QGuiApplication::setDesktopFileName("io.qt.QGuiApplication.desktop");
    QCOMPARE(QGuiApplication::desktopFileName(), QString::fromLatin1("io.qt.QGuiApplication.desktop"));

    QGuiApplication::setDesktopFileName(QString());
    QCOMPARE(QGuiApplication::desktopFileName(), QString());
}
```
It's definitely supposed to have the suffix.

heftig:
[The documentation](https://doc.qt.io/qt-5/qguiapplication.html#desktopFileName-prop) also says:
> This is the file name, without the full path, of the desktop entry that represents this application according to the freedesktop desktop entry specification.

[Here's some code from Octave which uses the suffix.](http://hg.savannah.gnu.org/hgweb/octave/file/4643682db06e/libgui/src/main-window.cc#l225)

---

[Here's some code from KDevelop which doesn't seem to use the suffix.](https://github.com/midenok/kdevelop/blob/edc1ab77d6fef8c8bdd425f807b5e9b959f57e7d/app/main.cpp#L364) But note this is `KAboutData::setDesktopFileName`, and [the documentation of this method](https://api.kde.org/frameworks/kcoreaddons/html/classKAboutData.html#a112d2fc20c31e7847995930e030cc67b) is noticeably different:
> This is the file name, without the full path and without extension, of the desktop entry that represents this application according to the freedesktop desktop entry specification (e.g. "org.kde.foo").

kbroulik:
It seems there's no real consensus here and judging from QtWayland code it strips the `.desktop` suffix off, so both ways are accepted by it.

kossebau:
For the record I filed https://bugreports.qt.io/browse/QTBUG-75521 to have this sorted out officially one day.

Sput42:
Can you guys please agree what we should do here now? I don't know enough about that topic to be able to decide...

kossebau:
IMHO, as stated in the filed qt bug report I linked above, you should be fine to simply follow KDE code behaviour and set a value without ".desktop" suffix. After all pretty all the KDE apps do that for years, without anything blowing up. Because actually the value without the suffix is the one used as identifier in any known protocols relying on it (like wayland).

But given quassel is not the only one also passing the suffix, and the only code known in Q using the value is QWaylandWindow::initWindow(), which chops off the suffix if present, as well as code present in Plasna/KDE Frameworks now having been updated to chop off any suffix as well (https://phabricator.kde.org/R289:eef093cd6adc1c62afba76944731c44e86a59922, 
https://phabricator.kde.org/R135:0e50f23c4c4344f86036c0a25c5c785d3f3354c0)  you might be also find to just keep what the patch does now.

When it comes to me: +1 for removing both ` + ".desktop"` from the patch.

Sput42:
I would agree then. @heftig could you change that please accordingly?

heftig:
@Sput42 done.

heftig:
Can this be merged now?

PR #490, opened 2019-03-23
improve reference for custom timestamp format

wolfpackmars2:
Provides reference to Qt::QDateTime::toString method used to convert timestamp
Custom timestamp format ToolTip text.


wolfpackmars2:
Unable to find where in the code the actual conversion occurs. Qt doesn't provide a day-of-year format. I propose using capital DD to indicate day of year format. Qt provides a dayOfYear method to the QDate object which could be used, but the capital DD would likely need to be manually parsed in code.

DD - day of year, no leading zeros
DDD - day of year including leading zeros

Nevcairiel:
Referencing Qt docs seems highly technical and possibly overwhelming for users clicking on that. If anything I think it should be a footnote at the end of the blurb, not the first thing.

wolfpackmars2:
removed the link. the tooltip is cluttered enough as is.

If someone is really interested in more information, they should be able to find it by oogling 'QDateTime'

digitalcircuit:
There's several other places that make use of `QDateTime` formatting, e.g the `Default Away` and `Away On Detach` settings for identities.

I agree with needing to document this better.  However, I'm not sure of the best way to do this.  Would adding an entry to [Quassel's Frequently Asked Questions](https://bugs.quassel-irc.org/projects/quassel-irc/wiki/FAQ ) be a good idea?  (How many power users actually check there?)

wolfpackmars2:
Yeah, or the wiki.
https://bugs.quassel-irc.org/projects/quassel-irc/wiki#Manuals

"Dates and time in Quassel"

Sput42:
I have rebased and squashed the commits into one, hope you don't mind!

PR #493, opened 2019-04-30
Set default action on notification

kbroulik:
This makes the entire notification popup clickable.
Also, mark the text for translation.

I couldn't find a minimum version requirement of KDE Frameworks for Quassel
but this method was added in Frameworks 5.32 in Early 2017 and supported
since Plasma 5.10. For older versions a regular button should be shown as is
since they will just not know what "default" means and take that as a regular
button action ID.


Sput42:
Minimum version currently is whatever is shipped in Ubuntu 16.04, which is why the CI builds fail for your change. So we need a version check and #ifdef this accordingly...

PR #494, opened 2019-05-27
Implement support for the HAProxy proxy protocol

justjanne:
## In Short

* Adds support for the HAProxy proxy protocol to the client-side auth handler
* Uses the forwarded IP information in the client list
* Adds a CLI option to specify which IP ranges are trusted proxies

| Criteria | Rank | Reason |
| - | - | - |
| Impact | ★☆☆  _1/3_ | Allows deploying Quassel behind load balancers |
| Risk | ★★☆  _2/3_ | Affects main protocol initialization code |

## Rationale

In cloud deployments, it’s common to deploy services behind load balancers or proxies. In such situations, it’s useful to still be able to show the original IP of the request.

Now that we show clients by IP in the client list, support for this protocol is useful.

### DIFF
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -41,6 +41,7 @@ target_sources(${TARGET} PRIVATE
     peerfactory.cpp
     presetnetworks.cpp
     quassel.cpp
+    proxyline.cpp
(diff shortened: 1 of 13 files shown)


justjanne:

PR #495, opened 2019-05-28
Fix protocol spam on part/quit/disconnect

justjanne:
## In Short
* Only transmit one message to part a channel instead of two per each user

| Criteria | Rank | Reason |
| - | - | - |
| Impact | ★☆☆  _1/3_ | Reduces protocol spam, reducing the perceived latency of some operations |
| Risk | ★☆☆  _1/3_ | Could trigger bugs in older clients |

## Rationale

When we leave a channel, we currently not only transmit the protocol message that we have left, but also a part (and potentially a quit) message for each user in that channel. Clients can synthesize these themselves, though, and already do.

As result, these messages get discarded by clients and only serve to fill the connection with spam.

### DIFF
--- a/src/common/ircchannel.cpp
+++ b/src/common/ircchannel.cpp
@@ -231,7 +231,7 @@ void IrcChannel::part(IrcUser* ircuser)
             _userModes.clear();
             foreach (IrcUser* user, users) {
                 disconnect(user, nullptr, this, nullptr);
-                user->partChannel(this);
+                user->partChannel(this, true);
             }
             emit parted();
             network()->removeIrcChannel(this);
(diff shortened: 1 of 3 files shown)


justjanne:

Sput42:
@digitalcircuit (or @justjanne), I'd appreciate some good testing involving different combinations of old/new clients and cores for this one...

justjanne:
@Sput42 Testing Matrix:

| --           | Core 0.13.1 | Core PR |
| ------------ | ----------- | ------- |
| Client 0.5.2 | :+1:        | :+1:    |
| Client 0.13  | :+1:        | :+1:    |
| Client PR    | :+1:        | :+1:    |

Any other combinations you’d like to see?

justjanne:
On that note, this is probably the most amazing screenshot in terms of backwards compatibility ever:

[Three generations of Quassel: 0.5, 0.13 and 0.14](https://i.k8r.eu/OBMM_A.png)

PR #496, opened 2019-05-28
Ignore ignored messages for highlights and buffer activity counts

justjanne:
## In Short

* Adds an "ignored" flag to messages in the database
* Sets it if the message is ignored
* Ignore such messages for highlightCount and bufferActivity

| Criteria | Rank | Reason |
| - | - | - |
| Impact | ★☆☆  _1/3_ | Fixes slightly wrong activity/highlight displays |
| Risk | ★★☆  _2/3_ | Could accidentally introduce protocol incompatibilities, includes a migration |
| Intrusiveness | ★★☆  _2/3_ | Includes a database migration |

## Rationale

Previously, we didn’t store on the core if a message was ignored, due to that we couldn’t include this information while compiling highlight counts and buffer activities on the core.

This has been fixed and highlight count and buffer activity information are now more accurate.

### DIFF
--- a/src/common/message.cpp
+++ b/src/common/message.cpp
@@ -34,7 +34,8 @@ Message::Message(BufferInfo bufferInfo,
                  QString senderPrefixes,
                  QString realName,
                  QString avatarUrl,
-                 Flags flags)
+                 Flags flags,
+                 bool ignored)
     : _timestamp(QDateTime::currentDateTime().toUTC())
     , _bufferInfo(std::move(bufferInfo))
     , _contents(std::move(contents))
@@ -44,6 +45,7 @@ Message::Message(BufferInfo bufferInfo,
     , _avatarUrl(std::move(avatarUrl))
     , _type(type)
     , _flags(flags)
+    , _ignored(ignored)
 {}
 
 Message::Message(QDateTime ts,
@@ -54,7 +56,8 @@ Message::Message(QDateTime ts,
                  QString senderPrefixes,
                  QString realName,
                  QString avatarUrl,
-                 Flags flags)
+                 Flags flags,
+                 bool ignored)
     : _timestamp(std::move(ts))
     , _bufferInfo(std::move(bufferInfo))
     , _contents(std::move(contents))
@@ -64,6 +67,7 @@ Message::Message(QDateTime ts,
     , _avatarUrl(std::move(avatarUrl))
     , _type(type)
     , _flags(flags)
+    , _ignored(ignored)
 {}
 
 QDataStream& operator<<(QDataStream& out, const Message& msg)
@@ -161,6 +165,7 @@ QDebug operator<<(QDebug dbg, const Message& msg)
                   << qPrintable(QString(", RealName:")) << msg.realName()
                   << qPrintable(QString(", AvatarURL:")) << msg.avatarUrl()
                   << qPrintable(QString(", Flags:")) << msg.flags()
+                  << qPrintable(QString(", Ignored:")) << msg.ignored()
                   << qPrintable(QString(")"))
                   << msg.senderPrefixes() << msg.sender() << ":"
                   << msg.contents();
--- a/src/common/message.h
+++ b/src/common/message.h
@@ -77,7 +77,8 @@ public:
             QString senderPrefixes = {},
             QString realName = {},
             QString avatarUrl = {},
-            Flags flags = None);
+            Flags flags = None,
+            bool ignored = false);
     Message(QDateTime ts,
             BufferInfo buffer = BufferInfo(),
             Type type = Plain,
@@ -86,7 +87,8 @@ public:
             QString senderPrefixes = {},
             QString realName = {},
             QString avatarUrl = {},
-            Flags flags = None);
+            Flags flags = None,
+            bool ignored = false);
 
     inline static Message ChangeOfDay(const QDateTime& day) { return Message(day, BufferInfo(), DayChange); }
     inline const MsgId& msgId() const { return _msgId; }
@@ -103,6 +105,8 @@ public:
     inline Type type() const { return _type; }
     inline Flags flags() const { return _flags; }
     inline void setFlags(Flags flags) { _flags = flags; }
+    inline bool ignored() const { return _ignored; }
+    inline void setIgnored(bool ignored) { _ignored = ignored; }
     inline const QDateTime& timestamp() const { return _timestamp; }
 
     inline bool isValid() const { return _msgId.isValid(); }
@@ -120,6 +124,7 @@ private:
     QString _avatarUrl;
     Type _type;
     Flags _flags;
+    bool _ignored;
 
     friend QDataStream& operator>>(QDataStream& in, Message& msg);
 };
--- a/src/core/SQL/PostgreSQL/insert_message.sql
+++ b/src/core/SQL/PostgreSQL/insert_message.sql
@@ -1,3 +1,3 @@
-INSERT INTO backlog (time, bufferid, type, flags, senderid, senderprefixes, message)
-VALUES ($1, $2, $3, $4, $5, $6, $7)
+INSERT INTO backlog (time, bufferid, type, flags, ignored, senderid, senderprefixes, message)
+VALUES ($1, $2, $3, $4, $5, $6, $7, $8)
 RETURNING messageid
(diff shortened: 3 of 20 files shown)


justjanne:

Sput42:
I think I may have asked this before, but why don't you just add this to Message::Flags? Introducing a schema change for this when we already have a perfectly valid flag field seems weird to me.

justjanne:
Because this allows proper indexing in the database later on, and I’m not sure if adding a flag would break protocol support.

digitalcircuit:
At the moment, `msg.ignored()` status is not sent over the protocol.  My added comment on this so far is…

```cpp
    // [...trimmed...]

    if (SignalProxy::current()->targetPeer()->hasFeature(Quassel::Feature::RichMessages)) {
        out << msg.realName().toUtf8();
        out << msg.avatarUrl().toUtf8();
    }

    // What about msg.ignored()?
    //
    // Dynamic message ignored status is not serialized as it does not communicate *which* ignore
    // rule resulted in the message being ignored.  That information is needed for a client to later
    // disable a specific dynamic ignore rule.
    //
    // For now, clients must apply all dynamic (and the dynamic aspect of permanent) ignore rules
    // locally, too, so there's no strong reason to communicate this.

    out << msg.contents().toUtf8();
    return out;
}
```

However, @justjanne , this seems like it'd be useful for an (interim) push notification client, and for Quassel bots - they wouldn't need to implement the entirety of the ignore rule logic.

Should I add a client feature flag to synchronize this?  I'd assume yes, but I'm open to not.

digitalcircuit:
I've implemented a few fixups and the protocol `Feature::MessageIgnoredStatus` in [`digitalcircuit/ft-justjanne-fix-ignore-activity`](https://github.com/quassel/quassel/compare/master...digitalcircuit:ft-justjanne-fix-ignore-activity).  Early tests using SQLite show promise, with the fix working and client/core compatibility working with new/new, new/old, and old/new.

I still need to run the full SQLite and PostgreSQL create/upgrade/migrate tests, including testing on Windows.  I doubt Windows is broken by this change, but testing to make sure Windows is still working is a good idea as we're nearing a new release.

Sput42:
I honestly still think that adding this to the existing flag rather than a new field and database column would make more sense, and probably also wouldn't require a new protocol feature, and no database migration. To me there seems to be no semantically sane reason why "ignored" should be handled differently from all the others.
I don't think we'll ever want to add a database index on the ignored flag really. I'm also not sure how much overhead adding a column like this incurs.

justjanne:
>  I don't think we'll ever want to add a database index on the ignored flag really. I'm also not sure how much overhead adding a column like this incurs.

This PR literally adds SQL queries querying for ignored != true, so we probably should add an index in this case anyway.

digitalcircuit:
@Sput42 Regarding expanding the `Message::Flags` enum, is this different from what [Implement Highlight-Only ignore #194](https://github.com/quassel/quassel/pull/194 ) did, [with `IgnoreListManager::StrictnessType`, and that broke the protocol](https://github.com/quassel/quassel/pull/194/files#r66396328 )?

That is a different enum, though, I'm not sure if this one is handled differently!

I'd prefer to test this to know for sure if it's broken or not, but I want to confirm what counts as "protocol not broken" - updating the core and sending a message with the new flag to a 0.13.1/before-this-change client?  And checking what happens if you roll back to an older core with messages containing the newer flag in the database?

---
Aside, if I can do it in time (no guarantees), I'm planning a database schema change and protocol addition for justJanne's `echo-message` support, so it can be turned off on misbehaving servers or if people dislike the extra latency it adds.

(In IRC → Network, rename `Encodings` tab to `Messages` tab, add checkbox to top for `Show my messages as delivered by the network`, with explanatory tooltip.  I might add that for `server-time`, too, [or attempt the more complex time sync that Kiwi IRC does](https://github.com/kiwiirc/kiwiirc/pull/1190 ).)

digitalcircuit:
Doing this via a `Message::Flag::Ignored` does not seem to break backwards compatibility.

I've [run the full testing on a new pull request, #541](https://github.com/quassel/quassel/pull/541 ) and I didn't encounter any issues.

Going between these pull requests, it's merely up to whether or not a new column is more efficient than filtering by another flag.  Is it possible to measure the efficiency differences?

Sput42:
> @Sput42 Regarding expanding the `Message::Flags` enum, is this different from what [Implement Highlight-Only ignore #194](https://github.com/quassel/quassel/pull/194) did, [with `IgnoreListManager::StrictnessType`, and that broke the protocol](https://github.com/quassel/quassel/pull/194/files#r66396328)?
> 
> That is a different enum, though, I'm not sure if this one is handled differently!
> 
> I'd prefer to test this to know for sure if it's broken or not, but I want to confirm what counts as "protocol not broken" - updating the core and sending a message with the new flag to a 0.13.1/before-this-change client? And checking what happens if you roll back to an older core with messages containing the newer flag in the database?

I honestly don't know what went wrong in #194. Unfortunately, while I still can compile the code on my box (with lots of warnings), I can't get it to sync properly with a remote core for some reason, so I can't easily reproduce.
Serialization of the Strictness enum goes through QVariant, and while I would expect that this would serialize as an integer rather than the enum value string, I can't rule that out without running more tests. However, the message flag is explicitly cast (by us) to an uint8, so as long as we don't add values larger than that, it should be fine.
Of course, both sides must be able to cope with the value not being set by an older peer, but that's true for both options... so testing would explicitly check for that (does ignore functionality still work as expected if either side does not set/understand the flag?).

Sput42:
> Doing this via a `Message::Flag::Ignored` does not seem to break backwards compatibility.
> 
> I've [run the full testing on a new pull request, #541](https://github.com/quassel/quassel/pull/541) and I didn't encounter any issues.
> 
> Going between these pull requests, it's merely up to whether or not a new column is more efficient than filtering by another flag. Is it possible to measure the efficiency differences?

I am mostly worried about actual disk usage, so it would be interesting to know for a large-ish database how much adding the extra column would increase storage size. We have users with hundreds of millions of entries, after all... then again, if we need a byte per row, then that's still only a MB per million rows, so.
Creating an index over such a column also increases disk space and slows down some operations (or so I've heard), which is why I'm sceptical that we would gain an advantage. If we add indexes for everything we might ever query for, do we still gain something? I really don't know enough about databases, so I can only trust my gut feeling and do appreciate the thoughts of actual experts (or benchmarks)!

Besides those hard technical arguments, I do have an aesthetical issue with introducing a new variable and method overloads for the sync methods when we already have a dedicated flags field whose purpose is literally to express such things. Looking at the proposed API immediately begs the question what makes "ignored" so special that we add a separate bool for it. On top of that, having bools in method signatures is really bad API in general - one should always use enums/flags for type safety and clarity (a lesson I've only learned later in my professional live, so we do have lots of `foo(true, false, false, true)` things in the code).

So, thinking about this, I would prefer to use the existing flag for API design reasons. If we find out that indexing the ignored flag (or any other) makes lots of sense for performance reasons, we can always later unpick the flag and store relevant values in a separate, indexed column on the storage layer, no matter how we choose to transmit it over the network.

digitalcircuit:
@Sput42 
> I honestly don't know what went wrong in #194. Unfortunately, while I still can compile the code on my box (with lots of warnings), I can't get it to sync properly with a remote core for some reason, so I can't easily reproduce.
>
> Serialization of the Strictness enum goes through QVariant, and while I would expect that this would serialize as an integer rather than the enum value string, I can't rule that out without running more tests. However, the message flag is explicitly cast (by us) to an uint8, so as long as we don't add values larger than that, it should be fine.

Thanks for taking a look!  Going forwards, we can test any `QVariant`-based enums if they're changed as needed, since the `Strictness` enum change is no longer needed with the other method of highlight ignore rules.  It'd be nice to know what happened, but it's probably not worth putting too much time into right now :slightly_smiling_face: 

If we run out of new message flag space in the future, justJanne's client/peer feature detection allow for serialization changes to fix this, too.

> Of course, both sides must be able to cope with the value not being set by an older peer, but that's true for both options... so testing would explicitly check for that (does ignore functionality still work as expected if either side does not set/understand the flag?).

And confirming, both this and #541 only modifies the core-side ignore handling.  The client never saw the `Message::Flags` set to `Ignored` before now.

I think the client may still want to apply ignore rules locally anyways, since it'll need to in order to toggle dynamic ignore rules off and on (the core isn't re-evaluating past messages and retransmitting them).  Still quite useful for simpler bots and push notification services.

> So, thinking about this, I would prefer to use the existing flag for API design reasons. If we find out that indexing the ignored flag (or any other) makes lots of sense for performance reasons, we can always later unpick the flag and store relevant values in a separate, indexed column on the storage layer, no matter how we choose to transmit it over the network.

I hadn't considered the separation of storage and network API, good point.  Thank you for explaining your reasoning, too.

PR #497, opened 2019-05-30
Sync last msgid

justjanne:
## In Short

* Fixes an issue where permanently merged buffers had wrong buffer activities / highlight counts
* Syncs the last message id per buffer via the buffersyncer

## Impact

| Criteria | Rank | Reason |
| - | - | - |
| Impact | ★☆☆  _1/3_ | Fixes slightly wrong activity/highlight displays and allows quasseldroid to provide new features |

## Rationale

In the future, we plan to resume chat wherever the user was in quasseldroid, also allowing users to jump to arbitrary points in the history (e.g. for fast-scrolling to a certain day, for jumping to search results or notifications).

To allow this, not only do we have to have bidirectional backlog requesting (which will come soon), but we also have to know where the currently last message id is (so that, once we actually have that latest message loaded, we stop dropping displayMsg calls, deactivate the archive UI, and automatically scroll chat again)

### DIFF
--- a/src/common/buffersyncer.cpp
+++ b/src/common/buffersyncer.cpp
@@ -26,17 +26,24 @@ BufferSyncer::BufferSyncer(QObject* parent)
     : SyncableObject(parent)
 {}
 
-BufferSyncer::BufferSyncer(QHash<BufferId, MsgId> lastSeenMsg,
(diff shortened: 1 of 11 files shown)


justjanne:

PR #498, opened 2019-05-30
[WIP] Allow peers to mark themselves or other peers as away

justjanne:
## In Short

* adds a /clientaway command to the qtclient to allow users to mark a client as away, it will not be counted towards active clients for detachaway
* adds a sync call to allow clients to mark themselves or other clients as away
* updates detachaway once a client sets itself as away

## Impact

| Criteria | Rank | Reason |
| - | - | - |
| Impact | ★☆☆  _1/3_ | Helps make detachaway work with always-connected mobile clients or bots |

## Rationale

Detach away is a useful and nice feature, it’s just relatively useless in the age of always-connected clients such as Quasseldroid, or if users are using tools such as the pushbullet notifier or quassel bots. Therefore, this PR introduces a way for clients to mark themselves as away, in a way that enables detachaway if all clients are gone or away.

### DIFF
--- a/src/common/peer.h
+++ b/src/common/peer.h
@@ -60,6 +60,9 @@ public:
     int id() const;
     void setId(int id);
 
+    bool away() const;
+    void setAway(bool away);
+
     AuthHandler* authHandler() const;
 
     virtual bool isOpen() const = 0;
@@ -94,6 +97,7 @@ public slots:
 
 signals:
     void disconnected();
+    void awayStateChanged();
     void secureStateChanged(bool secure = true);
     void lagUpdated(int msecs);
 
@@ -109,6 +113,8 @@ private:
     QString _buildDate;
     QString _clientVersion;
     Quassel::Features _features;
+
+    bool _away;
 
     int _id = -1;
 };
--- a/src/common/signalproxy.cpp
+++ b/src/common/signalproxy.cpp
@@ -658,6 +658,7 @@ QVariantList SignalProxy::peerData()
     for (auto&& peer : _peerMap.values()) {
         QVariantMap data;
         data["id"] = peer->id();
+        data["away"] = peer->away();
         data["clientVersion"] = peer->clientVersion();
         // We explicitly rename this, as, due to the Debian reproducability changes, buildDate isn’t actually the build
         // date anymore, but on newer clients the date of the last git commit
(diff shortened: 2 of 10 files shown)


justjanne:

src/common/peer.h
```diff
@@ -60,6 +60,9 @@ class COMMON_EXPORT Peer : public QObject
     int id() const;
     void setId(int id);
 
+    bool away() const;
+    void setAway(bool away);
+
```
digitalcircuit: **Minor:** Doxygen documentation here would be nice to clarify what this does.

src/common/signalproxy.cpp
```diff
@@ -658,6 +658,7 @@ QVariantList SignalProxy::peerData()
     for (auto&& peer : _peerMap.values()) {
         QVariantMap data;
         data["id"] = peer->id();
+        data["away"] = peer->away();
```
digitalcircuit: Should we add a simple line or icon to show if the client is inactive/away/idle in `Core Info`, and therefore won't count against Away On Detach?  The historical behavior is "if any clients connected, mark as available."

This might need distinguished if we use a new flag for whenever `KIdleTime`/etc is integrated.

*Aside, though stated in larger comment, if renaming, this should also be changed.*

src/core/coresession.h
```diff
@@ -151,6 +151,11 @@ public slots:
 
     void changePassword(PeerPtr peer, const QString& userName, const QString& oldPassword, const QString& newPassword);
 
+    /**
+     * Marks the own client as away
+     */
+    void markPeerAway(int peerId, bool away);
+
```
digitalcircuit: **Minor:** Doxygen documentation here would be nice to clarify what this does.  **Especially** with the magic value of `peerId = -1` to set the current peer as away.

Might be worthwhile considering if that should be a default argument value given how unlikely (?) it'll be for clients to set other clients as inactive.

(*I'm guessing that feature might be for a Quasseldroid push notification bot..?*)

src/common/signalproxy.cpp
```diff
@@ -672,6 +673,15 @@ QVariantList SignalProxy::peerData()
     return result;
 }
 
+bool SignalProxy::peersAllAway()
+{
+    for (auto&& peer : _peerMap.values()) {
```
digitalcircuit: Might be a good place for [`std::any_of`](https://en.cppreference.com/w/cpp/algorithm/all_any_none_of )?

**Untested** example:

```cpp
if (std::any_of(_peerMap.cbegin(), _peerMap.cend(), [](auto peer){ return !peer->away() })) {
    // At least one peer is not marked as away
    return false;
}
return true;
```

PR #501, opened 2019-07-05
Add konversation-like colour editing to the input and topic widgets in the client

colfrog:
There is currently no way to visualize or edit colours in the topic widget. This commit uses konversation's simple but efficient solution and adds support for it in the input widget. It uses konversation's code for the functions to replace markups and formatting codes.

I'm open to ideas on how to integrate it better with quassel or to other solutions.


justjanne:
This doesn’t correctly handle % chars — you convert %% to % in one direction, but % not to %% in the other.

Additionally, please handle and use all the standard formatting codes as specified in https://modern.ircdocs.horse/formatting.html

PR #504, opened 2019-07-26
Simplify the time formatting used in formatDateTimeToISO

justjanne:
## In Short

* Previously, a convoluted workaround using deprecated functionality was used to ensure we would format all types of QDateTime objects correctly
* A new workaround has been found which ensures this works reliably without using deprecated functionality

## Impact

| Criteria | Rank | Reason |
| - | - | - |
| Impact | ★☆☆  _1/3_ | Does not have any noticeable changes, fixes a compiler warning |
| Risk | ★☆☆  _1/3_ | Touches time code, but has unit tests |

## Rationale
Currently our builds for windows fail because of a compiler warning about using deprecated functionality in our previous workaround for time formatting.  
In the search for a way to fix this, this simpler workaround for the same issue has been found, and works reliably.

To ensure functionality remains the same, a unit test has been included. Please run it on various systems and versions of Qt to guarantee we won’t have any regressions.

### DIFF
--- a/src/common/util.cpp
+++ b/src/common/util.cpp
@@ -26,6 +26,7 @@
 
 #include <QCoreApplication>
 #include <QDateTime>
+#include <QTimeZone>
 #include <QDebug>
 #include <QTextCodec>
 #include <QVector>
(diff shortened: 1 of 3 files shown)


digitalcircuit:
Thank you for investigating this!  From a quick look, your approach seems correct.  I'll try to test soon (but don't wait forever for me).

[`QDateTime::toOffsetFromUtc()`](https://doc.qt.io/qt-5/qdatetime.html#toOffsetFromUtc ) was introduced in Qt 5.2, but thanks to dropping Qt 4 that doesn't matter :tada:

justjanne:
@digitalcircuit It seems to actually work now (after figuring out that travis does not have valid tzdata on its build containers)

PR #505, opened 2019-08-02
Add button and bindings for striking through text

AlexKraak:
Currently there are bindings for _italics_, **bold** and underline, but not ~strikethrough~, this adds one.


justjanne:
Your builds seem to fail because of a bug that was in quassel when you created your PR, could you try rebasing your PR?

digitalcircuit:
This is working great, thank you!

If you'd like to rebase, it'll fix the CI failures (Quassel's switched to GitHub actions with greater reliability), but it's not required.

Inspired by this, I'll try adding support for `monospace` in a follow-up PR, too.

digitalcircuit:
There's actually a small mistake…

```diff
 QString MultiLineEdit::convertMircCodesToHtml(const QString& text)
 {
     QStringList words;
-    QRegExp mircCode = QRegExp("(|||)", Qt::CaseSensitive);
+    QRegExp mircCode = QRegExp("(\x03|\x1f|\x02|\x1E)", Qt::CaseSensitive);
```

In order to fix up/down arrow history in the `MultiLineEdit` box.  I'm not sure why the raw values are in that QRegExp, but they need changed too.

(My attempts to add `Monospace` format have run into the space limits of `FormatType` in `UiStyle`, and I'm not sure how to address that.)

AlexKraak:
Oh cool, I'll rebase later today when I'm done work, I'd forgotten about this, thanks though!

digitalcircuit:
Thank you!

Looking at the Changes tab, when you have the time, the QRegExp still needs fixed - see my comment above, or this example commit: https://github.com/digitalcircuit/quassel/commit/e737a684b52132f733b7a7ab2032c7f39d411b15

Otherwise, the input history gets messed up when strikethrough is applied to part of a message, and up/down arrow is used to cycle through sent message history.  That regular expression splits on mIRC formatting codes, and it needs the strikethrough formatting code added, too.  It's very easy to overlook because it didn't use escaped values like everywhere else; I only noticed it when trying to add monospace support.

AlexKraak:
Oh ok, I see, I'll add this shortly

AlexKraak:
As far as I know, strikethrough on other clients (at leaest thelounge) works okay with this. I havent tested outside of it though.

PR #507, opened 2019-08-05
Fix deprecation warnings

justjanne:
## In Short

* removes and updates lots of outdated and deprecated code
* fixes compiler warnings on Qt 5.13

## Impact

| Criteria | Rank | Reason |
| - | - | - |
| Impact | ★☆☆  _1/3_ | Makes code more robust for future Qt versions |
| Risk | ★☆☆  _1/3_ | Code is not functionally different in any way |
| Intrusiveness | ★★★  _3/3_ | Changes throughout the whole codebase |

## Rationale

Our builds on Qt 5.13, as used on AppVeyor, fail currently entirely, leading to unreliable CI results. Additionally, using long-deprecated code is advised against.

This PR removes and replaces all deprecated code, making the codebase more robust for the future, as well as fixing compiler warnings

### DIFF
--- a/src/core/sqlitestorage.cpp
+++ b/src/core/sqlitestorage.cpp
@@ -192,7 +192,7 @@ UserId SqliteStorage::addUser(const QString& user, const QString& password, cons
         lockForWrite();
         safeExec(query);
         if (query.lastError().isValid()
-            && query.lastError().number() == 19) {  // user already exists - sadly 19 seems to be the general constraint violation error...
+            && query.lastError().nativeErrorCode() == "19") {  // user already exists - sadly 19 seems to be the general constraint violation error...
             db.rollback();
         }
         else {
@@ -1409,7 +1409,7 @@ bool SqliteStorage::renameBuffer(const UserId& user, const BufferId& bufferId, c
 
         error = query.lastError().isValid();
         // unexepcted error occured (19 == constraint violation)
-        if (error && query.lastError().number() != 19) {
+        if (error && query.lastError().nativeErrorCode() != "19") {
             watchQuery(query);
         }
         else {
@@ -1790,7 +1790,7 @@ bool SqliteStorage::logMessage(Message& msg)
 
         if (logMessageQuery.lastError().isValid()) {
             // constraint violation - must be NOT NULL constraint - probably the sender is missing...
-            if (logMessageQuery.lastError().number() == 19) {
+            if (logMessageQuery.lastError().nativeErrorCode() == "19") {
                 QSqlQuery addSenderQuery(db);
                 addSenderQuery.prepare(queryString("insert_sender"));
                 addSenderQuery.bindValue(":sender", msg.sender());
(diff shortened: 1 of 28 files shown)


TheOneRing:
As range loops are a bit dangerous (detaching) with Qt, could you compile once with clazy to make sure everything is right?

In the long run we probably should use clazy on the ci.

justjanne:
@TheOneRing Already working on that (that’s gonna be the next step), but I’m still looking into the most optimal solution for this, as `qAsConst` isn’t available in Qt5.5 (which we target), `Q_FOREACH` and `foreach` are deprecated, and always making a copy isn’t optimal either.

justjanne:
@TheOneRing I’ve moved the loop changes into a new branch for now, and will work on them later on, for now it’s more important to get these smaller issues fixed so we build without warnings on Qt 5.13 again

AlD:
There's more deprecated stuff in Apple land in case you're interested in fixing that, too: https://travis-ci.org/quassel/quassel/jobs/567613050

justjanne:
@AlD Except for `CSCopyUserName`, `GetFrontProcess`, `SetFrontProcess`, `ShowHideProcess`, they’re all already fixed with this. See the warnings at https://travis-ci.org/quassel/quassel/jobs/567974876.

The other issues there are OSX-specific, and as I don’t have a working OSX installation, I can’t test locally, so I won’t work on that.

src/core/sqlitestorage.cpp
```diff
@@ -1409,7 +1409,7 @@ bool SqliteStorage::renameBuffer(const UserId& user, const BufferId& bufferId, c
 
         error = query.lastError().isValid();
         // unexepcted error occured (19 == constraint violation)
-        if (error && query.lastError().number() != 19) {
+        if (error && query.lastError().nativeErrorCode() != "19") {
```
TheOneRing: Maybe QStringLiteral or QLatin1String?

src/qtui/settingspages/itemviewsettingspage.cpp
```diff
@@ -43,11 +42,11 @@ ItemViewSettingsPage::ItemViewSettingsPage(QWidget* parent)
 
     ui.bufferViewPreview->expandAll();
 
-    foreach (ColorButton* button, findChildren<ColorButton*>()) {
-        connect(button, &ColorButton::colorChanged, _mapper, selectOverload<>(&QSignalMapper::map));
-        _mapper->setMapping(button, button);
+    for (ColorButton* button : findChildren<ColorButton*>()) {
+        connect(button, &ColorButton::colorChanged, [this, button]() {
```
TheOneRing: never use the overload of connect that doesn't take a receiver as the life time of the lambda is not defined

justjanne: This is the officially recommended solution by Qt to replace the QSignalMapper...

PR #513, opened 2019-10-19
Install notifyrc on all platforms if WITH_KF5 is True

TheOneRing:


justjanne:
This seems to fail on some platforms – could you check this out, and try to make the builds pass? It seems like this is caused by inconsistent icon names, or missing icons on *nix, and by an entirely different error on Windows:

```
RCC: Error in '/src/icons/hicolor_icons.qrc': Cannot find file 'breeze/actions/24/connect-quassel.svg'
```

```
ImportError: No module named gn_helpers
```

TheOneRing:
@Sput42 please approve again now that the ci is green

PR #514, opened 2019-10-19
Use correct data Path on Windows

TheOneRing:
I tried QStandardPaths but the application name is not quassel but "Quassel Project" ....


justjanne:
This seems to fail with the same issue as https://github.com/quassel/quassel/pull/513#issuecomment-562087357, could you try rebasing both PRs?

TheOneRing:
@Sput42 ping

PR #521, opened 2020-02-01
Update aliasmanager.cpp

PeGaSuS-Coder:
I propose to change the alias expanded format from `/msg` to `/quote` since this is probably the most widely used method by other IRC clients and will also make them work out of the box with DALnet that requires that commands to NickServ (as example) to be sent explicitly as `/msg NickServ@services.dal.net` to work.

I've tested this myself in DALnet, Freenode and some other networks with different IRCd flavours and all worked without any problems.

This change proposal shouldn't interfere with previous/future PRs and gives user a more consistent aliases examples.

Cheers


Sput42:
I rebased and edited the commit message to adhere to our guidelines. Thanks for the contribution!

Sput42 (approved):

PR #523, opened 2020-03-13
Fix typo in spanish translation.

perezmeyer:
A simple typo fix.


AlD:
Thanks for contributing!
Translations are managed in https://www.transifex.com/quassel/quassel/. It causes merge issues if done outside of there, so please head over to transifex and apply the fix there. Thanks!

perezmeyer:
Uff, it requires me to create a user and then I need to set it up to forget me after a simple fix... Sorry, I'll file a bug then.

PR #524, opened 2020-03-15
Require SSL support in Qt, and remove fallback code

Sput42:
There is no sane reason for not using SSL in the year 2020, plus every distro should be shipping the Qt libraries with SSL support enabled. Thus, make it mandatory for Quassel, too.

This allows us to remove lots of fallback code paths that make the code much more complex in many places, and that have not been tested (or even been built) in a long time.

Thanks to @justjanne for doing the heavy lifting of removing `HAVE_SSL`-protected fallback code throughout the whole code base!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -179,6 +179,25 @@ set_package_properties(Qt5 PROPERTIES TYPE REQUIRED
 )
 message(STATUS "Found Qt ${Qt5Core_VERSION}")
 
+# Check for SSL support in Qt
+cmake_push_check_state(RESET)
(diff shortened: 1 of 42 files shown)


Sput42:
@digitalcircuit (or anyone else, of course), if you find the time I'd appreciate some basic testing (handshake/connection) with legacy and new versions of core and client, with and without SSL enabled, so that we know that I didn't break anything...

justjanne:
I’m honestly surprised about this commit from myself, I can’t remember writing that :sweat_smile:

I’ll take a look at compatibility later.

PR #531, opened 2020-06-06
Start a CocoaHelper class for macOS 10.12 and add a few improvements to the client

colfrog:
Hi, this fork has been lying around for a while and it brings some pretty good improvements to macOS 10.12 and later, as well as the start of macOS-specific code via the CocoaHelper class. I wanted to clean up the code and the commits, but some things got in the way and I forgot about this repository for a while.

Sadly, I don't have a Mac anymore and I don't plan to buy one in the near future. But if I ever do or if I eventually setup a macOS virtual machine, I'll definitely revisit it. The CocoaHelper class could be given its own pull request, but it should work right now, and I don't have something to test it on if I play with the code.

This is from my original commit:

* Added the start of a CocoaHelper class to customize how windows look. things in there should be turned into settings and piled up in the future.
* Removed the "stop editing on unfocus" behaviour the topic widget had
* Only the highest mode prefix shows in front of a nick when it's enabled. Ideally the core should sort them and there should be an attempt to grab the first one. This is an improvement that will probably be made later.
* ~Added makeHumanReadable and makeIrcReadable from Konversation's replaceFormattingCodes and replaceIRCMarkups. It was integrated in the topic widget, so that style bytes are converted to and from human-readable format when the topic is being edited. It was also integrated in the input widget.~
* Added the Edit menu from PR #176 on quassel's repo as a function to cocoahelper
* Fix emojis on macOS by leaving a tiny bit of space between the lines

I've reverted "Added makeHumanReadable and makeIrcReadable from Konversation's "replaceFormattingCodes and replaceIRCMarkups.", I'm still working on this on a separate branch.

### DIFF
--- a/src/common/message.cpp
+++ b/src/common/message.cpp
@@ -132,9 +132,20 @@ QDataStream& operator>>(QDataStream& in, Message& msg)
     in >> sender;
     msg._sender = QString::fromUtf8(sender);
 
-    QByteArray senderPrefixes;
+    QByteArray receivedPrefixes;
     if (SignalProxy::current()->sourcePeer()->hasFeature(Quassel::Feature::SenderPrefixes))
-        in >> senderPrefixes;
+        in >> receivedPrefixes;
+
+    QByteArray prefixes = "!~@%+";
+    QByteArray senderPrefixes;
+    for (char c : prefixes) {
+       if (receivedPrefixes.indexOf(c) != -1) {
+           senderPrefixes[0] = c;
+           senderPrefixes[1] = '\0';
+           break;
+       }
+    }
+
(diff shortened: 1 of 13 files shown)


digitalcircuit:
Belatedly, thank you for looking into this!

I'll have to figure out an (ahem) totally-legit macOS setup to properly review this, but, meanwhile, I appreciate the attention to macOS detail.  Fortunately, stock Quassel now has the "highest mode prefix" option, and the other aspects can be brought forward.

colfrog:
Thanks for your interest, I like the idea that I had with the CocoaHelper class, as it allows for some nice and sometimes needed tweaks on macOS without much effort.

I recently had the urge to fix up my 2011 MBP one last time, I will revisit this if I do.

PR #537, opened 2020-06-19
common: Default to SSL/TLS for new IRC networks

digitalcircuit:
## In short
* Enable SSL/TLS by default for new IRC networks
  * Existing networks will stay as they are
  * SSL/TLS can be disabled by editing the network
  * Default port will revert to `6667` when unchecking `Use encrypted connection`
  * Client-side change; core has no impact
* Fix up Doxygen comments for `Server::useSSL` and `PortDefaults`

Criteria | Rank | Reason
Impact | ★★☆ *2/3* | User-facing better security by default
Risk | ★★★ *3/3* | May frustrate, confuse those who don't know to try disabling SSL/TLS
Intrusiveness | ★☆☆ *1/3* | UI/defaults changes, shouldn't interfere with other pull requests

## Rationale
It's the year 2020, and several networks offer SSL/TLS with valid certificates.  Quassel could default to enabling SSL/TLS for newly added networks.

However, a few large networks still do not support SSL/TLS on port 6697, alongside locally-hosted bridges, such as Bitlbee.  This will require unchecking `Use encrypted connection` to connect those networks without SSL/TLS.

All preset networks will work as before, according to the preset (most have SSL/TLS ports added already).

**Subjective choice.  This may result in increased support requests in `#quassel`.**

When SSL/TLS is not available on the configured port, Quassel will fail to connect to the IRC network with an error message (i.e. `Connection refused`).  This can be remedied by editing the network and unchecking `Use encrypted connection`, but that's an extra step that might put off new users.

**There are other, more complex options**, [such as auto-detecting if an SSL/TLS port is available](https://github.com/ircdocs/best-practices/blob/master/client/tls-guidelines.md), or [supporting the IRCv3 `sts` capability](https://ircv3.net/specs/extensions/sts ).

Opinions in `freenode`/`#ircv3` are mixed - some clients are doing this already without reported issues, others are in favor of the more complex methods.

### Before
**Add Network dialog**
*On the Settings - IRC - Networks page under the `Networks` heading, `Add..` button clicked to add a new network to an existing network.*

Add Network dialog, showing a "Port" of "6667" and "Use encrypted connection" unchecked by default

**Server Edit dialog**
*On the Settings - IRC - Networks page under the `Servers` heading, `Add..` button clicked to add a new server to an existing network.*

Edit Server dialog, showing a "Port" of "6667" and "Use encrypted connection" unchecked by default

### After
**Add Network dialog**
*On the Settings - IRC - Networks page under the `Networks` heading, `Add..` button clicked to add a new network to an existing network.*

Add Network dialog, showing a "Port" of "6697" and "Use encrypted connection" checked by default

**Server Edit dialog**
*On the Settings - IRC - Networks page under the `Servers` heading, `Add..` button clicked to add a new server to an existing network.*

Edit Server dialog, showing a "Port" of "6697" and "Use encrypted connection" checked by default

### DIFF
--- a/src/common/network.h
+++ b/src/common/network.h
@@ -110,22 +110,22 @@ public:
         D_CHANMODE = 0x08
     };
 
-    // Default port assignments according to what many IRC networks have settled on.
-    // Technically not a standard, but it's fairly widespread.
-    // See https://freenode.net/news/port-6697-irc-via-tlsssl
+    /// Default port assignments according to what many IRC networks have settled on.
+    /// Technically not a standard, but it's fairly widespread.
+    /// See https://freenode.net/news/port-6697-irc-via-tlsssl
     enum PortDefaults
     {
-        PORT_PLAINTEXT = 6667,  /// Default port for unencrypted connections
-        PORT_SSL = 6697         /// Default port for encrypted connections
+        PORT_PLAINTEXT = 6667,  ///< Default port for unencrypted connections
+        PORT_SSL = 6697         ///< Default port for encrypted connections
     };
 
     struct Server
     {
         QString host;
-        uint port{6667};
+        uint port{PortDefaults::PORT_SSL};
         QString password;
-        bool useSsl{false};
-        bool sslVerify{true};  /// If true, validate SSL certificates
+        bool useSsl{true};     ///< If true, connect via SSL/TLS
+        bool sslVerify{true};  ///< If true, validate SSL certificates
         int sslVersion{0};
 
         bool useProxy{false};
--- a/src/qtui/settingspages/networkadddlg.ui
+++ b/src/qtui/settingspages/networkadddlg.ui
@@ -105,7 +105,7 @@
            <number>65535</number>
           </property>
           <property name="value">
-           <number>6667</number>
+           <number>6697</number>
           </property>
          </widget>
         </item>
@@ -132,6 +132,9 @@
         </property>
         <property name="text">
          <string>Use encrypted connection</string>
+        </property>
+        <property name="checked">
+         <bool>true</bool>
         </property>
        </widget>
       </item>
--- a/src/qtui/settingspages/servereditdlg.ui
+++ b/src/qtui/settingspages/servereditdlg.ui
@@ -58,7 +58,7 @@
             <number>65535</number>
            </property>
            <property name="value">
-            <number>6667</number>
+            <number>6697</number>
            </property>
           </widget>
          </item>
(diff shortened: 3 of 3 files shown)


digitalcircuit:
One course of action:

* Add boolean column to `ircserver` table called `sslprobed`
  * Default to `false`
* Make use of the existing code to update/set network info to store this
  * Inefficient, but should only happen once, and the server list doesn't have unique IDs at the moment
  * IRC server details may need updated anyways, too (port and connection settings)

Then…
1.  On `CoreNetwork` connect after selecting an IRC server, check `sslProbed` for the selected IRC server
     * If `sslProbed == true`, connect with current settings as before
     * If `sslProbed == false`, then…
2.  Check IRC server `useSsl` / "use encrypted connection" option
     * If already enabled, set `sslProbed = true`, connect with current settings as before
3.  Check IRC port number
     * If not `PortDefaults::PORT_PLAINTEXT`, connect with current settings as before
     * Intentionally *not* setting the `sslProbed` flag so if someone reverts to the default plaintext port in the future, SSL/TLS will get auto-attempted once
4.  Safe to probe.  Set `sslProbed = true`
     * If there's somehow a crash or hang from connecting to SSL/TLS port, Quassel won't get stuck
5.  Set temporary flag `probingSsl = true` (or such) in `CoreNetwork`
6.  Connect to the chosen IRC server over `PortDefaults::PORT_SSL` with `useSsl` enabled and `sslVerify` active
     * Making use of Quassel's default IRC network ping timeout detection for timing out
     * Put a message in the status buffer after existing connection message (`* Connecting to chat.freenode.net:6667...`)
     * Example message: `* Checking if encrypted connection available at chat.freenode.net:6697…`
7.  On successful connect (reached point of sending `CAP LS 302`), modify the IRC server to `PortDefaults::PORT_SSL` with `useSsl` enabled and `sslVerify` active, set `probingSsl = false`
    * Example message: `* Encrypted connection to chat.freenode.net:6697 successful, saved in settings`
8.  On ping timeout, SSL/TLS error, or **any** disconnect when `probingSsl` is set to `true`, emit message and auto-reconnect once
    * Example message: `* Encrypted connection to chat.freenode.net:6697 not successful, reverting to plain text...`
    * Trigger auto-reconnect even if normally disabled (this is a once-off per IRC server)
    * As `sslProbed` was previously set to `true`, a standard reconnect will skip this

digitalcircuit:
Okay, so… HexChat is doing this by default.  Time to consider Quassel switching as well..?

https://github.com/hexchat/hexchat/commit/747a52aae8806a9072a23ea68212767f352ac431 (thanks to `Sadie` in `#ircv3` for pointing this out, [pull request here](https://github.com/hexchat/hexchat/pull/2587 )).

`sts` would still be nice to support, of course.

PR #548, opened 2020-09-06
Keep editing topic while someone talks

lstarnes1024:
Without this change, it's almost impossible to edit the topic with the
topic widget in busy channels because every message would trigger the
dataChanged signal which would reset the topic and switch the topic
widget out of editing mode.

Fixes #1485.

### DIFF
--- a/src/qtui/topicwidget.cpp
+++ b/src/qtui/topicwidget.cpp
@@ -152,13 +152,17 @@ void TopicWidget::setTopic(const QModelIndex& index)
         }
     }
 
-    _topic = sanitizeTopic(newtopic);
-    _readonly = readonly;
+    QString sanitizedNewTopic = sanitizeTopic(newtopic);
+    if (readonly != _readonly || sanitizedNewTopic != _topic)
+    {
+        _topic = sanitizedNewTopic;
+        _readonly = readonly;
 
-    ui.topicEditButton->setVisible(!_readonly);
-    ui.topicLabel->setText(_topic);
-    ui.topicLineEdit->setPlainText(_topic);
-    switchPlain();
+        ui.topicEditButton->setVisible(!_readonly);
+        ui.topicLabel->setText(_topic);
+        ui.topicLineEdit->setPlainText(_topic);
+        switchPlain();
+    }
 }
 
 void TopicWidget::setReadOnly(const bool& readonly)


src/qtui/topicwidget.cpp
```diff
@@ -152,13 +152,17 @@ void TopicWidget::setTopic(const QModelIndex& index)
...
 
-    _topic = sanitizeTopic(newtopic);
-    _readonly = readonly;
-
-    ui.topicEditButton->setVisible(!_readonly);
-    ui.topicLabel->setText(_topic);
-    ui.topicLineEdit->setPlainText(_topic);
-    switchPlain();
+    QString sanitizedNewTopic = sanitizeTopic(newtopic);
+    if (ui.stackedWidget->currentIndex() != 1 || sanitizedNewTopic != _topic)
```
digitalcircuit, line 156: Not tested, but it might be fine to simply check if `_readonly != readonly` instead.  Reason being, if there's no changes to topic or readonly state, there shouldn't be any reason to set the topic, change the widget order, or set the edit button visibility.

```suggestion
    if (readonly != _readonly || sanitizedNewTopic != _topic)
```

(I'm surprised this function is even called on every new message; it feels like the `dataChanged` event is being too noisy.)

digitalcircuit (commented):
I haven't had the time to build and test this yet (will try to do that soon), but so far, it looks fine!  It might even be possible to optimize it a bit further to further avoid unnecessary updates.

Regardless, thank you for your contribution, including the proper [`Fixes #1485`](https://bugs.quassel-irc.org/issues/1485 ) line, too!

src/qtui/topicwidget.cpp
lstarnes1024: I just tested this and it looks like it works.

digitalcircuit (approved):
## Testing 2020-9-7 - ☑ pass
### Notes
* Tests show no regression in behavior
* `TopicWidget::dataChanged()` event fires for **any** change to buffer model, as noted by author
  * Appears to be correct behavior; having `TopicWidget` do the change checking seems to fit
* Tested on Kubuntu 20.04 LTS

### Test cases
Symbol | Meaning
---------|----------
🏁 | Existing behavior
🆕 | New/fixed behavior

* Entering topic edit mode, then changing topic after a delay
  * E.g. `/wait 5; /topic NewTopicHere`
  * 🏁 New topic shown, correctly exiting edit mode
* Entering topic edit mode, then receiving new message
  * E.g. `/wait 5; A sample message`
  * 🆕 Topic widget correctly remains in edit mode
* Editing topic, applying changes with <kbd>Enter</kbd>
  * 🏁 New topic is applied as if running `/topic <new topic>`
* Entering topic edit mode, using mouse to switch buffers
  * 🏁 New topics and read-only status (edit button visibility) correctly update
* Entering topic edit mode, using keyboard up/down arrow to switch buffers
  * 🏁 New topics and read-only status (edit button visibility) correctly update

*Note: `Sput` will need to do the final review and merge; you can check in on `freenode`/`#quassel`, too.  I don't foresee any problems, you may just need to wait a while as life's been busy for the project maintainer.  Thank you for your first pull request!*

digitalcircuit:
I've also merged this into my personal [`0.14rc1`-preparation branch `wip-prod-pr-tests`](https://github.com/quassel/quassel/compare/master...digitalcircuit:wip-prod-pr-tests ), so I'll be testing this client-side until `Sput` has time to review/merge PRs :slightly_smiling_face:

Sput42:
I took the liberty to rebase and squash the fixup commits. Thanks for your contribution!

PR #552, opened 2020-11-22
remove Boost dependency in favor of C++17

GravisZro:
This removes the needless Boost library requirement for Quassel by using std::optional from C++17 instead of Boost::optional.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -356,20 +356,6 @@ endif()
 # Non-Qt-based packages
 #####################################################################
 
-find_package(Boost 1.54 REQUIRED)
-set_package_properties(Boost PROPERTIES TYPE REQUIRED
-    URL "https://www.boost.org/"
-    DESCRIPTION "Boost libraries for C++"
-)
-# Older versions don't define the imported target
-if (NOT TARGET Boost::boost)
-    add_library(Boost::boost INTERFACE IMPORTED GLOBAL)
-    if (Boost_INCLUDE_DIRS)
-        set_target_properties(Boost::boost PROPERTIES
-            INTERFACE_INCLUDE_DIRECTORIES "${Boost_INCLUDE_DIRS}")
-    endif()
-endif()
-
 find_package(ZLIB REQUIRED)
 set_package_properties(ZLIB PROPERTIES TYPE REQUIRED
     URL "http://www.zlib.net"
--- a/cmake/QuasselCompileSettings.cmake
+++ b/cmake/QuasselCompileSettings.cmake
@@ -20,7 +20,7 @@ function(check_and_set_linker_flag flag name outvar)
 endfunction()
 
 # General compile settings
-set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED OFF)    # Rely on compile features if standard is not supported
 set(CMAKE_CXX_EXTENSIONS OFF)           # We like to be standard conform
 
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -67,7 +67,6 @@ target_include_directories(${TARGET} PRIVATE ${CMAKE_BINARY_DIR})
 
 target_link_libraries(${TARGET} PUBLIC
     ${CMAKE_DL_LIBS}
-    Boost::boost
     Qt5::Core
     Qt5::Network
     ZLIB::ZLIB
--- a/src/common/funchelpers.h
+++ b/src/common/funchelpers.h
@@ -26,7 +26,7 @@
 #include <type_traits>
 #include <utility>
 
-#include <boost/optional.hpp>
+#include <optional>
 
 #include <QDebug>
 #include <QVariant>
@@ -97,14 +97,14 @@ auto invokeWithArgs(const Callable& c, Args&&... args)
 
 // Helper for unpacking the argument list via an index sequence
 template<typename Callable, std::size_t ...Is, typename ArgsTuple = typename FunctionTraits<Callable>::ArgsTuple>
-boost::optional<QVariant> invokeWithArgsList(const Callable& c, const QVariantList& args, std::index_sequence<Is...>)
+std::optional<QVariant> invokeWithArgsList(const Callable& c, const QVariantList& args, std::index_sequence<Is...>)
 {
     // Sanity check that all types can be converted
     std::array<bool, std::tuple_size<ArgsTuple>::value> convertible{{args[Is].canConvert<std::decay_t<std::tuple_element_t<Is, ArgsTuple>>>()...}};
     for (size_t i = 0; i < convertible.size(); ++i) {
         if (!convertible[i]) {
             qWarning() << "Cannot convert parameter" << i << "from type" << args[static_cast<int>(i)].typeName() << "to expected argument type";
-            return boost::none;
+            return std::nullopt;
         }
     }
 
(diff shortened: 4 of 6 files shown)


Sput42:
We can only start using C++17 after the next release (which hopefully still happens this year). Once we've bumped build requirements, we can (and should) switch to `std::` equivalents.

However, I have some things pending that require some boost preprocessor features, so I doubt that we can get rid of the dependency completely, nor am I sure that it's really that big of a deal to make use of Boost where Qt or `std` don't provide something?

GravisZro:
> We can only start using C++17 after the next release (which hopefully still happens this year). Once we've bumped build requirements
Well... you can add C++17 features to C++14 compilers: https://github.com/akrzemi1/Optional

>  I have some things pending that require some boost preprocessor features

Arg! Why would you do that?

Sput42:
> > We can only start using C++17 after the next release (which hopefully still happens this year). Once we've bumped build requirements
> 
> Well... you can add C++17 features to C++14 compilers: https://github.com/akrzemi1/Optional

That sounds scarier than using Boost tbh...

> > I have some things pending that require some boost preprocessor features
> 
> Arg! Why would you do that?

Mostly for simplifying the use of variadic macros in the "new" SyncableObject implementation: https://github.com/quassel/quassel/pull/447/commits/e16d122cb405f332852294456052d46255e47a97#diff-2f59627dd4884d0af9891643c9e4c04dfb988afad86a6b3589274e8e2efceb2dR80

I will have to revisit the whole thing sometime after the next release, when we also enable C++17, but I don't see to get the introspection features required without the use of macros somehow (or continuing to rely on undocumented Qt internals and runtime lookups).

If I may ask @GravisZro: why are you opposed to depending on Boost? Seems to me the better choice compared to bundling 3rd party code or reinventing the wheel ourselves... and it's (so far) header-only libraries, too.

GravisZro:
> I don't see to get the introspection features required without the use of macros somehow (or continuing to rely on undocumented Qt internals

Are you thinking of something specific because introspection can be used with anything that's a QObject.
Qt has the Meta-Object system: https://doc.qt.io/qt-5/metaobjects.html
Really just calling QObject::metaObject() gets you a [QMetaObject](https://doc.qt.io/qt-5/qmetaobject.html)

> If I may ask @GravisZro: why are you opposed to depending on Boost?

- It's asinine to require a massive library to enable a single convenience feature that is in the C++ standard.
- It's a pain to have to install boost libraries and their corresponding dev packages just to compile Quassel.
- boost seems to have a new package for each version (e.g. libboost-regex1.71.0 vs libboost-regex1.74.0).  The result is Quassel upgrades being held up because it needs to install a new library.  It also results in multiple versions of libraries being installed and some unused libraries being left installed.
- I'm of the opinion that one should minimize external dependencies, only including them when no alternative exists and frankly, it seems like you're just adding the dependency for boost carelessly.

If any of this seems offensive or condescending then I do apologize because that is not my intent.

Sput42:
> > I don't see to get the introspection features required without the use of macros somehow (or continuing to rely on undocumented Qt internals
> 
> Are you thinking of something specific because introspection can be used with anything that's a QObject.
> Qt has the Meta-Object system: https://doc.qt.io/qt-5/metaobjects.html
> Really just calling QObject::metaObject() gets you a [QMetaObject](https://doc.qt.io/qt-5/qmetaobject.html)

Oh believe me, I know how QMetaObject works ;-) The current implementation of QObject syncing makes heavy use of that. Unfortunately, it also relies on some internal, undocumented features of Qt that may or may not continue to work in future versions of Qt. It also does most of the heavy lifting at runtime.

The new implementation would provide a mapping between method name and corresponding member function pointer at compile-time, avoiding the meddling in Qt internals, and providing compile-time checks. That mapping unfortunately is not available at compile time using the MOC or any other way that I know of, and so macros are unfortunately required.

That said, it's been two years since I last worked on this, I have forgotten the details and I will have to see if I follow that approach or do something differently. I might also look into [verdigris](https://github.com/woboq/verdigris), which might give us what we need.

> > If I may ask @GravisZro: why are you opposed to depending on Boost?
> 
> * It's asinine to require a massive library to enable a single convenience feature that is in the C++ standard.
> * It's a pain to have to install boost libraries and their corresponding dev packages just to compile Quassel.
> * boost seems to have a new package for each version (e.g. libboost-regex1.71.0 vs libboost-regex1.74.0).  The result is Quassel upgrades being held up because it needs to install a new library.  It also results in multiple versions of libraries being installed and some unused libraries being left installed.
> * I'm of the opinion that one should minimize external dependencies, only including them when no alternative exists and frankly, it seems like you're just adding the dependency for boost carelessly.

I see, and I actually agree in principle. However, my intent when I introduced the dependency was to use more Boost in the future (thus better justifying its use), because in different projects I have started to value it a lot as an extension of the STL. However, C++17 meanwhile removed a bunch of use cases, and C++20 will remove even more once we can rely on that. I haven't decided if keeping Boost will be justified in the long term or not, and will reconsider after we've switched to C++17 and see where my efforts to simplify and modernize the codebase will lead us.

That said, if you feel like it, you may update your PR with the parts that switch to std::optional and std::any, but leaving the find_package intact so I can still decide whether to remove the dependency independently. I'd include it then post-release. Would be a shame to let your efforts go to waste! Also the reason why I didn't reject your PR, but added a label instead. However, if you don't want to spend any more time, I'll just do it myself when I get to it.

> If any of this seems offensive or condescending then I do apologize because that is not my intent.

Absolutely not! I value these kinds of discussions!

GravisZro:
> Oh believe me, I know how QMetaObject works ;-) The current implementation of QObject syncing makes heavy use of that.

OK.... you're piqued my curiosity because only the _most_ sophisticated programs should need introspection and meta-objects.  Where is the core issue that requires the use the meta-object subsystem?  I'm getting this feeling that there is either a design issue or a misconception about the intended use of Qt that may be causing your troubles.

Sput42:
> > Oh believe me, I know how QMetaObject works ;-) The current implementation of QObject syncing makes heavy use of that.
> 
> OK.... you're piqued my curiosity because only the _most_ sophisticated programs should need introspection and meta-objects. Where is the core issue that requires the use the meta-object subsystem? I'm getting this feeling that there is either a design issue or a misconception about the intended use of Qt that may be causing your troubles.

The whole way we synchronize QObject instances over the network, and relay signal/slot connections across the network. We basically need to somehow represent the method called or property synced on one side to a message containing the method/property name as a string, and map that back to invoking the correct method on the other side. In the current release (0.13) we had to reimplement `qt_metacall()` (normally generated by the MOC) for a particular class to perform that conversion.  Since the upcoming release no longer supports Qt 4, we could get rid of this (completely undocumented and unsanctioned) hack, but we still do too many runtime lookups and AFAIR also still rely on some undocumented Qt functionality for marshalling arguments.

This never was "intended use of Qt", but Qt also didn't give us a sanctioned way to have a signal or property update transparently propagate over the network, and we didn't want to come up with a more explicit mechanism back in the day.
Obviously, there would have been different ways of implementing RPC, but our approach was really convenient when it came to using it in the codebase, and back in the day we actually impressed some of Qt's core developers with what we did to the point where there were discussions of including something like it upstream, but that never materialized. Today, we don't want to break protocol and don't want to rewrite a lot of our code, so...

If you're interested, feel free to have a look at `src/common/SignalProxy.cpp` and `src/common/SyncableObject.cpp`, plus the further improvements I have laying around in the PR I linked to above.

GravisZro:
1. https://doc.qt.io/qt-5.12/qtremoteobjects-gettingstarted.html
2. Why are you synchronizing QObject instances over a network in the first place?

GravisZro:
Wait, is this all simply to perform IPC between the server and client? You should be using generated code to do that. Not using generated code creates a vulnerability where the server could send arbitrary commands to the client or vice versa!

GravisZro:
Ok, I had time to look closer at the code and it doesn't connect QMetaObject::invokeMethod like I had thought.  The IPC mechanism you have rigged up is very clever and I absolutely do not recommend using it because it's quite powerful and makes doing something dangerous too easy.

That said, I now understand your desire to use boost introspection to replace the existing system.  Honestly though, I wish you would give up this entangled IPC syncronization system and instead build a stable IPC interface that is statically generated and can be inspected with ease.  I think doing so would radically reduce the complexity of your code, make it possible to create alternative clients, and reduce the number of resources required.  If nothing else, it would at the very least it would put my concerns about security to rest.

digitalcircuit:
Briefly chipping in - alternative clients do exist, including [Quasseldroid](https://quasseldroid.info/ ) which has a fairly robust and complete [implementation](https://git.kuschku.de/justJanne/QuasselDroid-ng/-/tree/master/lib/src/main/java/de/kuschku/libquassel/protocol ) by `justJanne`, [including documentation](https://quasseldroid.info/docs/protocol/overview/ ), but if I recall correctly it wasn't straightforward to figure out.

(I'm not yet familiar enough with the protocol and long-term projects to weigh in on which approach would be better.)

PR #555, opened 2020-12-15
Prepare for release of Quassel v0.14

Sput42:
- [x] Update the list of relevant Ubuntu releases
- [x] Update authors list
- [x] Update ChangeLog


ChangeLog
```diff
@@ -13,6 +13,57 @@ changes, the git history at <https://git.quassel-irc.org> is your friend.
...
+* Support many more features of IRCv3
+* Support the HAProxy protocol
+* Provide a daemon providing Prometheus metrics
+* Deprecate local highlight settings (prefer synced ones instead)
+* Improve configuration of SASL PLAIN and EXTERNAL
+* Add button for strike-through format
+* Allow choosing the account to connect to via command-line option
+* Make database schema migration more robust
+* Make backlog fetch more efficient, allow for fetch-on-demand
+* Make syncing and other aspects of the core/client protocol more efficienct
```
digitalcircuit, line 39: Minor typo: `efficienct` → `efficient`

ChangeLog
```diff
@@ -13,6 +13,57 @@ changes, the git history at <https://git.quassel-irc.org> is your friend.
...
+* Remove support for building against Qt4/KDE4
+* Require SSL support in Qt
+* Support many more features of IRCv3
+* Support the HAProxy protocol
+* Provide a daemon providing Prometheus metrics
+* Deprecate local highlight settings (prefer synced ones instead)
+* Improve configuration of SASL PLAIN and EXTERNAL
+* Add button for strike-through format
+* Allow choosing the account to connect to via command-line option
+* Make database schema migration more robust
```
digitalcircuit, line 37: This is referring to the SQLite to PostgreSQL migration, right?  Shouldn't be any need to mention this, just making sure I'm not overlooking anything :slightly_smiling_face:

digitalcircuit (commented):
One minor typo, otherwise looks good!

ChangeLog
Sput42: Fixed.

ChangeLog
Sput42: I was referring to the schema upgrade thing that you did.

digitalcircuit: Noted!  I first misread this as relating to [the database schema upgrade resumption changes already backported to 0.13.1](https://github.com/quassel/quassel/pull/474 ).  Instead, perhaps this is the [PostgreSQL text limits which fixes SQLite migrations](https://github.com/quassel/quassel/pull/478#issuecomment-683231671 ).

I think.  It's not a big deal regardless!  I may simply be missing something since it's a bit late.

Sput42: I was referring to https://github.com/quassel/quassel/commit/f91f3faa1432894e6d6ecaaf0a1f60a483dd0129 ;-)

Do you think the other things warrant an extra entry? Arguably "migrations" is generic enough to both cover upgrades and backend switch...

Sput42: Oh. Now I understand, this was already in 0.13.1, which I didn't realize as master branched off before that...

Sput42: Well, I guess we can just leave the entry as-is ;-)

digitalcircuit: Ah, whoops, so you did mean the past upgrade resumption fix.

I think the current wording is fine.  I was confused and mostly wanted to make sure there wasn't some other major database schema handling change I overlooked testing, since this mentioned in the 0.13.1 change log as well:

`* Make upgrading the database schema more robust in case of interruption`

With this clarification in mind, there is not in fact a major change I missed 🙂

For a higher level change log, what's here is good.  That the bug report on the tracker is closed is probably adequate - I don't imagine there's a lot of SQLite users with long user settings (what was previously limited) that were waiting for this to switch.

PR #556, opened 2021-01-03
Modernize and fix deployment on macOS

Sput42:
Replace our collection of ancient deployment scripts by using the official tooling for creating bundles and DMG image files. Clean up related CMake code and remove a variety of workarounds.

This also fixes the race condition between building, installing and deploying that regularly prevented some packages from being built.

Adapt the CI configuration accordingly.


digitalcircuit (approved):
Though I'm unable to test the macOS packages, the Windows builds seem to continue working just as before - core opens, client and monolithic connect and send messages, web previews work.  A review of the macOS changes here appear reasonable, and the output seems fine too (there's some warnings but it looks like a limitation of the CMake bundle utilities).

It's awesome to see the build scripts be replaced by better build scripts :slightly_smiling_face:

PR #557, opened 2021-01-14
Update upgrade_000_create_buffer_idx.sql

nvtkaszpir:
Create index if it does not exist already, some people have it applied prior doing upgrades.

### DIFF
--- a/src/core/SQL/SQLite/version/26/upgrade_000_create_buffer_idx.sql
+++ b/src/core/SQL/SQLite/version/26/upgrade_000_create_buffer_idx.sql
@@ -1,1 +1,1 @@
-CREATE INDEX backlog_buffer_msg_idx ON backlog (bufferid, messageid)
+CREATE INDEX IF NOT EXISTS backlog_buffer_msg_idx ON backlog (bufferid, messageid)


nvtkaszpir:
Background:
I'm having pretty ancient quassel setup up and runnign 24/7, so the storage was over 10GB over the years.
I added previosuly an index to sqlite (stop quassel core, make sqlite backup, run sqlite3:
```sql
CREATE INDEX backlog_buffer_msg_idx ON backlog (bufferid, messageid)
```
wait a bit, quit sqlite3, start quassel, enjoy better performance for years (and avoid upgrades at all costs).

Now how to fix the issue on upgrade?
Well, rever the change in sqlite, so that quassel upgrade works.

stop quassel core
make a backup of quassel-storage.sqlite
sqlite3 quassel-storage.sqlite
```sql
 DROP INDEX backlog_buffer_msg_idx;
 ```
 wait till it finishes, quit sqlite3
 start quassel-core service
 now it should be able to perform upgrade.

digitalcircuit (approved):
Thank you very much!

@justjanne had actually pointed out I should have made more SQLite upgrade queries idempotent, so if the change is already made then it won't error out.  The use of `IF NOT EXISTS` is one part of this.

SQLite appears to have supported `CREATE INDEX IF NOT EXISTS` *since 2006*, so we shouldn't have any issues adding this.

[SQLite changelog](https://sqlite.org/changes.html ):
> ### 2006-01-11 (3.3.0 alpha)
> 1. CHECK constraints
> 2. IF EXISTS and IF NOT EXISTS clauses on CREATE/DROP TABLE/INDEX.
> 3. DESC indices
> 4. […]

Though I haven't yet tested this myself, your changes are small and all looks correct :slightly_smiling_face:

nvtkaszpir:
Actually  I tested the query on the original copy of my sqlite file, no issues (it does nothing if index exists, and creates one if it is missing).

digitalcircuit:
@nvtkaszpir Pardon, I didn't mean to imply that these changes weren't tested at all.  I meant that to show I'm confident enough in your changes and your testing to approve this despite having not tested it personally.

nvtkaszpir:
No problem, my intention was merely to point out that this is no londer alpha feature from 2006 ;)

justjanne (approved):

nvtkaszpir:
if approved, then who has power and will to merge?

digitalcircuit:
@nvtkaszpir That's up to `freenode`→`#quassel`→`Sput` / [`@Sput42`](https://github.com/Sput42 ) - it might take a while due to busyness in real life, but I'd expect this to be merged before 0.14's release.

Sput42 (approved):

PR #560, opened 2021-02-19
core: Add support for Blowfish CBC encrypted messages prefixed with mcps

BaneHardy:
Previously only encrypted messages starting with `+OK *` were accepted for Blowfish CBC decryption.
However some implementations also use the prefix `mcps *` for CBC encrypted messages.


Sput42 (changes requested):
I think it would make sense to consolidate string handling here, and don't perform the same string comparison several times...

PR #561, opened 2021-02-23
[PR-561] Add support for weird tag names, per the spec.

justjanne:
## In Short

* Split tag names into at most 2 elements
* Split by the last /, not the first /

## Impact

| Criteria | Rank | Reason |
| - | - | - |
| Impact | ★☆☆  _1/3_ | Fixes handling of potential future extensions to IRCv3 |
| Risk | ★★☆  _2/3_ | Touches the main IRC parser, but is fully covered by tests |

## Rationale

IRCv3 currently does not specify (or recommend specifying) any tag names which would contain multiple slashes, but IRCv3 does recommend clients should gracefully handle any possible tag, ideally treating it as opaque string.

We’d like to avoid that, but also want to ensure even after a roundtrip through our core, tags we don’t support are still in identical condition.

### DIFF
--- a/src/common/ircdecoder.cpp
+++ b/src/common/ircdecoder.cpp
@@ -54,9 +54,11 @@ QString IrcDecoder::parseTagValue(const QString& value)
                 result.append(*it);
             }
             escaped = false;
-        } else if (it->unicode() == '\\') {
+        }
+        else if (it->unicode() == '\\') {
             escaped = true;
-        } else {
+        }
+        else {
             result.append(*it);
         }
     }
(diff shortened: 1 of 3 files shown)


digitalcircuit (approved):
I've read through the code and compiled this, connected to an IRCv3-compliant server (Oragono testing server), and didn't have any issues! :checkered_flag:

`account-tag` and `server-time` both continued to be parsed correctly, e.g.:
```
2021-02-23 22:14:02 [Debug] IRC net  1 << "@msgid=evtd7cagvyrdecrcpjd28bprae;time=2021-02-24T03:14:02.076Z;account=dcircuit_test2 :dcircuit_test2!~AdminUser@pw8quupeiazns.oragono PRIVMSG #chat :Test with example account!"
2021-02-23 22:14:02 [Debug] IRC net  1 << QHash(("(clientTag = 0, vendor = ,key = time", "2021-02-24T03:14:02.076Z")("(clientTag = 0, vendor = ,key = msgid", "evtd7cagvyrdecrcpjd28bprae")("(clientTag = 0, vendor = ,key = account", "dcircuit_test2")) "dcircuit_test2!~AdminUser@pw8quupeiazns.oragono" "PRIVMSG" ("#chat", "Test with example account!")
```

@Sput42 If feasible, it'd be nice to get this in before `0.14rc2` or the full `0.14` release.  I agree with the risk assessment given in the description.

It's not a major concern, but it helps ensure there won't be problems down the road if IRCv3 tag names add additional slashes.  Quassel currently does not make use of tags for much (just `server-time` and `account-tag`), so absolute worst-case it shouldn't introduce regressions compared to `0.13.1` (which didn't support any account tags).

Sput42 (approved):

PR #563, opened 2021-03-19
Set parent for new windows dialog

alkino:
This way wayland creates floating windows.


kadirselcuk:
It's worth using tags to add new features

kadirselcuk (approved):

Sput42 (approved):
This looks good, thx. However I wonder why we're not doing this for all the dialogs then? Should we add that to this or another PR, maybe?

alkino:
> This looks good, thx. However I wonder why we're not doing this for all the dialogs then? Should we add that to this or another PR, maybe?

I think another PR will be better.

PR #569, opened 2021-05-02
Fix parents for all dialog windows

alkino:
Without the parent correctly set, wayland cannot know it's a child of quassel and management is hazardeous.


alkino:
Seems because of my patch:

2021-05-02 12:54:12 [Warn ] setGrabPopup called with a parent, QtWaylandClient::QWaylandXdgSurface(0x55c7b09cac00) which does not match the current topmost grabbing popup, QtWaylandClient::QWaylandXdgSurface(0x55c7b0e463b0) According to the xdg-shell protocol, this is not allowed. The wayland QPA plugin is currently handling it by setting the parent to the topmost grabbing popup. Note, however, that this may cause positioning errors and popups closing unxpectedly because xdg-shell mandate that child popups close before parents

alkino:
It's not my patch. Can we merge this?

PR #571, opened 2021-05-19
Change default IRC network to Libera Chat

phuzion:
### Heads up! I suspect this will be a topic that gets discussed a LOT before a decision is made.

As some folks may be aware, there have been some recent issues that have come to light regarding the ownership and management of Freenode. As a result of this, the entire volunteer staff of Freenode has resigned.

Many projects are packing up and moving to Libera, which is a new IRC network founded by the previous staff of Freenode.

This PR makes the following changes:

* Removes Freenode as the "Default" network. This means that new installations of Quassel will no longer automatically connect to Freenode, unless the user specifically adds Freenode as a network.
* Removes #quassel from "DefaultChannels" under Freenode. This means that new installations of Quassel will no longer join the #quassel channel on Freenode.
* Adds Libera Chat as a new IRC network, and specifies it as the "Default" network, and adds #quassel under "DefaultChannels". This means that new installations of Quassel will now automatically connect to `irc.libera.chat:+6697` and automatically join #quassel.
* Changes the link in Help>About Quassel to `irc://irc.libera.chat/quassel` and links to Libera Chat's website instead of Freenode's. (Note that `ircs://` was specifically not chosen because `irc://` is a more widely recognized URI scheme.)
* Updates the README document with the new network name and address. The webchat link is temporarily removed, as the new network does not yet have one up and running. I will add it with another PR when the webchat is online.

For more information about this situation, please read this: https://gist.github.com/joepie91/df80d8d36cd9d1bde46ba018af497409

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,9 +47,8 @@ documentation.
 
 IRC is the preferred means of getting in touch with the developers.
 
-The Quassel IRC Team can be contacted on **`Freenode/#quassel`**
-(or **`#quassel.de`**).  If you have trouble getting Quassel to connect,
-you can use [Freenode's webchat][help-freenode].
+The Quassel IRC Team can be contacted on **`Libera Chat/#quassel`**
+(or **`#quassel.de`**). 
 
 We always welcome new users in our channels!
 
@@ -65,7 +64,6 @@ Thanks for reading,
 
 [web-home]: https://quassel-irc.org
 [dev-email]: mailto:devel@quassel-irc.org
-[help-freenode]: https://webchat.freenode.net?channels=%23quassel
 [repo-changelog]: ChangeLog
 [ci-badge]: https://github.com/quassel/quassel/workflows/Quassel%20CI/badge.svg?branch=master
 [ci-status-page]: https://github.com/quassel/quassel/actions?query=branch%3Amaster
--- a/data/networks.ini
+++ b/data/networks.ini
@@ -19,8 +19,6 @@ Servers=irc.efnet.info:6667,irc.efnet.org:6667,irc.inter.net.il:6667,irc.igs.ca:
 Servers=irc.enterthegame.com:6667
 
 [Freenode]
-Default=Yes
-DefaultChannels=#quassel
 Servers=chat.freenode.net:+6697,chat.freenode.net:6667
 
 [GalaxyNet]
@@ -34,6 +32,11 @@ Servers=irc.gamesnet.net:6667
 
 [IRCnet]
 Servers=irc.ircnet.com:6667,open.ircnet.net:6666,irc.us.ircnet.net:6665,ircnet.netvision.net.il:6666,irc.tokyo.wide.ad.jp:6666,irc.freebsd.org.tw:6666,linz.irc.at:6666,vienna.irc.at:6666,ircnet.realroot.be:6666,irc.datacomm.ch:6664,irc.felk.cvut.cz:6667,random.ircd.de:6666,irc.dotsrc.org:6666,irc.ircnet.ee:6666,irc.cs.hut.fi:6667,ircnet.club-internet.fr:6666,atw.irc.hu:6667,irc.simnet.is:6660,irc.eutelia.it:6664,irc.eutelia.it:7000,irc.tin.it:6665,irc.apollo.lv:6666,irc.apollo.lv:7000,irc.uunet.nl:6660,irc.xs4all.nl:6660,irc.snt.utwente.nl:6660,irc.ifi.uio.no:6667,irc.pvv.ntnu.no:6667,lublin.irc.pl:6666,warszawa.irc.pl:6666,irc.ludd.luth.se:6661,irc.swipnet.se:6660,irc.swipnet.se:8000,irc.arnes.si:6666,irc.link.si:6666,irc.nextra.sk:6666,ircnet.demon.co.uk:6665,ircnet.eversible.com:6665,ircnet.choopa.net:6665
+
+[Libera Chat]
+Default=Yes
+DefaultChannels=#quassel
+Servers=irc.libera.chat:+6697,irc.libera.chat:6667
 
 [OFTC]
 Servers=irc.oftc.net:+6697,irc.oftc.net:6667
--- a/src/qtui/aboutdlg.cpp
+++ b/src/qtui/aboutdlg.cpp
@@ -65,7 +65,7 @@ QString AboutDlg::about() const
     QString res{tr("<b>A modern, distributed IRC Client</b><br><br>"
                    "&copy;%1 by the Quassel Project<br>"
                    "<a href=\"https://quassel-irc.org\">https://quassel-irc.org</a><br>"
-                   "<a href=\"irc://irc.freenode.net/quassel\">#quassel</a> on <a href=\"https://www.freenode.net\">Freenode</a><br><br>"
+                   "<a href=\"irc://irc.libera.chat/quassel\">#quassel</a> on <a href=\"https://libera.chat\">Libera Chat</a><br><br>"
                    "Quassel IRC is dual-licensed under <a href=\"https://www.gnu.org/licenses/gpl-2.0.txt\">GPLv2</a> and "
                    "<a href=\"https://www.gnu.org/licenses/gpl-3.0.txt\">GPLv3</a>.<br>"
                    "<a href=\"https://api.kde.org/frameworks/breeze-icons/html\">Breeze icon theme</a> &copy; Uri Herrera and others, "


guppy:
Seems sensible.

digitalcircuit:
On a technical level, this looks reasonable!  Socially, well, the freenode situation sucks (to put it lightly), and I'm in favor of moving. :+1:

Mild positive: Quassel's channel `/list` dialog with the quick filtering options makes it easy to check which channels are available already.

As noted, **this will need discussed** and Sput will have to officially register the Quassel IRC channels **before this can be merged.**

AlD:
As someone who hasn't followed this issue at all and isn't emotionally attached to Freenode or its staffers, I have to say that https://gist.github.com/joepie91/df80d8d36cd9d1bde46ba018af497409 is very light on provable facts and all the stronger on unproven allegations.

I haven't read all the other pamphlets, but e.g. https://fuchsnet.ch/freenode-resign-letter.txt is quite nebulous as well.

Can we get something with more facts and less emotions to base a decision on?

Ketchup901:
Pointless outrage culture

crabbedhaloablution:
I support the move. Recent proposed changes to Freenode's TOS has convinced me that Andrew does not aim to engage with Freenode's users, choosing instead to allow more leeway for hate speech and punishing channels that leave Freenode.

iamgreaser:
Providing a point of reference, FreeNode staff performed a mass-takeover of a bunch of abandoned channels a bit over an hour ago in order to fool people into thinking they hadn't moved.

After moving you're definitely going to want to make it very clear on the website that the FreeNode channel is NOT the official channel, because they WILL pull the scam on what at the time of writing is probably still the official channel.

Heck, at this rate, it's probably worth removing FreeNode from the pre-installed networks list altogether, but that'd be the scope of a different PR.

phuzion:
Source for @iamgreaser's claim about Freenode forcibly taking over channels: https://news.ycombinator.com/item?id=27285774

deviant:
I can confirm that they took over a large number of channels which mentioned Libera in the topic (I'm in hundreds on freenode), and they also did this to @justjanne's `#quasseldroid`. This should get merged ASAP.

Sput42:
Meanwhile, we have established a presence at Libera.Chat. Feel free to join the corresponding channels there!

I was hoping for having a bit more of discussion between the regulars before we decide to move officially, but in the light of the current situation that seems to be getting worse at exponential speeds, bigger communities already moving, and the fact that so far, no one at all seems to be advocating for Quassel IRC staying on Freenode, it looks to me that the decision is basically clear at this point? In which case, we might consider merging this before the next release even...

Any objections?

kipe:
I'm not in any way a regular in the sense of active participation, just been hanging around in #quassel and #quasseldroid for a few years. Not actually in Libera for now, as my shell provider doesn't have Y-lines yet.

In the light of recent events I'm all for this change. The sooner the better, so new users (which IRC might possibly gain, thanks to the publicity) wouldn't wind up in Freenode, especially as it seems like the official channel is (and should be) moving at some point anyways.

phuzion:
@Sput42 I was also expecting this to be a many-weeks-long process where we discuss the merits and potential downsides of migrating to a new network.

As I mentioned in IRC, my hope with this PR was to start a discussion. I was hoping that we would discuss the pros and cons of moving, in the meantime things with Freenode would get ironed out in some way, and this PR would be rejected after we all decided that it was not necessary. Or something along those lines, I don't know exactly.

However, in light of last night's late-night forceful channel takeover, I cannot in good faith recommend that anyone stay on Freenode any longer.

I do believe that it would be best to expedite the merging of this PR, and if at all possible expedite the release of 0.14.0, so as to minimize the number of people connecting to an IRC network under hostile new management.

akhuettel:
Libera Chat is the "continuity solution". The network is less defined by name, hardware, and owner; more by the community of people running it.

AlD:
> Libera Chat is the "continuity solution". The network is less defined by name, hardware, and owner; more by the community of people running it.

Sounds nice, but so far it's still ultimately defined by ownership of the libera.chat domain.
IRCNet is an example of an IRC network to which this actually applies.

No objection to merging this PR though.

AlD:
Actually, having read https://gist.github.com/prawnsalad/4ca20da6c2295ddb06c1646791c61953 now, I wonder whether we should consider OFTC instead.

mamarley:
> Actually, having read https://gist.github.com/prawnsalad/4ca20da6c2295ddb06c1646791c61953 now, I wonder whether we should consider OFTC instead.

My main issue with OFTC (from a technical point of view) is that they (still) don't support SASL, which can make for annoying race conditions when rejoining channels after connecting.

justjanne:
> Actually, having read https://gist.github.com/prawnsalad/4ca20da6c2295ddb06c1646791c61953 now, I wonder whether we should consider OFTC instead.

Considering Libera.Chat is run by the same team that has previously run freenode, with whom we had no issues, no trouble, nothing over all these years, they've proven themselves to be trustworthy. We know them, and they know us.

That's in addition to the SASL issue.

(Also, short disclaimer, prawnsalad was, until that resignation, still an employee of Andrew Lee)

AlD:
It looks like tomaw is the head of OFTC as well, so that wouldn't be helping in that regard either.

> Considering Libera.Chat is run by the same team that has previously run freenode, with whom we had no issues, no trouble, nothing over all these years, they've proven themselves to be trustworthy. We know them, and they know us.

Not sure I agree 100% there. I don't see good alternatives either though.

eli-schwartz:
As far as I can tell, tomaw is not involved in libera.chat as of now in any way shape or form?

If I had to hazard a wild guess, then my opinion is he's blameless in the entire affair but he's under lawyer's orders to not do anything that might make Lee's lawyers better able to attack him. Like make public statements or be involved in the new network.

deviant:
That's my assumption, yes. It doesn't seem like he's the cause of any of this, just in the wrong place at the wrong time.

derlaft:
>Adds Libera Chat as a new IRC network, and specifies it as the "Default" network, and adds #quassel under "DefaultChannels". This means that new installations of Quassel will now automatically connect to irc.libera.chat:+6697 and automatically join #quassel.

I would highly recommend thinking this over and not repeating the same mistake again. The strength of IRC is federation and decentralization, and leading people to some huge server will IMO eventually lead to the same problem. So it would be awesome to find a smaller server for #quassel and avoid points of centralization from now on.

eli-schwartz:
I think of this kind of thing happens only once every 20 years it's not exactly the end of the world. But libera.chat staff explicitly set up a nonprofit and established bylaws with the intention of never allowing something like this to happen again.

Regardless, I don't see how "use a network no one else uses just to be different" is going to make quassel safer than following the continuity option. It's still centralized, just on a smaller, less dedicatedly maintained network.

(What on earth do you mean by federation? There is no federation here, only official network servers. Did you think we were talking about Matrix?)

derlaft:
>There is no federation here, only official network servers. 

A bridge can create a federation even between IRC channels :}

> I think of this kind of thing happens only once every 20 years it's not exactly the end of the world

Yep, there's should be no immediate danger in going with libera.chat, it's only about making a decision which sucks a bit less and not putting your eggs into the same bucket everyone else has put theirs.

eli-schwartz:
> A bridge can create a federation even between IRC channels :}

This is totally unrelated to choosing a network as the home of #quassel, and it's also not "the strength of IRC" if you use out of band federation platforms which aren't part of the IRC definition or any IRC spec.

Maybe it's "the strength of matrix".

> making a decision which sucks a bit less by
> 
> - not putting your eggs into the same bucket everyone else has put theirs,

Instead everyone else makes the decision to suck less by not putting their eggs into the same bucket with a hole in it that you've put yours? I don't understand the argument.

The project should choose a network based on what they believe will be a robust and welcoming network, not based on "most people don't use this odd network, therefore if most people get targeted by a network takedown, we won't be".

Sput42:
Personally I think that putting our eggs into the same bucket as most major FOSS projects we care about (including Qt, KDE, Gentoo, Ubuntu...) isn't the worst choice to make, and the fact that Libera.Chat is run by pretty much the same people that had made Freenode a good home for almost 20 years helps a lot, too.

The fact that Libera.Chat sprang to life within such a short amount of time after the hostile takeover of Freenode also proves that the people behind it are pretty capable to keep things going in a way that is good for the community.

All that is not to say that there aren't other really nice networks out there, and obviously everyone is welcome to join them too, but I believe we should stick with what a large majority of the FOSS community prefers.

winks:
> > Adds Libera Chat as a new IRC network, and specifies it as the "Default" network, and adds #quassel under "DefaultChannels". This means that new installations of Quassel will now automatically connect to irc.libera.chat:+6697 and automatically join #quassel.
> 
> I would highly recommend thinking this over and not to repeat the same mistake again. The strength of IRC is federation and decentralization, and leading people to some huge server will IMO eventually lead to the same problem. So it would be awesome to find a smaller server for #quassel and avoid points of centralization from now on.

Chiming in as a low-traffic regular reader on #quassel - I think there are 2 parts.

a) Change the official channel (where some of the development happens)
b) make the released client auto-connect.

Let me explain. Yes, Quassel is an IRC client, but I am not sure the users must be guided into that direction, it may be a bit of a higher bar to "find" the channel, but I don't think any default is needed. Neither freenode back then, nor libera.chat now if the channel decides to move. People can read websites and there's nothing against pointing them to the channel for questions, but I see this as something different than providing it as a default choice for a new connection. But it's the only choice that makes sense, because anything else would be random. On the other hand I've always found a default channel weird. I don't install an irc client just to talk to the irc client's developers, but that might be me being werid.

Regarding the actual move: 100% pro libera.chat

AlD:
Please move this discussion elsewhere. Quassel has a default network/channel right now, and this PR is just about changing its target.

winks:
> 
> 
> Please move this discussion elsewhere. Quassel has a default network/channel right now, and this PR is just about changing its target.

I fail to see how "how about we don't use a default channel to join" is not a valid alternative approach in exactly this PR? Which I only added as an aside, btw and I think what @phuzion answered is a very good reasoning to not do that and I think most people will agree here. I guess I mostly agree, but that doesn't make showing up an alternative wrong.

eli-schwartz:
It's unrelated to the goal of migrating the official quassel support channel from one network to another, which is what 2 of the 3 changes here do.

And the 3rd change causes quassel to autojoin the official support channel, but now on a different network. Continuing to autojoin is the continuity option of "default to join the official support channel".

Deciding to not autojoin anything at all is out of scope for pointing the status quo to the right location.  It's not an "alternative", it's an unrelated change that would then obsolete 1/3 of this PR. Asking to discuss it somewhere else (in a new issue) doesn't seem unreasonable.

Sput42:
> Chiming in as a low-traffic regular reader on #quassel - I think there are 2 parts.
> 
> a) Change the official channel (where some of the development happens)
> b) make the released client auto-connect.
> 
> Let me explain. Yes, Quassel is an IRC client, but I am not sure the users must be guided into that direction, it may be a bit of a higher bar to "find" the channel, but I don't think any default is needed. Neither freenode back then, nor libera.chat now if the channel decides to move. People can read websites and there's nothing against pointing them to the channel for questions, but I see this as something different than providing it as a default choice for a new connection. But it's the only choice that makes sense, because anything else would be random. On the other hand I've always found a default channel weird. I don't install an irc client just to talk to the irc client's developers, but that might be me being werid.

Adding a default channel as well as a way to easily change that default (by patching a simple text file, rather than code) was a request made by Ubuntu prior to making Quassel their default IRC client years ago. The idea being that distros would route their users not to #quassel, but to the distro's own support channel, e.g. #ubuntu. Even though Ubuntu has long since chosen to select a different default IRC client again, I still think the idea is a good one, and there are probably still distros adapting the default channel to their needs.

> Regarding the actual move: 100% pro libera.chat

Good ;-)

PR #572, opened 2021-05-20
Join current channel if no name is given

alkino:
If user type `/join` on a channel without name it will take the name of the current channel


deviant:
I do this somewhat frequently, expecting it to work. Would love to see this merged!

PR #575, opened 2021-05-30
3rdparty: Update GoogleTest to a3460d1

Sput42:
Google recommends to live at head, and thus has stopped creating
regular releases a while ago. Update the bundled GoogleTest
to current master, which happens to be a3460d1 at this time.


a17r:
Meanwhile even that wouldn't be enough to fix a CMake error for dropping <3.5 support: https://bugs.gentoo.org/957565

johu:
I have updated the library to 1.17.0 in #631 and upstream is preparing a 1.18.0 see https://github.com/google/googletest/pull/5051 for reference. I think the library should be unbundled at all.

PR #579, opened 2021-06-14
client: Swap order of "Configure" and "Connect/Disconnect" in context menu

phuzion:
This commit changes the order of the "Configure" and "Connect/Disconnect" items in the context menu, found when right-clicking on a network.

My rationale behind this is that when right-clicking on a network, it is extraordinarily more likely that the user's intent is to either connect to or disconnect from the network, and not to edit that network's settings.

I am not the only person to have this complaint, see [this bug](https://bugs.quassel-irc.org/issues/1494) filed in September 2018.

### DIFF
--- a/src/uisupport/contextmenuactionprovider.cpp
+++ b/src/uisupport/contextmenuactionprovider.cpp
@@ -277,10 +277,11 @@ void ContextMenuActionProvider::addNetworkItemActions(QMenu* menu, const QModelI
     if (!network)
         return;
 
-    addAction(ShowNetworkConfig, menu, index);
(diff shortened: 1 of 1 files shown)


genius3000:
Just so it doesn't go unsaid, I'll mention my opposite point of view.
As mentioned in my PR #284 the top use case for this button and it's usefulness as the first item is development and/or test networks. I completely understand that while I fit that use case and find myself using `Configure` far more often than anything else in this menu, I and this use case are not the majority.
Ultimately, I'm open to re-learning the order but would prefer it remains as is.

PR #581, opened 2021-06-16
core: Error out when --require-ssl is specified, but no cert can be loaded

phuzion:
As mentioned in [#1728](https://bugs.quassel-irc.org/issues/1728), cores launched with the `--require-ssl` flag, but no SSL/TLS certificate available (because the file does not exist, permissions are set incorrectly, the file does not contain a certificate, etc) will effectively be put into a plaintext-only mode, with little more than a small warning on the console to the user. Clients connecting to the core will be given the standard "Unencrypted Connection" warning.

This patch fixes that issue, requiring that cores launched with `--require-ssl` will successfully load a certificate before starting. If the certificate fails to load for any reason during startup, an exception is thrown and the core quits.

Thanks to @relrod for the help on this one.

### DIFF
--- a/src/core/sslserver.cpp
+++ b/src/core/sslserver.cpp
@@ -53,7 +53,8 @@ SslServer::SslServer(QObject* parent)
         // do not proceed, throw an exception and quit. This prevents the core from falling
         // back to a plaintext-only core when they should be expecting SSL/TLS only.
         if (Quassel::isOptionSet("require-ssl")) {
-            throw ExitException{EXIT_FAILURE, tr("--require-ssl is set, but no SSL certificate is available. Exiting.")};
+            throw ExitException{EXIT_FAILURE, tr("--require-ssl is set, but no SSL certificate is available. Exiting.\n"
+                                                 "Please see https://quassel-irc.org/faq/cert to learn how to enable SSL support.")};
         }
         if (!sslWarningShown) {
             qWarning() << "SslServer: Unable to set certificate file\n"
--- a/src/core/sslserver.cpp
+++ b/src/core/sslserver.cpp
@@ -49,6 +49,12 @@ SslServer::SslServer(QObject* parent)
 
     // Initialize the certificates for first-time usage
     if (!loadCerts()) {
+        // If the core is unable to load a certificate, and "--require-ssl" is specified,
+        // do not proceed, throw an exception and quit. This prevents the core from falling
+        // back to a plaintext-only core when they should be expecting SSL/TLS only.
+        if (Quassel::isOptionSet("require-ssl")) {
+            throw ExitException{EXIT_FAILURE, tr("--require-ssl is set, but no SSL certificate is available. Exiting.")};
+        }
         if (!sslWarningShown) {
             qWarning() << "SslServer: Unable to set certificate file\n"
                        << "          Quassel Core will still work, but cannot provide SSL for client connections.\n"


relrod:
This security issue has been assigned `CVE-2021-34825`.

justjanne:
@phuzion has verified that, when reloading certs, the old cert continues being used if the new certs can not be found :)

justjanne (approved):

phuzion:
As @justjanne mentioned, I did some verification that reloading certs can't put a core into the plaintext-only mode.

To test this, I did the following:

* Set up and launched a core, with a properly configured `quasselCert.pem` file in the configdir.
* Connected to the core to ensure that it works properly
* Disconnected from the core
* Removed the `quasselCert.pem` file from the configdir
* Sent `SIGHUP` (with `pkill -1 quasselcore`) to the core to trigger a configuration reload
* Reconnected to the core

After reconnecting, I was still presented with the previously configured SSL certificate.

src/core/sslserver.cpp
```diff
@@ -49,6 +49,12 @@ SslServer::SslServer(QObject* parent)
 
     // Initialize the certificates for first-time usage
     if (!loadCerts()) {
+        // If the core is unable to load a certificate, and "--require-ssl" is specified,
+        // do not proceed, throw an exception and quit. This prevents the core from falling
+        // back to a plaintext-only core when they should be expecting SSL/TLS only.
+        if (Quassel::isOptionSet("require-ssl")) {
+            throw ExitException{EXIT_FAILURE, tr("--require-ssl is set, but no SSL certificate is available. Exiting.")};
```
digitalcircuit, line 56:
```suggestion
            throw ExitException{EXIT_FAILURE, tr("--require-ssl is set, but no SSL certificate is available. Exiting.\n"
                                                 "Please see https://quassel-irc.org/faq/cert to learn how to enable SSL support.")};
```

Just an example, line-wrapping can be adjusted.

## Results
### Before
```
./quasselcore --listen="127.0.0.1" --configdir="/tmp/qtest/" --loglevel=Debug --require-ssl
2021-06-17 20:58:00 [Warn ] SslServer: Certificate file /tmp/qtest/quasselCert.pem does not exist
2021-06-17 20:58:00 [Error] --require-ssl is set, but no SSL certificate is available. Exiting.
```

### After
```
./quasselcore --listen="127.0.0.1" --configdir="/tmp/qtest/" --loglevel=Debug --require-ssl
2021-06-17 21:10:37 [Warn ] SslServer: Certificate file /tmp/qtest/quasselCert.pem does not exist
2021-06-17 21:10:37 [Error] --require-ssl is set, but no SSL certificate is available. Exiting.
Please see https://quassel-irc.org/faq/cert to learn how to enable SSL support.
```

digitalcircuit (commented):
This looks sensible!  Just some minor questions/nitpicks, some you might not be able to address:

### ~~Unlike other SSL error messages, no pointer to website~~

~~Now, one could make the argument you'd remove the `--require-ssl` flag to see them, however some distributions include `--require-ssl` by default.  I feel we should include the website link if feasible to maintain parity with the other SSL/TLS warning messages.~~

~~See the inline comment with suggestion.~~

**This has been fixed!**

### Reason for exit can be hidden

The exit exception message doesn't seem to be displayed if the `--logfile` option is specified - it's only stored in the log file.  However, this applies to all exit exceptions, e.g. `--config-from-environment`.  This is not a new bug, but is worthwhile noting.

@Sput42 , is it intended for the logging of exit exceptions that result in Quassel exiting to be consumed by a `--logfile`?  If I'm understanding correctly, on Debian and Ubuntu even `systemctl status quasselcore.service` won't show this explanatory message unless someone turns on logging to `--syslog`.

Example:
```
user@host:~$ /path/to/quasselcore --require-ssl --listen="127.0.0.1" --configdir="/tmp/qtest/" --logfile="/tmp/qtest/quasselcore.log" --loglevel=Debug
user@host:~$ 
```

No messages shown.  Also applies to `--config-from-environment`, etc.

### Should go in release notes - potentially breaking change

It's worthwhile noting this *might* break an existing setup as it turns a prior warning situation into an error.  I'm in favor of this - if someone (**which includes package maintainers** - [Debian defaults to `--require-ssl`](https://salsa.debian.org/qt-kde-team/extras/quassel/-/blob/master/debian/quassel-core.quasselcore.service ) alongside a script to auto-generate certs) sets `--require-ssl`, I feel Quassel *should* error out.

As a result, though, this should likely go in the release notes/ChangeLog when Sput updates them, perhaps with emphasis.  If someone deleted the cert and didn't care about encryption, they'll have to remove the flag or put a new cert in.

### New string for translators

Minor - this also introduces a new translatable string for the error message.

PR #587, opened 2021-06-27
Core/ircparser: Only trim whitespace from the end of the last parameter.

genius3000:
> QString::trimmed() removes both leading and trailing whitespace, which
> is not the intention here. As QString does not have a ltrim/rtrim
> function, we need to implement the equivalent here to only trim the
> trailing whitespace.

> As QString::trimmed() removed any whitespace for which QChar::isSpace()
> returns true, I have switched the trailing whitespace check to use the
> same function.

I know this is kind of ugly, with a trim function implemented directly
here. I did not see anywhere else that would benefit from a new Quassel
function of rtrim. Totally open to other ideas though.

Set this as draft so I can improve this PR body later when not on mobile.

### DIFF
--- a/src/core/ircparser.cpp
+++ b/src/core/ircparser.cpp
@@ -468,8 +468,17 @@ void IrcParser::processNetworkIncoming(NetworkDataEvent* e)
 
         // We want to trim the last param just in case, except for PRIVMSG and NOTICE
         // ... but those happen to be the only ones not using defaultHandling anyway
-        if (!decParams.isEmpty() && decParams.last().endsWith(' '))
-            decParams.append(decParams.takeLast().trimmed());
+        if (!decParams.isEmpty()) {
+            QString& lastParam = decParams.last();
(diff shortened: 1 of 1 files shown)


deviant:
Oh, I think this will fix a bug I noticed where Quassel would for some reason strip the indentation from a few lines of the MOTD of a server I'm on. I understand why, now: these lines had trailing whitespace, which triggered the stripping logic.

justjanne:
Why is it even trimmed at all? Ideally we'd keep the content completely intact.

genius3000:
@justjanne from the commits, it was initially due to trailing spaces in a NAMREPLY causing empty nicks to be created.
Commits https://github.com/quassel/quassel/commit/9fe8b8324d6b2501220beba4cefe1902ac6bb153 and https://github.com/quassel/quassel/commit/9e637b2c82e4c6222e33f07ce9944794a95b9b1a.
Those are from a number of years ago and this trimming is just a safety from edge cases (imo). Is this something we can build a test against (I'm completely unfamiliar with the/any test suite) to be fairly certain we don't introduce any regressions by removing this trimming?

PR #588, opened 2021-07-14
[PR-586] Allow setting outgoing IP per user

justjanne:
## In Short

* Adds a database column to the quasseluser table for per-user outgoing IPs
* Binds to that IP when creating new outgoing sockets

## Impact

| Criteria | Rank | Reason |
| - | - | - |
| Impact | ★☆☆  _1/3_ | Allows separating outgoing users by IP, useful as IPv6 identd alternative |
| Risk | ★★★  _3/3_ | Changes network socket creation, might cause IRC connectivity to fail |

## Rationale

Currently, Quassel supports identd for per-user ident to allow owners of larger quassel cores to avoid a ban of one user affecting the others. Many networks have stopped supporting this, though, instead opting to move to IPv6 based alternatives.

With this change, the operator of a Quassel Core can set an IPv6 for each individual user, which will be used for outgoing IRC connections, allowing bans against one user to avoid affecting other users.

### DIFF
--- a/src/core/SQL/PostgreSQL/select_outgoing_ip.sql
+++ b/src/core/SQL/PostgreSQL/select_outgoing_ip.sql
@@ -1,0 +1,3 @@
+SELECT outgoing_ip
+FROM quasseluser
+WHERE userid = :userid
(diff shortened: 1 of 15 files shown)


justjanne:
Due to the risk involved with this change, I’d like to subject it to significant testing before we merge it and roll it out.

justjanne:
To use this functionality, you’ll first have to assign the corresponding IPs to an interface.

For example, you can assign a whole block with

```
ip route add local 2a01:4f8:241:4712::/64 dev enp40s0
```

phuzion:
When merged, this may at least partially resolve [bug 125](https://bugs.quassel-irc.org/issues/125).

Perhaps a way to make this "per network" as requested in the bug would be to bind the IP address to an identity, not to a user? Not sure if that adds any complexity to the logic of this PR, but I figured I'd throw out the idea.

PR #595, opened 2022-01-07
cmake: avoid de-duplication of user's CXXFLAGS

floppym:
Use the "SHELL:" prefix behavior introduced in cmake 3.12 to avoid the
option de-duplication behavior of add_compile_options().

Bug: https://bugs.gentoo.org/830738


floppym:
Looks like cmake is too old on Ubuntu bionic (18.04). Do you still want to support that?

floppym:
I pushed an update with a CMAKE_VERSION check instead.

Sput42 (approved):

Sput42:
> Looks like cmake is too old on Ubuntu bionic (18.04). Do you still want to support that?

We'll be updating the minimum requirements for master branch soon, and probably take Ubuntu 20.04 as the baseline. For now, I'll take the patch with the version check. Thanks for the contribution!

PR #601, opened 2022-04-02
Add formatting for RPL_HELPSTART/RPL_HELPTXT/RPL_ENDOFHELP

progval:
For example with the reply to `/help mode` on InspIRCd:

Before:

screenshot-2022-04-02_13-59-15

After:

screenshot-2022-04-02_13-59-01


genius3000:
Totally untested and I might be missing a reason but I would think `Message::Server` would be more suited for these instead of `Message::Error`. By default I think both direct to the status buffer, but each can be redirected in the client settings.

progval:
Oh yes, of course, my mistake. I also forgot to implement ERR_HELPNOTFOUND and added it just now.

progval:
ping?

PR #609, opened 2022-10-18
fix client execwrapper not using full path to script

maxaudron:
The clients execwrapper was using the scriptName instead of the full path fileName to execute scripts leading to scripts only being found and executed if the script directory was also in $PATH or a executable with the same name as the script was in $PATH.

This could also lead to confusion as it executes whatever is in $PATH instead of the actual script in quassels directories.


justjanne (approved):
LGTM

PR #612, opened 2023-02-01
fix for SEGFAULT

return-misaPuding:
_I apologise if I don´t have all the details needed, but i didn´t make these changes - they were made by fsk99 on #quassel (https://pastebin.com/4mDNZiWi) and i´m just pasting them because they don´t have a github account_

- Fixes SEGFAULT
- Tested by the author and it didn´t break anything on first glance
- "fsk99> looks liek the bug triggered an infinite recursion and stack exhaustion by that"

### DIFF
--- a/src/qtui/topicwidget.cpp
+++ b/src/qtui/topicwidget.cpp
@@ -261,13 +261,19 @@ bool TopicWidget::eventFilter(QObject* obj, QEvent* event)
 QString TopicWidget::sanitizeTopic(const QString& topic)
 {
     // Normally, you don't have new lines in topic messages
-    // But the use of "plain text" functionality from Qt replaces
+    // But the use of "plain text" functionnality from Qt replaces
     // some unicode characters with a new line, which then triggers
     // a stack overflow later
+
     QString result(topic);
-    result.replace(QChar::CarriageReturn, " ");
-    result.replace(QChar::ParagraphSeparator, " ");
-    result.replace(QChar::LineSeparator, " ");
+    for(int i = 0; i< result.length(); i++)
+    {
+        QChar a = result.at(i);
+        if (a.isPrint()) 
+            continue;
+        else
+        result[i] = QChar('?');
+    }
 
     return result;
 }
--- a/src/qtui/topicwidget.cpp
+++ b/src/qtui/topicwidget.cpp
@@ -270,9 +270,9 @@ QString TopicWidget::sanitizeTopic(const QString& topic)
     {
         QChar a = result.at(i);
             if (a.isPrint()) 
-            continue;
+                continue;
             else
-            result[i] = QChar('?');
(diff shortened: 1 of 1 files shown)


src/qtui/topicwidget.cpp
```diff
@@ -261,13 +261,19 @@ bool TopicWidget::eventFilter(QObject* obj, QEvent* event)
 QString TopicWidget::sanitizeTopic(const QString& topic)
 {
     // Normally, you don't have new lines in topic messages
-    // But the use of "plain text" functionality from Qt replaces
+    // But the use of "plain text" functionnality from Qt replaces
```
reedy, line 264: Definitely introduces a typo

src/qtui/topicwidget.cpp
```diff
@@ -261,13 +261,19 @@ bool TopicWidget::eventFilter(QObject* obj, QEvent* event)
...
-    result.replace(QChar::CarriageReturn, " ");
-    result.replace(QChar::ParagraphSeparator, " ");
-    result.replace(QChar::LineSeparator, " ");
+    for(int i = 0; i< result.length(); i++)
+    {
+        QChar a = result.at(i);
+        if (a.isPrint()) 
+            continue;
+        else
+        result[i] = QChar('?');
```
reedy, line 275: And this line should be indented

return-misaPuding: If i didn't mess up, 2nd commit fixed both

src/qtui/topicwidget.cpp
```diff
@@ -264,10 +264,16 @@ QString TopicWidget::sanitizeTopic(const QString& topic)
...
-    result.replace(QChar::CarriageReturn, " ");
-    result.replace(QChar::ParagraphSeparator, " ");
-    result.replace(QChar::LineSeparator, " ");
+    for(int i = 0; i< result.length(); i++)
+    {
+        QChar a = result.at(i);
+            if (a.isPrint()) 
+            continue;
+            else
+            result[i] = QChar('?');
```
reedy, line 275:
```
        QChar a = result.at(i);
            if (a.isPrint()) 
            continue;
            else
            result[i] = QChar('?');
```

should be

```
        QChar a = result.at(i);
        if (a.isPrint()) 
            continue;
        else
            result[i] = QChar('?');
```

Though, it can also be (the else adds nothing when the conditional always returns):

```
        QChar a = result.at(i);
        if (a.isPrint()) 
            continue;

        result[i] = QChar('?');
```

Braces seem to be optional for 1 line conditiionals

Sput42:
Superseded by #620 which contains that fix. Thanks for your contribution!

return-misaPuding:
what in the long review... I guess the author from IRC can sleep peacefully knowing that quassel will no longer crash :P

PR #613, opened 2023-02-07
Allow basic setup to be disabled via environment variable

luckcolors:
Hello.

I'm wanted to propose a feature to quassel.

Sometimes it happens that quassel can lose connection to the database instance.
And by default if it restarts, it can get into the basic setup, even if that is something that won't be necessary as the database can come back shortly after.

It's also a good security feature as this ensures it can't ever enter it unexpectedly, this is useful for automated environments like Docker.

As to not break previous behaviour this has to be enabled manually by setting the DISABLE_BASIC_SETUP  environment variable to true.

Let me know what you think, and if i should also try to get support for this in the config file.
I think for most users wanting to disable basic setup this should be enough however.

I thank you all for your time developing Quassel.

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -183,8 +183,13 @@ void Core::init()
             if (writeError) {
                 throw ExitException{EXIT_FAILURE, tr("Cannot write quasselcore configuration; probably a permission problem.")};
             }
-
+            
             qInfo() << "Core is currently not configured! Please connect with a Quassel Client for basic setup.";
+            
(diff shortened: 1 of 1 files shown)


justjanne:
You already can do this by using `--config-from-environment` if you set the correct environment variables to configure the core. It'll never enter basic setup mode with that flag set.

luckcolors:
Allright thank you!
Couldn't find this on the documentation.
Glad to see it's already implemented. 😄

PR #615, opened 2023-04-18
uisupport: fix application name for .desktop shell integration

jeremyvisser:
When building for KDE (cmake -DUSE_KDE=1), the KAboutData constructor as invoked by uisupport causes the resulting application name to be "org.kde.quassel".

At least on GNOME, this "org.kde.quassel" doesn't match the corresponding "quasselclient.desktop" file, which means the app doesn't get a pretty name/icon in the app launcher, which looks pretty ugly:

1-bad

The solution is to call [KAboutData::setDesktopFileName()](https://api.kde.org/frameworks/kcoreaddons/html/classKAboutData.html) with the desired name. This means the corresponding .desktop is matched, and you get a pretty name/icon:

2-good

This issue doesn't occur when building with USE_KDE=0 for two reasons:

1. Because QtUiApplication already calls QGuiApplication::setDesktopFileName() with the correct value (which is, unfortunately, ignored when USE_KDE=1); and
2. Even if desktopFileName were to be unset, the binary name "quasselclient" would be used instead, which correctly matches "quasselclient.desktop".

For these reasons, we only need to fix the KAboutData settings.

An alternative workaround would be to set this flag in the quasselclient.desktop file:

    StartupWMClass=org.kde.quassel

But I would say this is worse because the "org.kde" doesn't make sense, since this is not a KDE project, and because the WM_CLASS would differ depending on the value of USE_KDE=<0|1>.

### DIFF
--- a/src/uisupport/aboutdata.cpp
+++ b/src/uisupport/aboutdata.cpp
@@ -114,13 +114,14 @@ AboutData& AboutData::addCredits(std::initializer_list<AboutPerson> credits)
 
 KAboutData AboutData::kAboutData() const
 {
-    KAboutData aboutData(Quassel::buildInfo().applicationName, tr("Quassel IRC"), Quassel::buildInfo().plainVersionString);
+    KAboutData aboutData(Quassel::buildInfo().clientApplicationName, tr("Quassel IRC"), Quassel::buildInfo().plainVersionString);
     aboutData.addLicense(KAboutLicense::GPL_V2);
     aboutData.addLicense(KAboutLicense::GPL_V3);
     aboutData.setShortDescription(tr("A modern, distributed IRC client"));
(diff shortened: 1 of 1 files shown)


jeremyvisser:
In case you're wondering why I'm using a client built with USE_KDE=1 when I'm using GNOME, that's because Fedora builds it that way: https://src.fedoraproject.org/rpms/quassel/blob/rawhide/f/quassel.spec

eloydegen:
I was looking up how to fix this myself, until I discovered there was already a fix here! Thanks :)

@Sput42, could you please approve and merge this?

Sput42 (approved):

Sput42:
Thanks a lot for the contribution, sorry for ignoring it for so long...

PR #619, opened 2024-09-18
Fix notifications when using KDE integration

wylfen:
Here's the following two changes:
1) fix notifications on KDE systems after they were accidentally broken in #615.
2) make sure to set the DesktopEntry correctly in notifyrc

I've tested this on my system (Gentoo w/ KDE Plasma 6.1.4, Frameworks 6.5.0) and a current Live CD of Fedora Workstation 40 to make sure that shell integration still worked. Whilst it seems that the second patch is not specifically necessary to make notifications work again  (at least on my system), it still seems prudent to follow the documentation.


data/CMakeLists.txt
```diff
@@ -1,6 +1,14 @@
...
-        install(FILES quassel.notifyrc DESTINATION ${CMAKE_INSTALL_KNOTIFY5RCDIR})
+		set(DESKTOP_FILE "quasselclient")
+
+		if (WANT_MONO)
+			set(DESKTOP_FILE "quassel")
+		endif()
+
+		configure_file(quassel.notifyrc.in quassel.notifyrc @ONLY)
+
+		install(FILES ${CMAKE_BINARY_DIR}/data/quassel.notifyrc DESTINATION ${CMAKE_INSTALL_KNOTIFY5RCDIR})
```
reedy, line 11: Tab indented when the file is space

digitalcircuit:
This fixed KNotifications for me on Kubuntu 22.04 LTS.  I'm not sure if there might be any unintended side effects, but for my use case, everything seems good and I can return to running from source.

Thank you!

PR #620, opened 2024-10-09
Broken key exchange fix (Compatability with other clients)

sm1rky:
Hello,

this fixes a bunch of crap in the crypto key exchange department. I tested it using other clients (mirccryption), irssi (falsovski irssi-fish) with the monolithic build.

basically fixes https://bugs.quassel-irc.org/issues/1857

simultaneously fixes a rare bug i encountered:   when channel and topic were set with a different key. decryption of topic will then lead to random garbage bytes. Some which will still trigger the stack overflow despite the previous fixes.

also seems to cover integrates/resolve these features pull reqs:
mcps * cbc trigger decryption: https://github.com/quassel/quassel/pull/620
crash on broken topic: https://github.com/quassel/quassel/pull/612

best regards

### DIFF
--- a/src/core/cipher.cpp
+++ b/src/core/cipher.cpp
@@ -40,14 +40,14 @@ bool Cipher::setKey(QByteArray key)
         return false;
     }
 
+    m_key = key;
+
     if (key.mid(0, 4).toLower() == "ecb:") {
         m_cbc = false;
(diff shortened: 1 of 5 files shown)


Sput42 (approved):
Fixed broken commit msg, otherwise LGTM. Thanks for your contribution!

PR #622, opened 2025-03-26
Added client-side support for implicit TLS connections

Avamander:
This PR adds client-side support for implicit TLS connections. It's intentionally very minimal, no large refactors, in order to make merging and feedback easier.

This is intended to work with proxies that terminate TLS before the application. From the operational side it significantly simplifies certificate management. Implicit TLS connections also reduce potential privacy leaks and prevent any potential (downgrade) attacks during Quassel protocol negotiation.

In terms of future improvements, I am thinking about implementing the Proxy Protocol (v2) for accurate logging. I would also like to enforce TLSv1.3 minimum on implicit TLS connections, if there are no objections. In general it would be nice to switch to "TLS" as the term used in user-facing strings and also code, instead of "SSL" which is obsolete for more than a decade now.


justjanne:
Quassel already implements Proxy Protocol v1, btw.

I'll try to add support for this in Quasseldroid, seems like a useful feature.

PR #626, opened 2025-07-12
Qt6 port

FuzzyGophers:
This should put Quassel at an alpha stage for the port.

This will allow Sput to begin reviewing the mass amount of changes and allow others to collaborate on the port.

I will continue working on issues as well.


digitalcircuit:
To guard against these commits getting lost (in case GitHub forgets them from this PR, etc), I've copied them over to [`ft-fg-qt6-port` on my fork of the repo](https://github.com/digitalcircuit/quassel/tree/ft-fg-qt6-port ).

I still need to set up a Qt 6 build environment since this appears to entirely drop Qt 5 support, and Ubuntu 24.04 does not have a new-enough Qt version:
```
  Could not find a configuration file for package "Qt6" that is compatible
  with requested version "6.8".

  The following configuration files were considered but not accepted:

    /usr/lib/x86_64-linux-gnu/cmake/Qt6/Qt6Config.cmake, version: 6.4.2
    /lib/x86_64-linux-gnu/cmake/Qt6/Qt6Config.cmake, version: 6.4.2
```

@Sput42 If you'd like me to open a PR to continue work, let me know.  Or feel free to open your own PR (I have not made any changes yet, I was waiting until I could get it to build to do that).

PR #627, opened 2025-09-17
Start of port to Qt 6

Tatsh:
[Feature: Port to QT6](https://bugs.quassel-irc.org/issues/1897)

Before you can use this, you must run the following to fix the `po/CMakeLists.txt` file:

```shell
sed -e 's/Qt5/Qt6/g' -i po/CMakeLists.txt
```

This gets the app running. All KDE parts are ported to KF6. Tests required a very minor change. Not a lot of effort was put here to keep it building on Qt 5 (and there are some removals). At this point I see no reason to have that as 0.14.0 is fine. Tested on Linux and macOS. Have not tested running the server-side yet.

The only part still needing QtCore5Compat is anything relying on `QTextCodec`. All usages of `QRegExp` have been updated to `QRegularExpression`. All other deprecated (except `QtSQL`) usages have been updated.

## Known issues

- ~~Tab completing nicks does not work properly.~~
- Chat list colours do not update appropriately (they stay blue for me regardless of activity).
- ~~Network online detection is removed due to no longer having `QNetworkConfigurationManager`.~~
- Issue with font weight being in the settings QSS file. `normal` is in the file alongside a numeric weight, but  normal takes effect and no weight is applied. Manually editing the file to have a line like `font-weight: bold 14pt;` is a workaround.
- Server needs a lot of work. Currently this just crashes on start. Tried a few fixes but it is a lot more than a few quick patches.

PR #630, opened 2026-01-07
FuzzyGopher's Qt6 port

Sput42:
Adding this back as a PR to make it easier to work with and compare with #627.


toralf:
I applied 630.patch here on top of a -9999 package of Gentoo, but got :
```
>> Preparing source in /var/tmp/portage/net-irc/quassel-9999/work/quassel-9999 ...
 * =====================================================================================================================
 * Applying user patches from /etc/portage/patches ...
 * Applying 630.patch ...
 [ ok ]
 * User patches applied.
 * =====================================================================================================================
 * Source directory (CMAKE_USE_DIR): "/var/tmp/portage/net-irc/quassel-9999/work/quassel-9999"
 * Build directory  (BUILD_DIR):     "/var/tmp/portage/net-irc/quassel-9999/work/quassel-9999_build"
 * QA Notice: Compatibility with CMake < 3.5 has been removed from CMake 4,
 * net-irc/quassel will fail to build w/o a fix.
 * See also tracker bug #951350; check existing or file a new bug for this package.
 * Please also take it upstream.
 * 
 * The following files are causing errors:
 *   3rdparty/googletest-1.8.1/googletest/CMakeLists.txt:2.6.4
 *   3rdparty/googletest-1.8.1/googlemock/CMakeLists.txt:2.6.4
 *   3rdparty/googletest-1.8.1/CMakeLists.txt:2.8.8
 * 
 * Compatibility w/ CMake < 3.16 will be removed in future ECM release.
 * See also tracker bug #964407; check existing or file a new bug for this package.
 * If not fixed in upstream's code repository, we should make sure they are aware.
 * 
 * CMake 4 detected; building with -DCMAKE_POLICY_VERSION_MINIMUM=3.5
 * This is merely a workaround to avoid CMake Error and *not* a permanent fix;
 * there may be new build or runtime bugs as a result.
 * 
 * An upstreamable patch should take any resulting CMake policy changes
 * into account. See also:
 *   https://cmake.org/cmake/help/latest/manual/cmake-policies.7.html
```
and later 
```
undefined reference to `operator<<(QDataStream&, Identity)'
```
[quassel.txt](https://github.com/user-attachments/files/24858537/quassel.txt)

PR #633, opened 2026-06-25
chatview: add hidpi support to sender column

justjanne:
Now we just use the dpr of the widget for our own pixmap.

image


justjanne:
And it just works, even at significantly higher resolutions:
Screenshot_20260625_230100

Sput42 (approved):

